PR #23, opened 2018-09-28
npm grunt-contrib packages are dev dependencies

### LINKED ISSUE #13: Isn't grunt a dev dependency, not a prod dependency?
See

https://github.com/ProjectMirador/research-and-demos/blob/master/minimal_redux_poc/package.json

note that babel and webpack are dev dependencies ...

If this is correct, do a PR for lcnetdev upstream repo (via LD4P/profile-edit branch then PR via github to upstream repo).

ndushay:
also uses the released version of grunt-contrib-uglify-es instead of the one from git.

Closes #13 

note upstream PR is lcnetdev/bfe#43

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "eslint-plugin-node": "^7.0.1",
     "eslint-plugin-security": "^1.4.0",
     "express": "^4.16.3",
+    "grunt-contrib-concat": "^1.0.1",
     "grunt-contrib-cssmin": "^3.0.0",
     "grunt-contrib-uglify-es": "^3.3.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -12,8 +12,8 @@
     "node": ">= 0.9.0"
   },
   "contributors": [
-     "Kirk Hess <khes@loc.gov>",
-     "Jeremy Nelson"
+    "Kirk Hess <khes@loc.gov>",
+    "Jeremy Nelson"
```
jermnelson: I would add @ndushay and @jgreben as well!

PR #24, opened 2018-09-28
package.json: lock to node version 10.11.x

ndushay:


ndushay:
ja - this actually didn't change package-lock because i cleverly used the version of node already on my machine.  :-)

PR #35, opened 2018-10-02
CI build includes eslint, which will fail on errors or additional warnings

### LINKED ISSUE #26: CI runs linter
We want CI to:

- run eslint. <-- fail for problems
- do grunt build. <-- fail for problems
- have tests that do something. <-- fail for problems
    - with code coverage. <-- fail for reducing coverage (see #14)
depends on #27

### LINKED ISSUE #27: linter runs cleanly
- is there a way to generate a TODO list so it can ignore existing problems, e.g. in static folder (so we don't diverge from upstream)
- is there an autocorrect we can use for the files of interest (e.g. bfe)

See https://github.com/LD4P/sinopia_profile_editor/pull/42

we can use `.eslintrc` to configure rules so build fails on what WE consider to be egregious errors and `.eslintignore` to ignore upstream js.

ndushay:
Adds eslint and hooks it up with CI build.  It would be good for both @jgreben and @jermnelson to review.

We have to start somewhere. This is a pretty bad place to start ... but it is a way to have our CI builds fail if we have an eslint error, or if the warning count goes up.

image

As we start refactoring and changing code, we'll need to adjust the max-threshold for eslint script in package.json:

image

I felt a few linting errors were worth keeping as errors (not warnings), so I have some small tweaks to files to address these errors from linting.

595070d, the commit moving express to be a regular dependency, not a devDependency *seems* right - we need express to actually run the javascript for a user, no?  "Express is a minimal and flexible Node.js web application framework that provides a robust set of features for web and mobile applications." (http://expressjs.com/).   Kirk may do things differently ...  but the real question is ... what will our users need (in AWS?)

Happy to squash commits, etc.

Closes #27 
Closes #26

### DIFF
--- a/src/bfelookups.js
+++ b/src/bfelookups.js
@@ -132,7 +132,6 @@ bfe.define('src/lookups/lcnames', ['require', 'exports', 'module', 'src/lookups/
       var triple = {};
       triple.s = subjecturi;
       triple.p = property.propertyURI;
-      selected.uri = selected.uri;
       triple.o = selected.uri;
       triple.otype = 'uri';
       triples.push(triple);
--- a/src/bfelookups.js
+++ b/src/bfelookups.js
@@ -1,16 +1,16 @@
 bfe.define('src/lookups/lcnames', ['require', 'exports', 'module', 'src/lookups/lcshared'], function (require, exports, module) {
     var lcshared = require('src/lookups/lcshared');
-  
+
     var cache = [];
-  
+
     // This var is required because it is used as an identifier.
     exports.scheme = 'http://id.loc.gov/authorities/names';
-  
+
     exports.source = function (query, process, formobject) {
       // console.log(JSON.stringify(formobject.store));
-  
+
       var triples = formobject.store;
-  
+
       var type = '';
       var hits = _.where(triples, {
         'p': 'http://www.w3.org/1999/02/22-rdf-syntax-ns#type'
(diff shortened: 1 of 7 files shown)


jermnelson:
Looks good to me.  We will be using Express for routing to the new components in the Sinopia Editor so moving it to a regular dependency is needed anyway.

jermnelson:
If we can squash some of these commits, I'm ready to merge. @jgreben any thoughts?

src/bfelookups.js
```diff
@@ -110,68 +110,68 @@ bfe.define('src/lookups/lcnames', ['require', 'exports', 'module', 'src/lookups/
...
+
     exports.getResource = function (subjecturi, property, selected, process) {
       var triples = [];
-  
+
       var triple = {};
       triple.s = subjecturi;
       triple.p = property.propertyURI;
-      selected.uri = selected.uri;
+      // selected.uri = selected.uri;  // FIXME: I *think* this line is useless but feel too green to remove it
```
ndushay: is there any reason to keep this line?

jgreben: I think that self-referential line can be totally deleted. It does absolutely nothing. If you want to push another commit to remove it that would be ok with me.

jgreben:
@jermnelson I'm ok with having a commit list remain on the PR as long as the commits signify atomic functional changes. I usually squash commits when they are of the nature of "FU" and the like (things that didn't really end up getting changed but were temporary or experiments). This PR is a good example of a helpful commit history with atomic functional changes. But I'll go with whatever practice you all want to do as far as squashing commits.

jermnelson:
Merging w/current commit history.

PR #36, opened 2018-10-03
[WIP] CI hooks coveralls to jest coverage

### LINKED ISSUE #14: Code coverage stats part of CI builds (requires #3)
We want to have an easy way to track code coverage stats pertaining to each PR and overall.  We want CI build to fail if coverage drops.

Note that CI build currently uses `--coverage` flag passed to jest but I suspect we need to configure something to get useful data.

See https://github.com/istanbuljs/nyc, esp pertaining to code coverage -- can have build fail if coverage drops  (not sure if we can use nyc with jest-puppeteer).

ndushay:
WIP until #35 is merged.

If this works like our ruby projects, the PR will fail if the coverage drops.  (Can't really drop much below 0, tho!). We should be able to configure coveralls in a variety of ways (how much % change do we care about, what is our threshhold, etc.).

Closes #14

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,8 +28,11 @@ jobs:
           paths:
             - node_modules
       - run:
+          name: Run eslint
+          command: npm run eslint --silent
+      - run:
           name: Build the project with grunt
-          command: npm run grunt
+          command: npm run grunt --silent
       - run:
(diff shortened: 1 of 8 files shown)


jermnelson:
Didn't mean to close this branch; reviewing now

Issue #10, opened 2018-09-26
Use QA instead of `lookups` array


ndushay:
The lookup array at https://github.com/LD4P/bfe/blob/master/src/bfe.js#L32-L122 should be removed/replaced with look-up in QA.

_copied from LD4P/bfe#10_

jermnelson:
QA is not supporting these vocabularies, (see https://github.com/LD4P/qa_server/wiki/Authority-Support-Plan#other-authorities) so closing this issue.

PR #65, opened 2018-10-11
add profile fixtures to __tests__/__fixtures__

### LINKED ISSUE #48: testing fixtures needed
- profiles to exercise editor functionality (small ones ok) (see https://github.com/lcnetdev/verso/tree/master/data/profiles (via https://github.com/LD4P/sinopia/issues/7) )
- config.js  ?   possibly winnowed way down for testing with jest-puppeteer

ndushay:
I added a bunch of profiles from https://github.com/lcnetdev/verso/tree/master/data/profiles

some minor concerns:
- I took most, but not all.  (not maps, not call numbers, not media, not notated music, not prints and photographs, not rare materials, not serials ...). I assume that's going to be okay for testing.
- I didn't keep the really long names ... do we care?
- I reformatted the JSON to have 2 spaces indentation ... any concern with upstream changes (I don't expect any)
- There is no way to comment in JSON, so there is no way to attribute the fixture data.
- It's often not a good thing to have fixtures you don't end up using.  But it seems handy to have a bunch of them for now.  Not sure how else to proceed easily with e2e tests, working in parallel.

Closes #48

### DIFF
--- a/__tests__/__fixtures__/admin.json
+++ b/__tests__/__fixtures__/admin.json
@@ -1,0 +1,489 @@
+{
+  "Profile": {
+    "resourceTemplates": [
+      {
+        "propertyTemplates": [
+          {
+            "mandatory": "false",
+            "repeatable": "true",
+            "type": "literal",
+            "resourceTemplates": [],
+            "valueConstraint": {
+              "valueTemplateRefs": [],
+              "useValuesFrom": [],
+              "valueDataType": {
+                "remark": "http://id.loc.gov/ontologies/bflc/catalogerId"
+              },
+              "defaults": []
+            },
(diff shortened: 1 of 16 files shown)


jermnelson:
Some non-blocking comments:

-  Renaming the long file names will help in testing, the current names with spaces makes it more difficult to properly escape when we want to use them in the integration tests
-   I doubt we will use most of these profiles in our initial tests but will be helpful as we identify and test edge cases
-   We may have to update these fixtures when the new profile properties are added to the Sinopia Profile Editor (see [#8](https://github.com/LD4P/sinopia_profile_editor/issues/8), [#9](https://github.com/LD4P/sinopia_profile_editor/issues/9), [#10](https://github.com/LD4P/sinopia_profile_editor/issues/10))

PR #68, opened 2018-10-12
tweak browse test descriptions

ndushay:
Actually, the tests were about the browse functionality, and they are currently just showing the datatable is loaded.

When the datatable is loaded, it contains data, not profiles.


jermnelson:
Yep, my mistake. Thanks!

PR #69, opened 2018-10-16
integration test can affirm that profile is loaded

### LINKED ISSUE #46: should eslintrc get globals?
so I was wondering if adding this to .eslint.js would make sense:
```globals: {
        "bfe": true,
        "bfeditor": true,
        "bfelog": true
        "config": true,
        "document": true,
        "module": true,
        "page": true,
        "window": true
      }
```
i don't know enough js yet to be certain these would be appropriate adds.

### LINKED ISSUE #49: test: able to run app in jest-puppeteer with profile(s) loaded
requires #48 (fixture profiles).  May require a fixture config.js file. 

 This is meant to be a basic test showing we can spin up app and the profiles are available - that's it.

Alternatively, could put a puppeteer.debug or whatever statement in existing test to stop the app while browser is avail and see if fixture config is used and fixtures profiles are avail.

ndushay:
Aside from writing the test code itself, this involved:

- adding some fixture json files to be accessed via local url
- a testing copy of config.js (__tests__/integration/test-config.js)
    - changes hardcoded url for config.profiles from verso to localhost for short(er) profiles fixture json
    - changes the hardcoded url in "retrieve" function to localhost for instance ontology fixture json
- a testing copy of index.html using test-config.js
- turning off web security (CORS) when running puppeteer in order to be able to get localhost json
- tweaked eslint globals (got busted on my test code!)

This approach is brittle on a few counts:
- test-config.js will drift from actual config.js
- test index.html will drift from actual index.html
- hardcoding only allows retrieval of bibframe instance ontology resource template, no others.
- can't run server locally, as CORS will prevent profiles and resource template from loading

However, it is a test showing that a profile is loaded, and it's a start.

Also created tickets
-  #70 - a small refactoring of config.js to greatly improve code testability
- #71 - split the config code from the data - have a json (or yml or whatever) file populating menu, not part of code itself

Closes #49
Closes #46

### DIFF
--- a/__tests__/integration/test-config.js
+++ b/__tests__/integration/test-config.js
@@ -1,0 +1,494 @@
+var ie = (function(){
+    var undef,
+        v = 3,
+        div = document.createElement('div'),
+        all = div.getElementsByTagName('i');
+    while (
+      div.innerHTML = '<!--[if gt IE ' + (++v) + ']><i></i><![endif]-->',
+      all[0]
+    );
+    return v > 4 ? v : undef;
+  }());
+
+  if (ie < 10) {
+    $("#iealert").addClass("alert alert-danger");
+    $("#iealert").html("Sorry, but the BIBFRAME Editor will not work in IE8 or IE9.")
+  }
+
+  function myCB(data) {
+    document.body.scrollTop = document.documentElement.scrollTop = 0;
+  }
+
+  function csrfSafeMethod(method) {
+    // these HTTP methods do not require CSRF protection
+    return (/^(GET|HEAD|OPTIONS|TRACE)$/.test(method));
+  }
+
+  function getCSRF(){
+    //eventually you'll have to login
+    var cookieValue = null;
+    if (document.cookie && document.cookie != '') {
+      var cookies = document.cookie.split(';');
+      for (var i = 0; i < cookies.length; i++) {
+        var cookie = jQuery.trim(cookies[i]);
+        var name = "csrftoken";
+        if (cookie.substring(0, name.length + 1) == (name + '=')) {
+          cookieValue = decodeURIComponent(cookie.substring(name.length + 1));
+          break;
+        }
+      }
+    }
+    return cookieValue;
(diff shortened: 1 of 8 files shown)


ndushay:
Ha!  I got busted by eslint.  Not that it wasn't a problem that already appears in the code a lot ...

jgreben:
Are we ok with the editor looking up the Instance json from bibframe.org for this test?

http://bibframe.org/bibliomata/profile-edit/server/whichrt?uri=http%3A%2F%2Fid.loc.gov%2Fontologies%2Fbibframe%2FInstance.json

__tests__/integration/test-config.js
```diff
@@ -0,0 +1,494 @@
...
+  }
+
+  /* Config object profiles
+   * Editor profiles are read from a WS endpoint
+   * The data are expected to be in a JSON array, with each object
+   * in the array containing a "json" property that has the profile
+   * itself. The "versoURL" variable is a convenience for setting the
+   * base URL of verso in the "config" definition below.
+   */
+  var rectoBase = "http://bibframe.org/bibliomata";
```
jgreben: Are we ok with the editor looking up the Instance json from bibframe.org for this test? E.g.

http://bibframe.org/bibliomata/profile-edit/server/whichrt?uri=http%3A%2F%2Fid.loc.gov%2Fontologies%2Fbibframe%2FInstance.json

I cannot replicate the steps of this test manually in my browser when running it locally because of the reference to that json lookup. Maybe the intention is to always look up this json from LOC?

ndushay: my experience was that this was addressed by hijacking the url for config.profiles.  I did the minimum to get the tests to pass, and there were some 404 errors noted in the console.

jgreben: Ok, I see that the `whichrt` things are shortcut if I use the `index.html` and `test-config.js` in place of the real files. This PR LGTM 👍 .

Issue #53, opened 2018-10-10
editor tab tests: startingPoints menu works


ndushay:
Requires #52
-  Able to select something from the displayed menu (monograph, etc) (from fixture config.js - see https://github.com/LD4P/sinopia_editor/blob/master/static/js/config.js#L263) (see PR #69 for issue #49)
- resulting popup menu shows correct options (from fixture config.js) (see PR #79 for issue #52)
- selecting from popup menu shows
    - correct text (e.g. "BIBFRAME Instance") (see PR #69 for issue #49)
    - correct input fields (and labels based on profile)
        - Instance of xxx
        - 2-3 fields specific to that entity
    - correct buttons (cancel, preview, ? clone)

PR #74, opened 2018-10-17
tests use expect_regex_in_selector_textContent function

ndushay:
I was doing wAAAAAy too much copypasta working on #53 so I figured out a better way* that also makes our tests a lot more readable.

I know the factored out method is repeated in each of the test files that use it;  I spent some time trying to figure out how to dry that up and ultimately gave up.  If anyone thinks it's easily doable, please make a ticket for it -- perhaps we could have a `__tests__/support` directory and load/require/import files from there as needed?  I don't know enough js yet to do it, and there didn't seem a way to do this wired into jest or jest-puppeteer (I read all about different flavors of `setup` but didn't find anything useful for this use case.)

* what *I* think is a better way, anyhow.


__tests__/integration/editorTab.test.js
```diff
@@ -6,56 +6,42 @@ describe('Editor Tab', () => {
...
-    await expect(blu_ray).toMatch(/Moving Image: BluRay DVD/)
-    const moving_img_35 = await page.$eval('#bfeditor-menudiv > ul:nth-child(10) > li > a',  e => e.textContent)
-    await expect(moving_img_35).toMatch(/Moving Image: 35mm Feature Film/)
-    const print_photos = await page.$eval('#bfeditor-menudiv > ul:nth-child(11) > li > a', e => e.textContent)
-    await expect(print_photos).toMatch(/Prints and Photographs/)
-    const rare = await page.$eval('#bfeditor-menudiv > ul:nth-child(12) > li > a', e => e.textContent)
-    await expect(rare).toMatch(/Rare Materials/)
-    const authorities = await page.$eval('#bfeditor-menudiv > ul:nth-child(13) > li > a', e => e.textContent)
-    await expect(authorities).toMatch(/Authorities/)
+    await expect_regex_in_selector_textContent('#bfeditor-menudiv > ul:nth-child(2) > li > a', /^Monograph$/)
```
atz: I'm new to this particular tooling, but it seems like overkill to use regex matching for things that are just literal strings.  

I also don't understand why every criteria must be independently asynchronously paired (i.e. `await`ed). 

Is drilling down after a single `await` like this not viable?
```javascript
const menudiv = await(page.$eval('#bfeditor-menudiv'))
expect(menudiv.$eval('ul:nth-child(2) > li > a', e => e.textContent)).toBe('Monograph')
expect(menudiv.$eval('ul:nth-child(3) > li > a', e => e.textContent)).toBe('Notated Music')
```

ndushay: I am hearing two concerns in your comment:

1.  all the awaits.

$eval is a method on the page object via puppeteer's api (https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#pageevalselector-pagefunction-args-1 ), and it comprises a promise, which requires the await.  Presumably the promise/await mechanism of javascript is implemented such that the full page object is not reloaded.

Here's what I got when I tried your suggestion:

image

2.  regex vs exact match

This is a good point.  I spent some time on this page https://jestjs.io/docs/en/expect and managed to miss "toBe".  I will update.

__tests__/integration/profilesLoad.test.js
```diff
@@ -19,8 +19,11 @@ describe('Profiles load', () => {
 
     // await page.screenshot({path: 'screenshot.png', fullPage: true})
 
-    const ptitle_sel = 'div#bfeditor-formdiv > form > div > h4'
-    const profile_title = await page.$eval(ptitle_sel, e => e.textContent)
-    await expect(profile_title).toMatch(/BIBFRAME Instance/)
+    await expect_regex_in_selector_textContent('div#bfeditor-formdiv > form > div > h4', /^BIBFRAME Instance/)
```
ndushay: Note that this one does need to be a regex because "Clone Instance" text also turns up inside the h4.  :-P

__tests__/integration/editorTab.test.js
jermnelson: In your example `menudiv` is actually a promise and not the element itself. If `await` is not used on the second line, the expect will fail immediately because the promise has not been fulfilled returning the element that is being tested.

ndushay:
@jermnelson @jgreben I think Joe is politey just "commenting" here so one of you should merge this if you think it's okay.

__tests__/integration/editorTab.test.js
atz: Thanks for the reply.  If I'm reading the docs correctly, would that be fixed by adding [`.resolves`](https://jestjs.io/docs/en/asynchronous#resolves-rejects)?

I get that I was treating return of `$eval` like it was jquery.  I guess if I wanted to do that, I would follow jest's ["DOM Manipulation" docs](https://jestjs.io/docs/en/tutorial-jquery) for actually having jquery.

atz:
Right, I would "request changes" (i.e. block) if I felt like anything was actually blocking.

PR #83, opened 2018-10-24
[WIP] Build and post a Docker Container from Circle-CI

atz:


atz:
Currently failing only from unrelated intermittent failures.

PR #86, opened 2018-10-27
Remove serve-static from direct dependencies

### LINKED ISSUE #25: npm serve-static isn't used?
If this is true, then npm package can be removed.

https://www.npmjs.com/package/serve-static implies that the only public method provided is serveStatic, and that's not called anywhere.

If this is true, also submit upstream PR to lc bfe code via ld4p/bfe-defunct.

depends on #16 and probably some other basic integration tests

atz:
Notice the lockfile doesn't change because serve-static is still an *indirect* dependency via `express` framework.

Fixes #25

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,6 @@
     "jest-puppeteer": "^3.4.0",
     "puppeteer": "^1.8.0",
     "react-hot-loader": "^4.3.11",
-    "serve-static": "^1.13.2",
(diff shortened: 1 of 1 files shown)


jermnelson:
Hi @atz, if your can rebase this branch, I'll merge this PR into master.

Thanks,
Jeremh

PR #87, opened 2018-10-30
Setup news panel

SaravShah:
- Add JQuery, Bootstrap, CSS etc.  to `index.html`
- Add `react-bootstrap`, `css-loader`, and `style-loader` dependency. Need the `loaders` because they interpret `import/requires` . If we don't have the loaders, tests will fail.
- LoginPanel currently uses raw html/css and is not done the "react" way. When implementing the Login stuff we will need to change this.

screen shot 2018-10-30 at 1 47 21 pm

When clicking on `Help and Resources` we get: 

screen shot 2018-10-30 at 3 07 12 pm

connects LD4P/sinopia#111

Note: Tests will be in the next PR I put up :)


src/components/LoginPanel.jsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+
+const LoginPanel = () => (
+  <form class="login-form">
```
jermnelson: Maybe add a TODO to convert this form into React Component(s)?

src/components/NewsPanel.jsx
```diff
@@ -0,0 +1,27 @@
...
+import LoginPanel from './LoginPanel.jsx'
+
+const NewsPanel = () => (
+  <div class="jumbotron banner center-block">
+    <div class="panel panel-default">
+      <div class="panel-body">
+        <div class="row">
+          <div class="col-md-8">
+            <h1> Latest News </h1>
+            <ul>
```
jermnelson: We could convert these `<li>` into a React Component like `NewsItem` or something. This is fine for now.

SaravShah: #88

PR #95, opened 2018-11-02
Updating merge version 

jermnelson:
Fix security alert from Github about updating version of `merge` used in package-lock.json.


ndushay:
i might've put the new version in the commit message, too.

PR #102, opened 2018-11-02
remove config-dev.js and development.html

ndushay:
AFAIK, we're not using either of these, and the spurious matches on things like verso Urls are annoying me.

@jgreben and @SaravShah are okay with this;  @jermnelson can you think of a reason we'd want to keep these?

I also regenned eslint todo for good measure

@SaravShah you might want to look at what the linter picked up for the react code ... or maybe it doesn't matter.  I didn't look closely.


jermnelson:
Although the `config-dev.js` implements the core BFE API calls to the verso like `save` and `publish`, I'm good with removing it. If we need to refer to this code we can always look in the `bfe-defunct` code base.

Issue #96, opened 2018-11-02
Implement React Router


SaravShah:
[x] Create a link from home page to an empty page to test Naomi and Josh's Findings
[  ] Update tests so circleci can pass

PR #117, opened 2018-11-07
React router

SaravShah:
- Create separate React Component for the BFF editor page. 
- The route is `/editor`
- implement router testing via `react-router`
- per my last commit, in Enzyme, if you want to use `mount()` which renders full DOM, we need to install `jsdom`. -> https://airbnb.io/enzyme/docs/guides/jsdom.html

closes LD4P/sinopia_editor#96


SaravShah:
@ndushay yup- also removed the semicolons from `components/` and `__tests__/components` :)

PR #146, opened 2018-11-12
Apache 2.0 Header added to Source Files 

### LINKED ISSUE #114: Update LICENSE to Apache 2.0
Update README, `package.json`, and the LICENSE.txt to use Apache 2.0 or approved license for Stanford use. Need to reference code license that is used in the `bfe` codebase

jermnelson:
Following discussion with Vivian, updated all Stanford originated source code with the Apache 2 header.  Used a Library of Congress upstream commit, removed LICENSE.txt, added new LICENSE with current `bfe` text and added the full Apache 2.0 license to the file.

Closes #114


jermnelson:
It would be nice if there is a convention for a single line notation in javascript

SaravShah:
What if we move the license to another Stanford specific `LICENSE.txt`. Then at the top of each file refer to that `LICENSE.txt`. I think this would be better + less noisy.

ndushay:
I'm certain in the past we haven't had to do a per-file thingy, but perhaps because there is still some belief we will have a mix of files from LC and files from us we need something?  I like the one-liner idea.  Take an ask-for-forgiveness approach?

jermnelson:
Replaced Apache 2 header with a single line  `Copyright 2018 Stanford University see Apache2.txt for license`, added Apache2.txt license, keeping LICENSE for original BFE code.

PR #148, opened 2018-11-13
Adds components to render resource template value constraint properties

### LINKED ISSUE #123: parse ValueConstraints from PropertyTemplates
- finish code
- write tests
- do PR

(blocked by #122)

jgreben:
Fixes #123 

- adds carrier property to static resource template json in Editor.jsx
- adds a new `<ValueConstraints />` component with sub-components:
  - `<ValueTemplateRefs />`
  - `<UseValuesFrom />`
  - `<ValueDataTypes />`
  - `<Defaults />`
- function to check whether a hash has a key/value pair, used for conditional display
- test that uses enzyme full rendering with `mount` and a way to test a component's properties directly.

### DIFF
--- a/__tests__/components/editor/ValueConstraints.test.js
+++ b/__tests__/components/editor/ValueConstraints.test.js
@@ -1,0 +1,199 @@
+import React from 'react'
+import { mount } from 'enzyme';
+import ValueConstraints from '../../../src/components/editor/ValueConstraints'
+
+const jsdom = require("jsdom");
+
+setUpDomEnvironment();
+
+const vcProps = {
+  "propertyTemplates": [
+    {
+      "valueConstraint": {
+        "defaults": [
+          {
+            "defaultLiteral": "DEFAULT",
+            "defaultURI": "http://default"
+          }
+        ],
+        "editable": "true",
+        "languageLabel": "LANGUAGE LABEL",
+        "languageURI": "http://id.loc.gov/vocabulary/languages/eng",
+        "remark": "REMARK",
+        "repeatable": "false",
+        "useValuesFrom": [
+          "http://VALUES"
+        ],
+        "validatePattern": "PATTERN",
+        "valueDataType": {
+          "dataTypeLabel": "Classification item number",
+          "dataTypeLabelHint": "HINT",
+          "dataTypeURI": "http://id.loc.gov/ontologies/bibframe/itemPortion",
+          "remark": "REMARK"
+        },
+        "valueLanguage": "VALUE LANGUAGE",
+        "valueTemplateRefs": [
+          "profile:bf2:Identifiers:Barcode"
+        ]
+      }
+    }
+  ]
+}
+
+const props = {}
+const vcProp = vcProps.propertyTemplates[0].valueConstraint
+props.defaults = vcProp.defaults
+props.editable = vcProp.editable
+props.languageLabel = vcProp.languageLabel
+props.languageURI = vcProp.languageURI
+props.remark = vcProp.remark
+props.repeatable = vcProp.repeatable
+props.useValuesFrom = vcProp.useValuesFrom
+props.validatePattern = vcProp.validatePattern
+props.valueDataType = vcProp.valueDataType
+props.valueLanguage = vcProp.valueLanguage
+props.valueTemplateRefs = vcProp.valueTemplateRefs
+
+let wrapper = mount(<ValueConstraints {...props} />)
+
+describe('<ValueConstraints />', () => {
+  it('has div with class "ValueConstraints"', () => {
+    expect(wrapper.find('h6').text()).toEqual('VALUE CONSTRAINTS')
+    expect(wrapper.find('div.ValueConstraints > ul > li')).toHaveLength(11)
+    wrapper.find('div.ValueConstraints > ul > li').filterWhere(n => n.children().length === 0).forEach((node) => {
+      expect(node.containsAnyMatchingElements([
+        'editable: true',
+        'languageLabel: LANGUAGE LABEL',
+        'languageURI: http://id.loc.gov/vocabulary/languages/eng',
+        'remark: REMARK',
+        'repeatable: false',
+        'validatePattern: PATTERN',
+        'valueLanguage: VALUE LANGUAGE'
+      ]))
+    })
+  })
+})
+
+describe('when values are present', () => {
+
+  describe('<Defaults />', () => {
+    it('has defaults', () => {
+      expect(wrapper.find('Defaults > div > ul > li').length).toBe(2)
+      wrapper.find('Defaults > div > ul > li').forEach((node) => {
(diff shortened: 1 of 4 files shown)


src/components/editor/PropertyTemplate.jsx
```diff
@@ -16,6 +17,16 @@ class PropertyTemplate extends Component {
...
           <li>repeatable: <strong>{this.props.propertyTemplates[0].repeatable}</strong></li>
           <li>contained resourceTemplates[] length: <strong>{this.props.propertyTemplates[0].resourceTemplates.length}</strong></li>
+          <li>
+            <ValueConstraints
+              valueTemplateRefs = {this.props.propertyTemplates[0].valueConstraint.valueTemplateRefs}
+              useValuesFrom = {this.props.propertyTemplates[0].valueConstraint.useValuesFrom}
+              valueDataType = {this.props.propertyTemplates[0].valueConstraint.valueDataType}
+              repeatable = {this.props.propertyTemplates[0].valueConstraint.repeatable}
+              editable = {this.props.propertyTemplates[0].valueConstraint.editable}
+              defaults = {this.props.propertyTemplates[0].valueConstraint.defaults}
```
ndushay: this seems like a prime case for using the spread operator:  https://zhenyong.github.io/react/docs/jsx-spread.html  or  https://www.robinwieruch.de/react-pass-props-to-component/#react-props-syntax-spread-rest   

"The props spreading can be used to spread a whole object with key value pairs down to a child component. It has the same effect as passing each value of the object by its own to the component as before. The child component gets the props the same way as before too."

If we're eager to get this PR in, make a follow-on ticket for this?  And refer to that ticket in the code?

__tests__/components/editor/ValueConstraints.test.js
```diff
@@ -0,0 +1,93 @@
...
+})
+
+describe('<UseValuesFrom />', () => {
+  it('has valueTemplateRefs', () => {
+    expect(wrapper.find('UseValuesFrom > div > ul > li').text()).toEqual('http://id.loc.gov/foo/works')
+  })
+})
+
+describe('<ValueDataTypes />', () => {
+  it('has valueTemplateRefs', () => {
```
ndushay: copypasta typo in the description?

__tests__/components/editor/ValueConstraints.test.js
```diff
@@ -0,0 +1,93 @@
+import React from 'react'
+import { mount } from 'enzyme';
+import ValueConstraints from '../../../src/components/editor/ValueConstraints'
+
+const jsdom = require("jsdom");
+
+setUpDomEnvironment();
+
+const vcProps = {
```
jermnelson: Should this `const` be moved to the __mocks__ folder?

__tests__/components/editor/ValueConstraints.test.js
```diff
@@ -0,0 +1,93 @@
...
+}
+
+const valueConstraints = (
+  <ValueConstraints
+    valueTemplateRefs = {vcProps.propertyTemplates[0].valueConstraint.valueTemplateRefs}
+    useValuesFrom = {vcProps.propertyTemplates[0].valueConstraint.useValuesFrom}
+    valueDataType = {vcProps.propertyTemplates[0].valueConstraint.valueDataType}
+    repeatable = {vcProps.propertyTemplates[0].valueConstraint.repeatable}
+    editable = {vcProps.propertyTemplates[0].valueConstraint.editable}
+    defaults = {vcProps.propertyTemplates[0].valueConstraint.defaults} >
```
ndushay: per actual code comment, is this a good place to use the spread operator to pass in props?

__tests__/components/editor/ValueConstraints.test.js
```diff
@@ -0,0 +1,93 @@
...
+})
+
+describe('<ValueTemplateRefs />', () => {
+  it('has valueTemplateRefs', () => {
+    expect(wrapper.find('ValueTemplateRefs > div > ul > li').text()).toEqual('profile:bf2:Monograph:Work')
+  })
+})
+
+describe('<UseValuesFrom />', () => {
+  it('has valueTemplateRefs', () => {
```
ndushay: is this copyPasta typo in description?  If not, is there a way to clarify what is going on for my "new" eyes?

__tests__/components/editor/ValueConstraints.test.js
```diff
@@ -0,0 +1,93 @@
...
+    expect(wrapper.find('UseValuesFrom > div > ul > li').text()).toEqual('http://id.loc.gov/foo/works')
+  })
+})
+
+describe('<ValueDataTypes />', () => {
+  it('has valueTemplateRefs', () => {
+    expect(wrapper.find('ValueDataTypes > div > ul > li').text()).toEqual('data type URI: http://id.loc.gov/ontologies/bibframe/Instance')
+  })
```
ndushay: given the logic added, I would like to see:
- a test showing the defaults are always rendered (?)
- a test showing that it does NOT render anything when a constraint is missing.
- a test showing a remark or DTU (the two properties in the ValueConstraints component itself)

Also, I'm trying to mentally map the tests to the props and/or the code and I'm confused.  Is this because it's confusing?  Would we want to clarify which props are output by which components?   Or anything along those lines?

jgreben:
screen shot 2018-11-13 at 3 42 55 pm

__tests__/components/editor/ValueConstraints.test.js
jgreben: Cp typo...👌

src/components/editor/PropertyTemplate.jsx
jgreben: 👌

__tests__/components/editor/ValueConstraints.test.js
jgreben: Cp typo 👌

__tests__/components/editor/ValueConstraints.test.js
jgreben: 👌

jgreben:
I'm now wondering to what extent we want to abstract the parsing of profiles so that we are not dependent on the specific "profile editor" schema? Right now we are setting the props to be the specific profile name/values. Should we just loop through all of the name/values and display them? FWIW, the BFW does not actually do this either. It also looks for specific name/values from the expected profile.

src/components/editor/PropertyTemplate.jsx
jgreben: There was no need to send all of them. Now I am just passing `{...this.props.propertyTemplates[0].valueConstraint}`

__tests__/components/editor/ValueConstraints.test.js
jgreben: I have a fuller test suite now. I am clearing out the wrapper of all values to show that the child nodes are not rendered, as well as adding in tests for missing json.

PR #150, opened 2018-11-13
Adds a placeholder starting points menu

### LINKED ISSUE #147: create "Starting Points" component as placeholder
Create a component "StartingPoints" which is the left nav "Create Resources" part of the existing BFE.  In our case, it's just a placeholder, or maybe there's a link to "monograph instance" only and a "select a resource template" sort of button.

I'm just making that up, as it's meant as a placeholder, with a "Starting Points" heading visible.

jgreben:
Fixes #147

### DIFF
--- a/src/components/editor/Editor.jsx
+++ b/src/components/editor/Editor.jsx
@@ -1,6 +1,7 @@
 import React, {Component} from 'react'
 import ResourceTemplate from './ResourceTemplate'
 import EditorHeader from '../EditorHeader'
+import StartingPoints from './StartingPoints'
 
 class Editor extends Component {
   constructor(props) {
@@ -122,12 +123,13 @@ class Editor extends Component {
       <div id="editor">
         <EditorHeader triggerEditorMenu={this.props.triggerHandleOffsetMenu}/>
         <h1> Editor Page </h1>
-        <p>(will have a "Choose Resource Template" area + editing area)</p>
         <p>The selected resource template is <strong>{this.state.resourceTemplates[0].id}</strong></p>
+        <StartingPoints/>
         <ResourceTemplate
           resourceTemplates = {this.state.resourceTemplates}
           />
       </div>
+
     )
   }
 }
--- a/src/components/editor/ResourceTemplate.jsx
+++ b/src/components/editor/ResourceTemplate.jsx
@@ -12,9 +12,12 @@ class ResourceTemplate extends Component {
       border: '1px dotted',
       padding: '10px'
     }
-
+    let float = {
+      float: 'left',
+      width:'75%'
+    }
     return (
-      <div className='ResourceTemplate' style={dashedBorder}>
+      <div className='ResourceTemplate' style={Object.assign(dashedBorder, float)}>
         <h3>Resource Template Container </h3>
         <p>Resource Template selected:</p>
         <ul>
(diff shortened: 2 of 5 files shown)


src/styles/main.css
```diff
@@ -31,6 +31,7 @@ body {
   font-size: 13.5px;
 }
 #footer {
+  float: right;
```
jermnelson: Is there reason why the footer is floating right? I wasn't able to see any difference in the appearance of the footer when this css was removed and when it was present.

jgreben: It has to do with when the browser window is made narrower, otherwise it bumps to the right side of the page.

jgreben: @jermnelson see comment above, but it does looks like there is some further CSS refinement that needs to happen. Not sure if this is the PR for that given that we will be doing work on the form later on.

src/components/editor/Editor.jsx
```diff
@@ -118,16 +119,25 @@ class Editor extends Component {
...
+    let startingPoints = {
+      border: '1px dotted',
+      float: 'left',
+      padding: '20px'
+    }
     return(
       <div id="editor">
         <EditorHeader triggerEditorMenu={this.props.triggerHandleOffsetMenu}/>
         <h1> Editor Page </h1>
         <p>(will have a "Choose Resource Template" area + editing area)</p>
```
ndushay: don't need this line anymore, correct?

src/components/editor/ResourceTemplate.jsx
```diff
@@ -12,22 +12,27 @@ class ResourceTemplate extends Component {
...
-          <li>resourceURI: <strong>{this.props.resourceTemplates[0].resourceURI}</strong></li>
-          <li>id: <strong>{this.props.resourceTemplates[0].id}</strong></li>
-          <li>remark: <strong>{this.props.resourceTemplates[0].remark}</strong></li>
-        </ul>
-        <form id="resourceTemplate" className="form-horizontal" role="form" style={dottedBorder}>
-          <h4>BEGINNING OF FORM</h4>
-          <PropertiesWrapper propertyTemplates = {[this.props.resourceTemplates[0].propertyTemplates]} />
-          <h4>END OF FORM</h4>
-        </form>
+      <div>
```
ndushay: why do we need another div?

src/components/editor/StartingPoints.jsx
```diff
@@ -0,0 +1,18 @@
+import React, { Component }  from 'react'
+
+class StartingPoints extends Component {
+  render() {
+    return (
+      <div>
```
ndushay: no identifying class or id for this div?

src/styles/main.css
SaravShah: hmmm @jgreben i'm not seeing the problem when the browser is made narrower. Could you show me whenever you have time?

SaravShah: ^saw when webpage was enbiggened or however you spell it

src/components/editor/Editor.jsx
jgreben: 👍

src/styles/main.css
jgreben: 👌

src/components/editor/ResourceTemplate.jsx
jgreben: 🔨 Fixed.

PR #154, opened 2018-11-15
Fixed 404 when GET editor directly or refreshing page

### LINKED ISSUE #153: Directly loading editor route causes error
With a running the Sinopia Editor, the `editor` route works only if the homepage is loaded first. Directly loading the `editor` route returns an Error page **Cannot GET /editor**

jermnelson:
Fixes #153

### DIFF
--- a/server.js
+++ b/server.js
@@ -11,6 +11,7 @@ app.listen(port);
 
 const versoSpoof = require('./src/versoSpoof.js')
 
+
 app.all("/verso/api/configs", function (req, res, next) {
   if (req.query.filter.where.configType === 'profile') {
     res.json(versoSpoof.profiles)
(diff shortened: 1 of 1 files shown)


server.js
```diff
@@ -7,10 +7,12 @@ var express = require('express');
 var app = express();
 
 app.use(express.static(__dirname + '/'));
+console.debug(`Dirname is ${__dirname}`)
```
SaravShah: could probably get rid of this 🐛

jermnelson: Thanks for the catch! Now removed from code.

ndushay:
You left in the blank line, but I merged it.  Let's not stop progress for a blank line!

PR #155, opened 2018-11-15
Minor improvements I stumbled into while working on sinopia server spoofing

ndushay:
WIP - i think some of the "if" statements are rendering ...

- package scripts to compile without optimizing and otherwise speed up dev cycle
- README updated to reflect ^^ and also be clearer about diff between running webpack and running express
- some error handling on editor react component rendering

### DIFF
--- a/src/components/editor/PropertyTemplate.jsx
+++ b/src/components/editor/PropertyTemplate.jsx
@@ -7,23 +7,29 @@ class PropertyTemplate extends Component {
       border: '1px dotted',
       padding: '10px'
     }
-    return (
-      <div className='PropertyTemplate' style={dottedBorder}>
-        <h5>BEGIN PropertyTemplate elements</h5>
-        <ul>
-          <li>propertyLabel: <strong>{this.props.propertyTemplates[0].propertyLabel}</strong></li>
-          <li>propertyURI: <strong>{this.props.propertyTemplates[0].propertyURI}</strong></li>
-          <li>type: <strong>{this.props.propertyTemplates[0].type}</strong></li>
-          <li>mandatory: <strong>{this.props.propertyTemplates[0].mandatory}</strong></li>
-          <li>repeatable: <strong>{this.props.propertyTemplates[0].repeatable}</strong></li>
-          <li>contained resourceTemplates[] length: <strong>{this.props.propertyTemplates[0].resourceTemplates.length}</strong></li>
-          <li>
-            <ValueConstraints {...this.props.propertyTemplates[0].valueConstraint} />
-          </li>
-        </ul>
-        <h5>END of PropertyTemplate elements</h5>
-      </div>
-    )
+    if (this.props.propertyTemplates.length == 0) {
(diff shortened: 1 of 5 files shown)


ndushay:
@SaravShah and @jermnelson  would be great to get your feedback on the tweaks to the webpack start scripts ...

src/components/editor/PropertyTemplate.jsx
```diff
@@ -7,23 +7,29 @@ class PropertyTemplate extends Component {
...
-          <li>
-            <ValueConstraints {...this.props.propertyTemplates[0].valueConstraint} />
-          </li>
-        </ul>
-        <h5>END of PropertyTemplate elements</h5>
-      </div>
-    )
+    if (this.props.propertyTemplates.length == 0) {
+      return <h1>We should have propertyTemplates but we don't.</h1>
+    } else if (this.props.propertyTemplates[0].resourceTemplates == undefined ) {
```
jgreben: Can we just use `if(!this.props.propertyTemplates)` here for both `length == 0` and `'undefined'` ?

src/components/editor/ResourceTemplate.jsx
```diff
@@ -39,17 +39,21 @@ class PropertiesWrapper extends Component{
...
-        <p>BEGIN PropertiesWrapper</p>
-          <div>
-            {this.props.propertyTemplates[0].map(function(pt, index){
-              return <PropertyTemplate key={index} propertyTemplates={[pt]}/>
-            })}
-          </div>
-        <p>END PropertiesWrapper</p>
-      </div>
-    )
+    if (this.props.propertyTemplates.length == 0 || this.props.propertyTemplates[0] == {}) {
```
jgreben: Maybe we need a global `isEmpty()` function. This is something I used in the ValueConstraints component and it may have broader use. See https://coderwall.com/p/_g3x9q/how-to-check-if-javascript-object-is-empty

src/components/editor/PropertyTemplate.jsx
ndushay: didn't think of it - good idea!

src/components/editor/ResourceTemplate.jsx
ndushay: Interesting idea.  I wonder what @jermnelson thinks.

src/components/editor/PropertyTemplate.jsx
ndushay: @jgreben and I discussed F2F;  plan is to use #158 ultimately

src/components/editor/ResourceTemplate.jsx
ndushay: @jgreben and I discussed F2F; plan is to use #158 ultimately

src/components/editor/PropertyTemplate.jsx
jgreben: 👌

PR #156, opened 2018-11-15
Adds tests for rendering of PropertiesWrapper component

### LINKED ISSUE #132: add tests for PropertiesWrapper
May need to split it out into separate file, or deal with exports so the component is available for testing.

jgreben:
Fixes #132


ndushay:
Thanks, Josh!

Issue #175, opened 2018-11-16
props validation - we should be doing it??


ndushay:
In properly hooking up eslinter to the sinopia editor, we have a few errors flagged re props validation:

```
sinopia_editor/src/components/CanvasMenu.jsx
  13:41  error    'closeHandleMenu' is missing in props validation      react/prop-types
/Users/ndushay/ld4p/sinopia_editor/src/components/editor/Editor.jsx
  46:53  error    'triggerHandleOffsetMenu' is missing in props validation  react/prop-types

sinopia_editor/src/components/editor/ResourceTemplate.jsx
  26:50  error  'resourceTemplates' is missing in props validation                      react/prop-types
  26:71  error  'resourceTemplates[].resourceLabel' is missing in props validation      react/prop-types
  27:69  error  'resourceTemplates[].resourceURI' is missing in props validation        react/prop-types
  28:60  error  'resourceTemplates[].id' is missing in props validation                 react/prop-types
  29:64  error  'resourceTemplates[].remark' is missing in props validation             react/prop-types
  33:84  error  'resourceTemplates[].propertyTemplates' is missing in props validation  react/prop-types
  47:20  error  'propertyTemplates' is missing in props validation                      react/prop-types
  47:38  error  'propertyTemplates.length' is missing in props validation               react/prop-types
  54:48  error  'propertyTemplates[].map' is missing in props validation                react/prop-types

sinopia_editor/src/components/editor/ValueConstraints.jsx
   12:36  error    'defaults' is missing in props validation                  react/prop-types
   12:45  error    'defaults.length' is missing in props validation           react/prop-types
   13:43  error    'editable' is missing in props validation                  react/prop-types
   14:48  error    'languageLabel' is missing in props validation             react/prop-types
   15:46  error    'languageURI' is missing in props validation               react/prop-types
   16:41  error    'remark' is missing in props validation                    react/prop-types
   17:45  error    'repeatable' is missing in props validation                react/prop-types
   18:41  error    'useValuesFrom' is missing in props validation             react/prop-types
   18:55  error    'useValuesFrom.length' is missing in props validation      react/prop-types
   19:50  error    'validatePattern' is missing in props validation           react/prop-types
   20:52  error    'valueDataType' is missing in props validation             react/prop-types
   21:48  error    'valueLanguage' is missing in props validation             react/prop-types
   22:45  error    'valueTemplateRefs' is missing in props validation         react/prop-types
[...]

Issue #174, opened 2018-11-16
target="_blank" uses need more attention


ndushay:
In properly hooking up eslinter to the sinopia editor, we have a few errors flagged re: target="_blank"

```
sinopia_editor/src/components/CanvasMenu.jsx
  52:79  error    Using target="_blank" without rel="noopener noreferrer" is a security risk: see https://mathiasbynens.github.io/rel-noopener                                                                                                                                                                                                                                      react/jsx-no-target-blank
  53:57  error    Using target="_blank" without rel="noopener noreferrer" is a security risk: ...
  54:50  error    Using target="_blank" without rel="noopener noreferrer" is a security risk: ...
  55:55  error    Using target="_blank" without rel="noopener noreferrer" is a security risk: ...
  56:47  error    Using target="_blank" without rel="noopener noreferrer" is a security risk: ...
  57:46  error    Using target="_blank" without rel="noopener noreferrer" is a security risk: ...
  58:66  error    Using target="_blank" without rel="noopener noreferrer" is a security risk: ...
  59:50  error    Using target="_blank" without rel="noopener noreferrer" is a security risk: ...                                                                                                                                                             
  60:73  error    Using target="_blank" without rel="noopener noreferrer" is a security risk: ...                                                                                                                                                             
  61:69  error    Using target="_blank" without rel="noopener noreferrer" is a security risk: ...
  62:89  error    Using target="_blank" without rel="noopener noreferrer" is a security risk: ...
  63:43  error    Using target="_blank" without rel="noopener noreferrer" is a security risk: ...
  64:74  error    Using target="_blank" without rel="noopener noreferrer" is a security risk: ...

/Users/ndushay/ld4p/sinopia_editor/src/components/Footer.jsx
   8:106  error  Using target="_blank" without rel="noopener noreferrer" is a security risk: ...
  15:75   error  Using target="_blank" without rel="noopener noreferrer" is a security risk: ...
  16:177  error  Using target="_blank" without rel="noopener noreferrer" is a security risk: ...

/Users/ndushay/ld4p/sinopia_editor/src/components/NewsItem.jsx
  9:179  error  Using target="_blank" without rel="noopener noreferrer" is a security risk: ...

Issue #173, opened 2018-11-16
design issue?  association of labels with fields


ndushay:
In hooking up the a11y linter to the sinopia editor, we have a few errors flagged re: association of labels with fields.

```
sinopia_editor/src/components/LoginPanel.jsx
   9:7  error    Form label must have ALL of the following types of associated control: nesting, id                                                                                                                                                                                                                                                                                jsx-a11y/label-has-for
  13:7  error    Form label must have ALL of the following types of associated control: nesting, id                                                                                                            
```

In researching the fix, I found this:

- https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/label-has-for.md#rule-details
    -  "To fully cover 100% of assistive devices, you're encouraged to validate for both nesting and id."
    -  https://github.com/evcohen/eslint-plugin-jsx-a11y/issues/455#issuecomment-403105932
        -  "use label-has-associated-control instead of label-has-for. We deprecated label-has-for in v6.1.0."

However, since the best solution is to use BOTH nesting and id, following their advice and using this instead:
- https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/label-has-associated-control.md 

only checks for the htmlFor tag associated with an id.

Do we care?   I tried to apply the nesting and it messed up the display ... so perhaps a conversation is in order?

SaravShah:
Additional failure: `/Users/saravs/Desktop/dlss/sinopia_editor/src/components/editor/InputLiteral.jsx
  7:9   error    Form label must have ALL of the following types of associated control: nesting, id`

astridu:
I'm not sure, but I found this. Someone said the 'every' needs to change to 'some'

"required": {
                "some": [ "nesting", "id" ]
            }

https://github.com/evcohen/eslint-plugin-jsx-a11y/issues/302
If that doesn't help, I'll ask Jennifer if she knows.

ndushay:
https://ucomm.stanford.edu/policies/accessibility-policy/

ndushay:
See https://github.com/evcohen/eslint-plugin-jsx-a11y/issues/457#issuecomment-450724848

```
    "jsx-a11y/label-has-associated-control": [ 2, {
    "labelComponents": ["label"],
    "labelAttributes": ["htmlFor"],
    "controlComponents": ["input"]
  }]
```

Issue #172, opened 2018-11-16
accessibility linter warnings


ndushay:
I hooked up an a11y linter to the sinopia editor.  It gives us a bunch of errors, in a couple flavors.  I'm not sure if this is about how we implemented stuff in the UI, or if there's a design issue ...

```
...
sinopia_editor/src/components/CanvasMenu.jsx
 13:9   error    Anchor used as a button. Anchors are primarily expected to navigate. Use the button element instead. Learn more: https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-is-valid.md                                                                                                                                                      jsx-a11y/anchor-is-valid
...
  22:85  error    The href attribute requires a valid value to be accessible. Provide a valid, navigable address as the href value. If you cannot provide a valid href, but still need the element to resemble a link, use a button and change it with appropriate styles. Learn more: https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-is-valid.md
  27:85  error     The href attribute requires a valid value to be accessible. ...
  29:15  error     The href attribute requires a valid value to be accessible. ...
  34:85  error     The href attribute requires a valid value to be accessible. ...
  36:15  error     The href attribute requires a valid value to be accessible. ...
  37:15  error     The href attribute requires a valid value to be accessible. ...
  38:15  error     The href attribute requires a valid value to be accessible. ...
  42:85  error     The href attribute requires a valid value to be accessible. ...
  44:15  error     The href attribute requires a valid value to be accessible. ...
  45:15  error     The href attribute requires a valid value to be accessible. ...
  46:15  error     The href attribute requires a valid value to be accessible. ...
  50:85  error    The href attribute requires a valid value to be accessible. ...

sinopia_editor/src/components/EditorHeader.jsx
   20:15  error    Anchor used as a button. Anchors are primarily expected to navigate. Use the button element instead. Learn more: https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-is-valid.md  jsx-a11y/anchor-is-valid

sinopia_editor/src/components/EditorHeader.jsx
  20:15  error    Anchor used as a button. Anchors are primarily expected to navigate. Use the button element instead. Learn more: https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-is-valid.md  jsx-a11y/anchor-is-valid

[...]

PR #176, opened 2018-11-16
make eslint more useful

ndushay:
I had eslint ignore the BFE code, and then I fixed our eslint configuration to include jsx files and use the plugins that we've npm installed, specifically jsx-a11y (accessibility) and security ... and react, now that jsx files are in play.   Even with the new configuration, our warnings went down to 86 or so, so I dialed that max down for our CI builds.

Next I addressed some simple a11y rule violations, and turned off the rule that cares about semicolons (not worth our time to be 100% conformant).

This PR has many commits, but they are all small.  I recommend reviewing by commit.

Note that THESE eslint rules tell us useful stuff in the warnings.   It caused me to create 4 tickets:  #172, #173, #174, #175 -- each is referenced by a bespoke rule configuration in .eslintrc.js

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -2,11 +2,10 @@ build_support/*
 builds/*
 coverage/*
 docs/*
-static/css/*
-static/images/*
-static/js/jsonld-vis.js
-static/js/n3-browser.min.js
-static/js/r*
-static/js/short-uuid.min.js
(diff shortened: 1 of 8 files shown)


SaravShah:
LGTM. 👍

PR #184, opened 2018-11-19
SinopiaServerSpoof: exports getResourceTemplate so it can be used anywhere

ndushay:
It turns out that the work to render individual form elements and the work to create modals both can benefit from making `getResourceTemplate` an importable function.  This enables that.  To use:

```javascript
const { getResourceTemplate } = require('../../sinopiaServerSpoof.js')
...
// Use how you wish:
    this.state = { resourceTemplates: [getResourceTemplate('foo')]}
```


ndushay:
I changed the coveralls minimum for passing to be 82%.  We may have to keep tweaking for a while.  I wouldn't worry about that.

PR #199, opened 2018-11-26
Adds form select options from spoofed target list

### LINKED ISSUE #189: Lookup field using QA
The lookup field 
 - [x] Should query QA (requires #188)
 - [x] Display results
-  [x] Allow the user to populate the field based on the results.

### LINKED ISSUE #192: Input field with values from target URI
Fixed by #199

For properties with `"type": "target"` the target is just a lookup to a constrained list of values (as opposed to something that uses a typeahead using QA)

For this we will want to fetch the values from the "useValuesFrom" valueConstraint, and parse the json fetched from the given URI, e.g. (http://id.loc.gov/vocabulary/frequencies.json):

```
"propertyTemplates": [
                    {
                        "mandatory": "false",
                        "repeatable": "true",
                        "type": "target",
                        "resourceTemplates": [],
                        "valueConstraint": {
                            "valueTemplateRefs": [],
                            "useValuesFrom": [
                                "http://id.loc.gov/vocabulary/frequencies"
                            ],
                            "valueDataType": {
                                "dataTypeURI": "http://id.loc.gov/ontologies/bibframe/Frequency"
                            },
                            "defaults": []
                        },
                        "propertyURI": "http://id.loc.gov/ontologies/bflc/target",
                        "propertyLabel": "Frequency (RDA 2.14)",
                        "remark": "http://access.rdatoolkit.org/2.14.html"
                    }
```

jgreben:
Fixes #189 
Fixes #192 

- Adds an option select component for a target list of input options. Uses the react-bootstrap-typeahead component with options selected from a fetch to a spoofed serial frequencies.json file.

- Going forward, we will need a way to be able to dynamically select a given resource template as I had to hard-code the resourceTemplate id for this PR.

- It is not clear what the exact JSON model will be returned from the QA server. For now, I just hand-coded the most simple json for this PR, but if QA uses the same model as they do for a name or subject query, then nothing will need to change as the typeahead component will easily recognize this format, e.g. (from lookup.ld4l.org): 

```
[
   {"uri":"http://id.loc.gov/authorities/subjects/sh85118553",
   "id":"http://id.loc.gov/authorities/subjects/sh85118553","
    label":"Science"}
]
```


.eslintrc.js
```diff
@@ -44,7 +44,8 @@ module.exports = {
     "jsx-a11y/label-has-for": "warn", // see #173
     "no-console": "warn",
     "no-extra-semi": "off", // because it isn't that important
-    "react/prop-types": "warn" // see #175
+    "react/prop-types": "warn", // see #175
+    "jsx-a11y/no-onchange": "warn"
```
ndushay: did you look at this? https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/no-onchange.md (I did a google search on "jsx-a11y/no-onchang"). Will adding or substituting onBlur be acceptable here, instead of adding a rule exception?

__tests__/components/editor/InputTarget.test.js
```diff
@@ -0,0 +1,50 @@
...
+          "dataTypeURI": "http://id.loc.gov/ontologies/bibframe/Frequency"
+        }
+      }
+    }
+}
+
+describe('<InputTarget />', () => {
+  const wrapper = shallow(<InputTarget {...plProps} />)
+
+  it('contains a label with "Instance of"', () => {
```
ndushay: copyPasta error.  best to possibly say "value of propertyLabel" or even actually have test compare with `parsedJSON.propertyLabel`

__tests__/components/editor/InputTarget.test.js
```diff
@@ -0,0 +1,50 @@
...
+    expect(wrapper.find('label').text()).toMatch('Frequency (RDA 2.14)')
+  })
+
+  it('typeahead component should have a placeholder attribute with value propertyLabel', () => {
+    expect(wrapper.find('#targetComponent').props().placeholder).toBe('Frequency (RDA 2.14)')
+  })
+
+  it('sets the typeahead component required attribute', () => {
```
ndushay: have test compare with parsedJSON.mandatory and/or add that to the description.

__tests__/components/editor/InputTarget.test.js
```diff
@@ -0,0 +1,50 @@
...
+    expect(wrapper.find('#targetComponent').props().required).toBe(false)
+  })
+
+  it('sets the typeahead component multiple attribute', () => {
```
ndushay: have test compare with parsedJSON.repeatable and/or add that to the description.

__tests__/components/editor/InputTarget.test.js
```diff
@@ -0,0 +1,50 @@
...
+    expect(wrapper.find('#targetComponent').props().multiple).toBe(true)
+  })
+
+  it('sets the typeahead component placeholder attribute', () => {
```
ndushay: have test compare with parsedJSON.propertyLabel and/or add that to the description.

src/sinopiaServerSpoof.js
```diff
@@ -5,23 +5,27 @@ const monographInstanceRt = require('../static/spoofedFilesFromServer/fromSinopi
...
   {'id': 'resourceTemplate:bf2:ParallelTitle', 'json': parallelTitleRt},
+  {'id': 'resourceTemplate:bf2:Serial:Frequency', 'json': serialRt},
   {'id': 'resourceTemplate:bf2:Title', 'json': titleRt},
   {'id': 'resourceTemplate:bf2:Title:Note', 'json': titleNoteRt},
   {'id': 'resourceTemplate:bflc:TranscribedTitle', 'json': transcribedTitleRt},
   {'id': 'resourceTemplate:bf2:Title:VarTitle', 'json': varTitleRt},
   {'id': 'resourceTemplate:bf2:WorkTitle', 'json': workTitleRt},
-  {'id': 'resourceTemplate:bf2:WorkVariantTitle', 'json': workVariantTitleRt}
+  {'id': 'resourceTemplate:bf2:WorkVariantTitle', 'json': workVariantTitleRt},
+  {'id': 'vocabulary:bf2:frequencies', 'json': qaLocVocabFrequencies}
```
ndushay: ok, I wouldn't include the qaLocVocabFrequencies here.  I would make a separate spoof for QA or better yet, mock it for the tests.   The code can do a real live call;  it's only the tests that shouldn't correct?

Plus, it is putting something that is NOT a resourceTemplate in a var called  resourceTemplateId2Json.   If you need to spoof it, rather than mock, i would recommend you create a qaServerSpoof.js  instead.   

If i came to this code like this in 6 months, I'd be ... where's the qaSpoof?  What?  It's in the sinopiaServerSpoof?  The sinopiaServer is not holding qa data.  And why the heck is it included as a resourceTemplateId?

src/components/editor/InputTarget.jsx
```diff
@@ -0,0 +1,65 @@
...
+
+  render() {
+    const target_url = './static/spoofedFilesFromServer/fromQA/frequencies.json'
+
+    return (
+      <div>
+        <label htmlFor="targetComponent">{this.props.propertyTemplate.propertyLabel}
+        <Typeahead
+          id="targetComponent"
+          required={JSON.parse(this.props.propertyTemplate.mandatory)}
```
ndushay: ditto comment in #209 about only parsing the JSON once.

src/components/editor/InputTarget.jsx
```diff
@@ -0,0 +1,65 @@
...
+          placeholder={this.props.propertyTemplate.propertyLabel}
+          multiple={JSON.parse(this.props.propertyTemplate.repeatable)}
+          useCache={true}
```
ndushay: same "ew" on having to put `true` in {}.  Do we really have to?  (this is not only spot, so I'm guessing no?)

.eslintrc.js
jgreben: 👌 It's not throwing an error anymore, so I removed the rule.

__tests__/components/editor/InputTarget.test.js
jgreben: 👌

src/components/editor/InputTarget.jsx
```diff
@@ -0,0 +1,65 @@
...
+    super(props)
+    this.state = {
+      isLoading: true,
+      options: [],
+      selected: []
+    }
+  }
+
+  render() {
+    const target_url = './static/spoofedFilesFromServer/fromQA/frequencies.json'
```
ndushay: i think one of the points of the way I did spoofing is that it can get the URL it needs directly from the information in the resourceTemplate.   You wouldn't find "vocabulary:bf2:frequencies" in the resourceTemplate (which is how it's spoofed) -- you find the whole URL, no?   And something about that URL tells you to go to QA ... and THAT is what you need code for -- to translate what is in the resource template to getting the correct query to QA, with the help of generated javascript code from swagger client.

I think what we have here is actually an example of how we can use a Typeahead component ... which is something we need ... but is not the whole problem.  We need tickets for the rest of the work, no?

__tests__/components/editor/InputTarget.test.js
jgreben: 👌

__tests__/components/editor/InputTarget.test.js
jgreben: 👌

__tests__/components/editor/InputTarget.test.js
jgreben: 👌

src/sinopiaServerSpoof.js
jgreben: 👌

src/components/editor/InputTarget.jsx
jgreben: 👌 Nothing to change here right now? If I'm wrong, let's talk. I made ticket #226 and removed the QA spoof altogether. If we want a QA spoof that can be a separate ticket, but we should discuss whether we even need that fore the time being.

Issue #211, opened 2018-11-29
Add isRequired and change PropType to React Component props


jermnelson:
For React components in the editor, 
- [ ] Determine what props should be required and set using `isRequired` to props along with accompanying tests. 
- [ ] For arrays, use `.arrayOf()` to be more specific
- [ ] Investigate using `PropTypes.shape` for objects such as propertyTemplate or resourceTemplate props. (and even targets types of arrayOf, as in ModalToggle, etc.

(see comments in PR #193)

Issue #212, opened 2018-11-29
Input Literal Integration test


SaravShah:
work on integration test for input literal. Here is what naomi put about the tests: BFF coders:  we should use puppeteer and jest for integration tests - Jeremy and I both didn't find there was a better way

enzyme is great for unit tests, but not integration tests
and we need to start thinking of user interaction with what our PRs are doing and test those user interactions as integration tests.
here's some links I thought were interesting talking about testing in our context:
https://medium.com/@newyork.anthonyng/testing-react-components-state-b57bfc712b90
https://stackoverflow.com/questions/40725536/testing-changes-to-react-component-state-and-spying-on-instance-methods-using-en

Issue #208, opened 2018-11-14
Add Missing Integration Tests for existing React code for BFF


jgreben:
Basic tests to ensure that aggregate components get rendered with expected data
Per #115, use Puppeteer.

(e.g. we load a resource template in the editor, and the Editor, ResourceTemplate, PropertyWrapper, Property, PropertyValueConstraint components (etc) all have data from the selected resource template)

e.g.
- [ ] Editor page has sinopia header
- [ ] Editor page has sinopia footer
future
- [ ] clicking in starting points menu brings up appropriate resource template form
- [ ] resource template form has appropriate fields and buttons
- [ ] modal form with lookup does the right thing
- [ ] modal form shows text fields appropriately
etc. ?

Perhaps similar to our list of original integration tests for the BFE?

justinlittman:
Closing in favor of more specific testing tickets.

PR #216, opened 2018-11-30
Add modals for properties with valueTemplateRefs 

ndushay:
To review:
- take a careful look at existing components I *changed*:
    - ResourceTemplate  (replaced FormWrapper with ResourceTemplateForm, improved tests)
    - sinopiaServerSpoof  (added barcode rt to quiet console errors during testing)
    - MonographInstance (changed barcode rtId prefix)
- look at the new classes carefully
     - ResourceTemplateForm renders
     - ModalToggle renders
     - ResourceTemplateModal 
- think about whether I am testing the new classes sufficiently

If there's objection to the removal of unused components, I can take those commits out of this PR.

Known problems/todos:
- integration tests (#223)
- need to accommodate these this:  `repeatable` (#217), `mandatory` (#218), `editable` and `defaultValue` (#219 for both)
- use propTypes.shape for more complex objects in arrays (part of #211)
- ~~ensure valueTemplateRefs are always resourceTemplate ids (and not vocabularies, etc) (#220)~~
- holding the data from a ResourceTemplateModal  (once we decide how we will be modelling that:  redux or ...) (#221)
- styling of buttons, etc:  (#222)

Editor spins up like this:

image

and if you click a button (e.g. "Title"), you get something like this:

image

and if you click the "Title note" button, you get this:

image

Accomplished by replacing FormWrapper component with ResourceTemplateForm component, which is reused in the Modals.


src/components/editor/ResourceTemplateModal.jsx
```diff
@@ -0,0 +1,57 @@
...
+    this.props.toggleVisibility()
+  }
+
+  handleSave = () => {
+    this.saveDataFromModal()
+    this.handleClose()
+  }
+
+  saveDataFromModal = () => {
+    console.log('TODO: implement holding data from ResourceTemplateModal')
```
SaravShah: Should we create a ticket # for this?

src/components/editor/ModalToggle.jsx
```diff
@@ -0,0 +1,51 @@
...
+    this.toggleModal = this.toggleModal.bind(this)
+    this.state = {
+      visible: false
+    }
+  }
+
+  toggleModal = () => {
+    const prevState = this.state
+    prevState.visible = !prevState.visible
+    this.setState(prevState => (prevState))
```
SaravShah: i think you could also do this if you want: 

```
this.setState({
    visible: !this.state.visible
})
```

__tests__/components/editor/ResourceTemplateForm.test.js
```diff
@@ -0,0 +1,94 @@
...
+    {
+      "propertyLabel": "Literally",
+      "type": "literal"
+    },
+    {
+      "propertyLabel": "Look up, look down",
+      "type": "lookup"
+    },
+    {
+      "propertyLabel": "Bullseye!",
```
SaravShah: loving the `propertyLabels`

src/components/editor/ModalToggle.jsx
ndushay: I tried some stuff;   dunno if I tried this.

static/spoofedFilesFromServer/fromSinopiaServer/resourceTemplates/MonographInstance.json
```diff
@@ -175,7 +175,7 @@
         },
```
jermnelson: This is not specific to this code, but I realized another source we should use for determining such variables as `repeatable` and `mandatory` is the Library of Congress Bibframe documentation at https://www.loc.gov/bibframe/. For example, in this BIBFRAME file the 'instanceOf` propertyTemplate has false for repeatable, but I as understand BIBFRAME 2.0 ontology, at this level of modeling, Instance should have only one Work, while Works have multiple Instances.

src/components/editor/ResourceTemplateModal.jsx
jermnelson: Yeah, I think this is a candidate for using Redux.

__tests__/components/editor/ResourceTemplateForm.test.js
```diff
@@ -0,0 +1,94 @@
...
+
+  describe('buttons for resource template modals (b/c we have a property of type "resource" with correct valueTemplateRefs)', () => {
+    it('renders ButtonToolbar containing ModalToggle(s)', () => {
+      expect(wrapper
+        .find('div.ResourceTemplateForm ButtonToolbar ModalToggle').length)
+        .toEqual(2)
+    })
+    it('passes appropriate props to ModalToggle(s)', () => {
+      wrapper.find('ModalToggle').forEach((node) => {
+        expect(node.prop('buttonLabel')).toEqual('Note') // TODO: ensure we get from spoofed resourceTemplate
```
jgreben: Maybe say a bit more about this TODO or issue#?

ndushay:
Josh, et al:

yes, sorry about the length.  At least I removed as many files as I added - heh.

I made a separate ticket for the integration tests(#223);  hopefully this is one of the largest PRs we'll have to do for a long time.  (cough cough).   I think whenever possible we should include integration tests with the PR ... if we think the UI is settled down enough.  I already made a ticket for the "backlog" of missing integration tests (#208): and Sarav made a ticket for the InputLiteral (#212), so ... ahem.

 :-)

__tests__/components/editor/ResourceTemplateForm.test.js
ndushay: haha - I have no idea what I was thinking when I wrote that TODO comment.  I will remove it.

PR #227, opened 2018-11-30
Adds form select options from spoofed target list (again)

jgreben:
@ndushay new PR per your comments on https://github.com/LD4P/sinopia_editor/pull/199

PR #229, opened 2018-12-01
Adds input form select options from spoofed resource list 

jgreben:
After rebasing from master decided to change property type to resource instead of target.

See https://github.com/LD4P/sinopia_editor/pull/227 for previous comments and review.


__tests__/components/editor/ResourceTemplateForm.test.js
```diff
@@ -18,8 +18,13 @@ const rtProps = {
       "type": "lookup"
     },
     {
-      "propertyLabel": "Bullseye!",
-      "type": "target"
+      "propertyLabel": "What's the frequency Kenneth?",
```
jermnelson: Nice reference!

PR #254, opened 2018-12-17
Remove item from redux

### LINKED ISSUE #245: Input type literal when you click on the X remove item from the global redux store
-Need to pass in ID and item to the reducer, have the reducer filter out the item needed to be removed and return new state

SaravShah:
closes #245 

screen shot 2018-12-17 at 3 21 11 pm

When clicking on the X:

screen shot 2018-12-17 at 3 21 15 pm

### DIFF
--- a/__tests__/reducers/literal.test.js
+++ b/__tests__/reducers/literal.test.js
@@ -1,6 +1,6 @@
 import literal from '../../src/reducers/literal'
 
-describe('todos reducer', () => {
+describe('literal reducer', () => {
   it('should handle initial state', () => {
     expect(
       literal(undefined, {})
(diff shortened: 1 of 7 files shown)


__tests__/reducers/literal.test.js
```diff
@@ -1,6 +1,6 @@
 import literal from '../../src/reducers/literal'
 
-describe('todos reducer', () => {
+describe('literal reducer', () => {
```
jermnelson: Good catch!

src/components/editor/InputLiteral.jsx
```diff
@@ -68,9 +69,15 @@ export class InputLiteral extends Component {
   }
 
   handleClick(event) {
+    // console.warn(event.target.dataset["label"])
```
jermnelson: Should probably just delete lines 72 and 73

PR #260, opened 2018-12-18
Input list redux

### LINKED ISSUE #259: Use Redux for InputList component

jgreben:
Fixes #259

- Combined reducer for list and literal components
- Changes InputList component to dispatch state to reducer action
- Tests for modified component and list reducer
- bumps up max-warnings to allow for security/detect-object-injection, but this is safe because user cannot inject anything in place of `var i` needed to parse the json to create the selection list

### DIFF
--- a/src/components/editor/InputList.jsx
+++ b/src/components/editor/InputList.jsx
@@ -49,9 +49,9 @@ class
               .then(json => {
                 for(var i in json){
                   try{
-                    // warns about security/detect-object-injection, but is safe because user cannot inject anything for `var i`
-                    const uri = json[i]["@id"]
-                    const label = json[i]["http://www.loc.gov/mads/rdf/v1#authoritativeLabel"][0]["@value"]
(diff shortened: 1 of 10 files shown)


src/components/editor/InputList.jsx
```diff
@@ -49,22 +49,30 @@ class
               .then(json => {
                 for(var i in json){
                   try{
-                    opts.push(json[i]["http://www.loc.gov/mads/rdf/v1#authoritativeLabel"][0]["@value"])
+                    // warns about security/detect-object-injection, but is safe because user cannot inject anything for `var i`
```
jermnelson: I was able to suppress these warnings using the following:
```              
                    const item = Object.getOwnPropertyDescription(json, i)
                    const uri = item["@id"]
                    const label = item["http://www.loc.gov/mads/rdf/v1#authoritativeLabel"][0]["@value"]```

jgreben: 😎 🆒 

Made this change....

PR #273, opened 2018-12-20
Generate RDF from ResourceTemplateForm

jermnelson:
Starts support for generating a simple RDF graph for components and props in a `ResourceTemplateForm`.  

- New reducer generates RDF based on the based on the Redux stores for `InputList`, `InputLiteral`, and `InputLookup`. 
- Adds a new key, **rtId** for storing the `ResourceTemplate` id in `InputList`, `InputLiteral`, and `InputLookup`, this is used to scope the generation of triples to a single ResourceTemplate
- Uses [N3](https://github.com/rdfjs/N3.js) JS module for RDF manipulation. 
- New Preview RDF button displays a JS alert with the serialized graph in Turtle format. 
- All subjects are blank nodes
### New Preview RDF Button
screen shot 2019-01-11 at 12 25 41 pm

### Alert with RDF Serialization in Turtle
screen shot 2019-01-11 at 9 57 47 am

## Know Issues and Future Work
- Only works for `ResourceTemplateForm`s that are **NOT** embedded in a modal; issue is ticketed in #281. 
- Once #281 is completed, will need to return the blank node or uri to the calling `ResourceTemplateForm` to complete the triple, ticketed in #282
- Should use the N3 Store instead of a N3 writer for adding triples to the graphs. Using `N3.Store` may make it easier to send RDF back to Trellis in the Sinopia Server. Ticket #280 
- Need to either mint a new URI from a Sinopia Server call or create a dummy URI for client-side RDF subjects for those `ResourceTemplateForm`s that need a URI for the subject. Ticket #283

### DIFF
--- a/src/reducers/rdf.js
+++ b/src/reducers/rdf.js
@@ -1,0 +1,48 @@
+const N3 = require('n3')
+
+const { DataFactory } = N3
+const { namedNode, literal, blankNode, quad } = DataFactory
+
+const DEFAULT_STATE = {
+  graph: null
+}
+
+const generateMyRDF = (state, action) => {
+  const writer = N3.Writer({
+    prefixes: { bf: 'http://id.loc.gov/ontologies/bibframe/'}
+  })
+  const subject = blankNode()
+  action.payload.literals.formData.forEach(field => {
+    field.items.forEach(item => {
+      if(item.hasOwnProperty('content')) {
+        writer.addQuad(
+          quad(subject, namedNode(field.id), literal(item.content))
+        )
+      }
+    })
+  })
+  action.payload.lookups.formData.forEach(field => {
+    field.items.forEach(item => {
+      if(item.hasOwnProperty('uri')) {
+        writer.addQuad(subject, namedNode(field.id), namedNode(item.uri))
+      }
+    })
+  })
+  // Temp solution displays RDF in Turtle, likely use a Modal Component
+  // to display graph in multiple formats.
+  writer.end((error, result) => {
+    alert(result)
+    return { ...state, graph: result }
+  })
+}
+
+const generateRDF = (state=DEFAULT_STATE, action) => {
+  switch(action.type) {
+    case 'GENERATE_RDF':
+      return generateMyRDF(state, action)
+    default:
+      return state
+  }
+}
+
+export default generateRDF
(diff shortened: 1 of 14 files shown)


jermnelson:
Thanks @SaravShah! I don't think the InputLookup and InputList needs the language selection because they actually are just URI references in the RDF to external resources. Those sources would have the actual literals for such things as labels that may have the language attribute set.

jgreben:
When I run this branch locally I get some console errors:

`checkPropTypes.js:19 Warning: Failed prop type: Invalid prop `literals` of type `object` supplied to `ResourceTemplateForm`, expected `array`.`

`checkPropTypes.js:19 Warning: Failed prop type: Invalid prop `lookups` of type `object` supplied to `ResourceTemplateForm`, expected `array`.`

`checkPropTypes.js:19 Warning: Failed prop type: The prop `rtType` is marked as required in `ResourceTemplateModal`, but its value is `undefined`.`

Also, the RDF for the LOC names lookup does not seem right:
screen shot 2019-01-10 at 9 48 54 am

ndushay:
actually, would be awesome to see images of button added and what clicking the button looks like.  I use "command - control - shift - f4" to take cropped screenshot and then I paste it into the description.

I found this practice super helpful when last I worked with folks on access team.

jermnelson:
Hi @jgreben, latest commit fixes prop types errors and adjusted `InputLookup` to use propertyURI instead of propertyLabel for the **id** which fixes  the RDF for the **LOC names lookup** above.

Good idea about the screenshots @ndushay, added screen shots to the PR description.

src/reducers/rdf.js
```diff
@@ -0,0 +1,72 @@
...
+
+const { DataFactory } = N3
+// TODO: Use N3.Store() for graph state
+const { namedNode, literal, blankNode, quad } = DataFactory
+
+const DEFAULT_STATE = {
+  graph: {}
+}
+
+const addTriples = (subject, formData, rtId, testProperty, graph) => {
```
jgreben: I wouldn't mind seeing if we can refactor this so that we are not always passing in all these arguments. Maybe separate functions to handle 'uri' and 'content'/other (e.g. addTriplesForUri())? I'd be happy to work on this as part of a follow-on PR. Non-blocker for this PR.

jermnelson: Yeah, not sure what the best option is here; originally I had this functionality within the `generateMyRDF` with two forEach loops.

PR #274, opened 2018-12-20
Set default values for InputList component if they exist

### LINKED ISSUE #265: Implement defaultLiteral for InputList
e.g.
```
"defaults": [
          {
            "defaultURI": "http://id.loc.gov/vocabulary/carriers/nc",
            "defaultLiteral": "volume"
          }
```

jgreben:
Fixes #265

### DIFF
--- a/src/components/editor/InputList.jsx
+++ b/src/components/editor/InputList.jsx
@@ -12,8 +12,30 @@ class
     super(props)
     this.state = {
       isLoading: false,
-      options: []
+      options: [],
+      defaults: []
     }
+    let defaultValue
+    try {
+      defaultValue = this.props.propertyTemplate.valueConstraint.defaults[0]
+      const defaults = [{
+        id: defaultValue.defaultURI,
+        uri: defaultValue.defaultURI,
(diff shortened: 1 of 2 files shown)


src/components/editor/InputList.jsx
```diff
@@ -12,10 +12,32 @@ class
...
+    }
+    let defaultValue
+    try {
+      defaultValue = this.props.propertyTemplate.valueConstraint.defaults[0]
+      const defaults = [{
+        id: defaultValue.defaultURI,
+        uri: defaultValue.defaultURI,
+        label: defaultValue.defaultLiteral
+      }]
+      this.state.defaults = defaults
```
SaravShah: im guessing we need to store defaults in the state due to the TypeAhead library right?

jgreben: Yeah, the typeahead component needs to have the defaultSelected attribute set with the values before it loads up, so the InputList's state seemed like a good place to keep that since it gets set in it's constructor prior to the typeahead component loading. I'm open to other suggestions if you think there are better ways to do it.

PR #286, opened 2019-01-09
Use a json config doc to match template uris to component and connection details

### LINKED ISSUE #284: Add additional support for QA lookup
a JSON config doc, to include:
 - all LOC vocab lists
 - all possible QA lookups
 - which component to use (to parse json)
 - URI that gets JSON
 - label to display to user

JSON config doc lives....where?

e.g. http://sinopia.ip/something/vocabLookup.json

BFF: Rename current components viz QA & LOC
PE: dropdown list to choose lookup/list options

jgreben:
Fixes #284
Select input components and pass connection details as props based on template match to config details; Renamed Input components that do lookups

### DIFF
--- a/__tests__/components/editor/InputLiteral.test.js
+++ b/__tests__/components/editor/InputLiteral.test.js
@@ -16,7 +16,7 @@ const plProps = {
 }
 
 describe('<InputLiteral />', () => {
-  const wrapper = shallow(<InputLiteral {...plProps} />)
+  const wrapper = shallow(<InputLiteral {...plProps} id={10}/>)
   
   it('contains a label with "Instance of"', () => {
     expect(wrapper.find('label').text()).toBe('Instance of')
@@ -41,7 +41,7 @@ describe('When the user enters input into field', ()=>{
   // our mock formData function to replace the one provided by mapDispatchToProps
   const mockFormDataFn = jest.fn()
   const removeMockDataFn = jest.fn()
-  mock_wrapper = shallow(<InputLiteral {...plProps} id={"11"}
+  mock_wrapper = shallow(<InputLiteral {...plProps} id={11}
                                        handleMyItemsChange={mockFormDataFn}
                                        handleRemoveItem={removeMockDataFn}/>)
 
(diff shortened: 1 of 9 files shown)


__tests__/components/editor/InputLiteral.test.js
```diff
@@ -16,7 +16,7 @@ const plProps = {
 }
 
 describe('<InputLiteral />', () => {
-  const wrapper = shallow(<InputLiteral {...plProps} />)
+  const wrapper = shallow(<InputLiteral {...plProps} id={10}/>)
```
SaravShah: nice! i made this change in my PR as well, so i'll rebase if your PR gets merged first

src/components/editor/ResourceTemplateForm.jsx
```diff
@@ -62,35 +63,57 @@ class ResourceTemplateForm extends Component {
...
+                    templateUri = pt.valueConstraint.useValuesFrom[0]
+                    for(var i in lookupConfig){
+                      lookupConfigItem = Object.getOwnPropertyDescriptor(lookupConfig, i)
+                      if(lookupConfigItem.value.uri === templateUri){
+                        listComponent = lookupConfigItem.value.component
+                        break
+                      }
+                    }
+                  } catch (error) {
+                    console.log(`Some properties were not defined in the json file: ${error}`)
```
ndushay: could/should this message:
1.  be console.error
2.  be clearer - which JSON file?
3.  does ${error} display real info or does it display "object"?   should it be `${util.inspect(error)}` ?

jgreben: ${error} displays real info, e.g. `TypeError: Cannot read property 'useValuesFrom' of undefined`

If I change it to error, we will get a lot of red lines in otherwise passing tests for templates that do not have `useValuesFrom` as well as in the console. The try/catch here is not totally intended to catch errors, but also to otherwise allow rendering where it would normally break because there was not a value in the template needed for setting the variables. I suppose I could check for the undefined variables instead of using try/catch here.

jgreben: Removed try/catch altogether.

PR #300, opened 2019-01-16
RDF Reducer: Refactor and use N3 Store

### LINKED ISSUE #280: Use N3.Store in RDF Reducer
The RDF reducer uses a N3.Writer to add triples  in order to display an alert with the serialized graph. The N3 Store has the graph in memory and will likely be easier to post to the backend. 

Will need to generate multiple serializations with a N3 writer based the graph(s) in the store.

jgreben:
Fixes #280 
- modify existing tests to handle necessary props
- Refactor rdf reducer with filter for resource template scope
- refactor rdf reducer tests
- create modal to display RDF filtered to RT scope, except for top-level (parent) RT.


src/reducers/rdf.js
```diff
@@ -1,63 +1,73 @@
...
-  addTriples(
-    subject,
-    action.payload.lookups.formData,
-    action.payload.rtId,
-    'uri',
-    writer
-  )
+
   // Temp solution displays RDF in Turtle, likely use a Modal Component
   // to display graph in multiple formats.
```
SaravShah: maybe we should create a ticket for this, just so it doesnt get overlooked?

Issue #301, opened 2019-01-18
BFF integration tests for JSON schema validation on loading resource template


ndushay:
currently, the StartingPoint code writes message to console log - indicating it passes schema validity or writing out the error messages.  There should be at least two tests:

1.  writes message to console when schema passes (and displays the correct template form)
2.  writes error messages to console (with a TODO of what should happen in browser)

Because of the interactions between StartingPoint and DropZone, testing for this must be done as an integration test.

While the code will currently support puppeteer syntax, chromium cannot get to an appropriate web server.  jest-dev-server  is in the mix, and webpack-dev-server .

Issue #295, opened 2019-01-11
JSON schema validation:  display errors to end user (not just console.log)


ndushay:
Obviously, it is not sufficient for json schema validation errors to only show in the browser console - the user must know the specifics of validation errors and there must be UX design when they can't proceed to creating metadata in BFF.

Naomi already starting discussing this with Astrid;  designs await discussion clarifying whether BFF or profile editor will be where the resourceTemplates are written to sinopia server.

Places in the code:

```bash
src/components/editor/StartingPoints.jsx
50:            // TODO: don't continue on with display of ResourceTemplate component (#295)
55:          // TODO: don't continue on with display of ResourceTemplate component (#295)
59:        // TODO: don't continue on with display of ResourceTemplate component (#295)
83:          // TODO:  indicate template validity to user (#295)
86:            // TODO:  indicate template errors to user (#295)

__tests__/integration/schemaValidation.test.js
40:      // TODO: beef this up after implementing display of validation errors to end user (#295)

.eslintrc.js
61:        // until we stop indicating schema validation with console logging (see #295)
```

Issue #294, opened 2019-01-11
JSON schema validation:  get schema from url (w version) within resourceTemplate


ndushay:
Fetching the schemas should be from the schema URL within the resourceTemplate as a property (URL will also contain a version), rather than hardcoded ones (which can only guess at the correct version of the schemas to use).

Of course, this depends on the resourceTemplates being updated to have said schema URL property ...

ndushay:
Also tweak the fixtures as needed.

Issue #292, opened 2019-01-11
cache resourceTemplate JSON schemas in local storage


ndushay:
rather than re-fetch JSON schemas in the BFF every time StartingPoints component mounts, it probably makes sense to cache the schemas in local storage.

PR #302, opened 2019-01-18
BFF does JSON schema validation of resource template on import

### LINKED ISSUE #258: schema validate resourceTemplate as part of import
JSON schemas currently live in sinopia_sample_profiles.  This will involve:

- [ ] a stable place to get JSON schemas for resource template (a web home?  they are currently in sinopia_sample_profiles repo but may need a web home) (see LD4P/sinopia#143)
- [ ] adding validator code to BFF app (this issue)
- [ ] running schema through validation as part of "import" action (this issue)

ndushay:
Uses ajv for schema validation of resource template on import.   Results of validation are written to console;  nothing else is changed.

- Tests are incomplete as integration tests are required, and it was non-trivial to untangle that (see #301)  --> hence coverage drop.

Note that we also still need to:
- get schema url/$id from template being validated (#294)
- indicate template validity to user and adjust UX flow (#295)
- cache the schemas in local storage rather than doing a get each time (#292)

Closes #258  (????)

### DIFF
--- a/__tests__/components/editor/StartingPoints.test.js
+++ b/__tests__/components/editor/StartingPoints.test.js
@@ -5,15 +5,23 @@ import { shallow, mount } from 'enzyme'
 import StartingPoints from '../../../src/components/editor/StartingPoints'
 import DropZone from '../../../src/components/editor/StartingPoints'
 
-const jsdom = require("jsdom");
+const jsdom = require("jsdom")
+require('isomorphic-fetch')
 
-setUpDomEnvironment();
+setUpDomEnvironment()
 
 describe('<StartingPoints />', () => {
   let wrapper = shallow(<StartingPoints />)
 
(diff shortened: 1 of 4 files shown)


jgreben:
@ndushay do you want to me to merge regardless of coverage drop? Do you want to decrease the threshold for the time being so we all don't all fail the checks after merge?

ndushay:
I would say it's @jermnelson's call if we want to decrease the threshold.  And I'm not sure if Jeremy wanted to review this himself, too.

jermnelson:
I think decreasing this threshold is fine to pass for now as @ndushay mentioned and ticketed in the description.

ndushay:
coverage threshold is 83%;  i suspect this is failing due to decrease, not from threshold.  I recommend decrease be ignored in this PR.

PR #319, opened 2019-02-01
Displaying generated json-ld and handle switching resource templates

jgreben:
Adds reducer+action, modal, and tests for resource template forms and modals and handles the reset of form state. Specifically:

- Create a root reducer with action to reset state of lookups, literals and generateLD graph
- Increase max warnings for eslint because of template validation console logging 
- Set the application state with the currently selected resource template data when Editor component is updated from StartingPoints menu or from an imported file via the DropZone component
- InputLiteral sends the currently generated blanknode to the literal redux state for proper generation of LD
- Pass rdfOuterSubject and rdfPredicate props through ModalToggle and ResourceTemplateModal in order to finally generate proper LD for a modal
- Reducer to generate LD based on redux state for literals and lookups


__tests__/components/editor/RDFModal.test.js
```diff
@@ -0,0 +1,50 @@
+// Copyright 2018 Stanford University see Apache2.txt for license
```
jermnelson: Since this file was created in 2019, maybe update the Copyright statement to 2019? Just in general, wonder if we should do a mass update on all of our code?

jgreben: ¯\_(ツ)_/¯ do you want me to just change this one for now or make a ticket to do them all?

PR #336, opened 2019-02-06
Adds the default literal value if it exists

### LINKED ISSUE #196: add a default value to form from propertyTemplate
- for default value given, want to show that for type: resource, target and look up
- the one instance that is type literal is actually a mistake.

Requires #265

jgreben:
Fixes #196 

- Gets default Literal value from property template and sets the "defaults" state with that data, plus any rdf-related properties that get passed in (`bnode` and `propPredicate`)
- Adds a test for this new feature
- Minor formatting of `InputListLOC`
- adds a default literal to the Barcode resource template so we can see that this is working.

### DIFF
--- a/src/components/editor/InputLiteral.jsx
+++ b/src/components/editor/InputLiteral.jsx
@@ -27,9 +27,8 @@ export class InputLiteral extends Component {
     }
     this.lastId = -1
 
-    let defaultValue
     try {
-      defaultValue = this.props.propertyTemplate.valueConstraint.defaults[0]
+     const defaultValue = this.props.propertyTemplate.valueConstraint.defaults[0]
       this.state.defaults = [{
         content: defaultValue.defaultLiteral,
         id: ++this.lastId,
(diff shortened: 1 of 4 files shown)


src/components/editor/InputLiteral.jsx
```diff
@@ -22,9 +22,33 @@ export class InputLiteral extends Component {
     this.notRepeatable = this.notRepeatable.bind(this)
     this.addUserInput = this.addUserInput.bind(this)
     this.state = {
-      content_add: ""
+      content_add: "",
+      defaults: []
     }
     this.lastId = -1
+
+    let defaultValue
```
mjgiarlo: Do we need the `let` statement here? It looks as though `defaultValue` is only used in the `try` block.

jgreben: You're right, I guess as a bit of defensive programming, I thought it would barf on setting that variable if there were no `valueConstraint.defaults[0]` but it appers that is not the case. I also did this same thing in the InputListLOC component, so I changed that as well. Also, since the variable is not mutating it can be a `const` I think.

jermnelson: 👍

Issue #347, opened 2019-02-08
Indicate when a field is non-repeatable


jgreben:
Currently when a field is non-repeatable, you cannot enter a second value, but there is no indication and it appears buggy.

jgreben:
Need to do this at the `type: literal` and `type:resource` levels

michelleif:
please see #204 also incl the comments there

michelleif:
still need to address the case where repeatable=false for InputLiteral *and* there's a default value. if the field is non-repeatable, and there's already a default value, the field should be greyed out / uneditable like we did in #350 .

jermnelson:
^^ Now fixed with [PR#35](https://github.com/LD4P/sinopia_editor/pull/354)

jgreben:
Closed by #350

PR #350, opened 2019-02-12
Support for repeatable=false for InputLiteral

jermnelson:
For the `InputLiteral` component, if the `PropertyTemplate`'s *repeatable* property is set to false, then the InputLiteral doesn't allow a second item to be added but the problem is there is no visual indication and users can still type in a value but nothing is saved. This PR address one of the two components mention in ticket [#347 Indicate when a field is non-repeatable](https://github.com/LD4P/sinopia_editor/issues/347). 

Starting with a propertyTemplate with repeatable=false:
screen shot 2019-02-12 at 9 38 37 am

I can enter text in the input field but hitting enter doesn't do anything:
screen shot 2019-02-12 at 9 39 39 am

This PR set's the input field to the Bootstrap-styles *disabled* attribute after one item is saved following BFE design:
screen shot 2019-02-12 at 9 42 49 am

This PR also fixes a bug where if the user clicks and removes the single item, the input field is no longer disabled and allows the user to add in a new value:
screen shot 2019-02-12 at 9 44 41 am

### DIFF
--- a/src/components/editor/InputLiteral.jsx
+++ b/src/components/editor/InputLiteral.jsx
@@ -32,7 +32,8 @@ export class InputLiteral extends Component {
     this.state = {
       show: false,
       content_add: "",
-      defaults: []
+      defaults: [],
+      disabled: false
     }
     this.lastId = -1
 
@@ -76,9 +77,10 @@ export class InputLiteral extends Component {
     this.setState({ content_add: usr_input })
   }
 
-  notRepeatable(userInputArray,currentcontent){
-    if (this.props.formData == undefined){
+  notRepeatable(userInputArray, currentcontent){
+    if (this.props.formData == undefined || this.props.formData.items < 1){
       this.addUserInput(userInputArray, currentcontent)
+      this.setState({ disabled: true})
     }
   }
 
@@ -125,6 +127,12 @@ export class InputLiteral extends Component {
     {
       id: idToRemove, label: labelToRemove
     })
+    this.props.formData.items.forEach(item => {
+      if(item.id === idToRemove) {
+        this.props.formData.items.pop(item)
+      }
+    })
+    this.setState({disabled: false})
   }
 
   checkMandatoryRepeatable() {
@@ -232,6 +240,7 @@ export class InputLiteral extends Component {
             onChange={this.handleChange}
             onKeyPress={this.handleKeypress}
             value={this.state.content_add}
+            disabled={this.state.disabled}
             id={"typeLiteral" + this.props.id}
             onClick={this.handleFocus}
           />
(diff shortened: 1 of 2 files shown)


src/components/editor/InputLiteral.jsx
```diff
@@ -76,9 +77,10 @@ export class InputLiteral extends Component {
     this.setState({ content_add: usr_input })
   }
 
-  notRepeatable(userInputArray,currentcontent){
-    if (this.props.formData == undefined){
+  notRepeatable(userInputArray, currentcontent){
+    if (this.props.formData == undefined || this.props.formData.items < 1){
       this.addUserInput(userInputArray, currentcontent)
+      this.setState({ disabled: true})
```
mjgiarlo: Is the whitespace between the opening brace and `disabled:` intentional or stray?

src/components/editor/InputLiteral.jsx
```diff
@@ -125,6 +127,12 @@ export class InputLiteral extends Component {
     {
       id: idToRemove, label: labelToRemove
     })
+    this.props.formData.items.forEach(item => {
+      if(item.id === idToRemove) {
```
mjgiarlo: Obviously not a big deal, but I notice this `if` statement has no space between the keyword and the opening paren, but the new `if` added on line 81 does have a space. Are we using a common linter for this codebase? Should we? Or are folks OK with variability in code style within the codebase?

As someone who's new to React (and not a Javascript expert generally), I'll largely learn by what I see, and the inconsistency can throw me off. That said, I can get over it if this isn't an area the team wants to care about at the moment. :smiley:

src/components/editor/InputLiteral.jsx
ndushay: we're not terribly consistent on that particular whitespace, and we've been ignoring the inconsistency.

src/components/editor/InputLiteral.jsx
ndushay: Perhaps we should discuss at next retro?  We've been living with some whitespace inconsistency (but not indentation) in the js apps.   We were using at least 3 diff editors.

mjgiarlo:
@ndushay I am :100: OK with this PR being merged. I think the questions I raised are more about process (appropriate for retro or a dev-focused sidebar, like you suggested) than about this particular PR. Feel free to merge, folks.

src/components/editor/InputLiteral.jsx
jermnelson: Definitely a stray

michelleif:
looks good from user side, tested with a profile I made with a non-repeatable literal field

PR #363, opened 2019-02-13
When setting default literal values, make sure to increment last id

### LINKED ISSUE #361: Fix Warning: Encountered two children with the same key, `0`.
Adding another literal value for aliteral on the main RT form gives an error:

Warning: Encountered two children with the same key, `0`

screen shot 2019-02-13 at 3 02 44 pm

jgreben:
Fixes #361 

Also, making the proptype validation for `formData.uri` not required, because we do not always have this value, for example, in the tests and in the removeAllItems function when we empty out the formData state completely.

### DIFF
--- a/__tests__/components/editor/InputLiteral.test.js
+++ b/__tests__/components/editor/InputLiteral.test.js
@@ -174,15 +174,25 @@ describe('When the user enters input into field', ()=>{
 describe('when there is a default literal value in the property template', () => {
   it('sets the default values according to the property template if they exist', () => {
     plProps.propertyTemplate['valueConstraint'] = valConstraintProps
+
     const setDefaultsForLiteralWithPayLoad = jest.fn()
     const defaultsForLiteral = jest.fn()
-    shallow(<InputLiteral {...plProps} id={12}
+    defaultsForLiteral.mockReturnValue({
+      content: "content",
+      id: 0,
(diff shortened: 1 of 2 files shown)


__tests__/components/editor/InputLiteral.test.js
```diff
@@ -174,15 +174,25 @@ describe('When the user enters input into field', ()=>{
...
+    const wrapper = shallow(<InputLiteral {...plProps} id={12}
                           blankNodeForLiteral={{ termType: 'BlankNode', value: 'n3-0'}}
                           rtId={'resourceTemplate:bf2:Monograph:Instance'}
                           setDefaultsForLiteralWithPayLoad={setDefaultsForLiteralWithPayLoad}
-                          defaultsForLiteral={defaultsForLiteral} />)
+                          defaultsForLiteral={defaultsForLiteral}
+    />)
     expect(setDefaultsForLiteralWithPayLoad).toHaveBeenCalledTimes(1)
     expect(defaultsForLiteral).toHaveBeenCalledTimes(1)
+    expect(wrapper.instance().lastId).toEqual(0)
```
ndushay: I'm glad this is being tested, but I barely understand it now, let alone a month from now.  Maybe add a comment as to why it needs to be zero, or make a focused separate test with a better description about it?

PR #395, opened 2019-03-06
Adds a message if the user is not logged in andvisits the editor path

### LINKED ISSUE #349: JWT Authentication for Editor route
If a user is missing a JWT or does not have a valid JWT, display an error message when trying to access the `/editor` route.

jgreben:
Fixes #349 

- From the Editor page we don't need to force the user to authenticate, just warn them and provide a login link
- Use the same alert message style when redirecting to Login from Import, and increase the timeout so the user can see the message before being redirected to Cognito.

### DIFF
--- a/__tests__/components/Login.test.js
+++ b/__tests__/components/Login.test.js
@@ -5,10 +5,23 @@ import { mount } from "enzyme"
 
 describe('Login', () => {
   it('should call the cognitoLogin function', () => {
-    const context = { state: 'fake' };
-    const wrapper = mount(<Login location={context}/>)
-    const spy = jest.spyOn(wrapper.instance(), 'cognitoLogin');
-    spy.mockImplementation(() => "mock");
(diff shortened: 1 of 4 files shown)


jermnelson:
LGTM; merging.

__tests__/components/Login.test.js
```diff
@@ -5,10 +5,23 @@ import { mount } from "enzyme"
...
     expect(wrapper.instance().cognitoLogin('url')).toEqual('mock')
   })
+
+  it('displays a message for the import path', () => {
+    const fromImport = {state: {from: { pathname: '/import'}}}
+    const wrapper = mount(<Login location={fromImport} test={true}/>)
+    expect(wrapper.find('div.alert-warning').text()).toMatch('You must be logged in to access the "import" path')
+  })
+
+  it('displays', () => {
```
ndushay: this seems like a bad description ...

PR #399, opened 2019-03-07
Add docker-compose configuration to make it easier to develop/test editor integration with back-end components

mjgiarlo:


Dockerfile
```diff
@@ -1,4 +1,12 @@
-FROM node:10.11
+FROM node:10-alpine
```
mjgiarlo: Do this to make sure the `apk add` command on line 4 works. (Needed for `dockerize`.)

Dockerfile
```diff
@@ -1,4 +1,12 @@
...
+
+# Use dockerize to force indexer to wait for the broker to be ready
```
mjgiarlo: Add `dockerize` to make docker-compose services wait on each other as appropriate.

Dockerfile
```diff
@@ -7,4 +15,6 @@ COPY . .
 # docker daemon maps app's port
 EXPOSE 8000
 
+RUN npm install
```
mjgiarlo: Have the container install its dependencies instead of including `node_modules` in the container. (Not sure if this is a good pattern or an anti-pattern, tbh. Would like feedback from folks with more Docker experience.)

docker-compose.yml
```diff
@@ -0,0 +1,71 @@
+version: "3"
+services:
+  editor:
+    build:
+      context: .
+    ports:
+      - 8000:8000
+    depends_on:
+      - pipeline
+  pipeline:
```
mjgiarlo: There's a crap-ton of copy-pasta here that's from the pipeline's docker-compose config. I don't know if there's a better way to declare dependencies so that we don't need all this identical config in both repos. Advice requested.

Dockerfile
ndushay: I think we were caching node_modules to speed up our builds and to aid in deployment?  Maybe?  @jermnelson can confirm.

Dockerfile
ndushay: I'm muzzy on all this, but check out https://hub.docker.com/r/circleci/node  -- it has dockerize, and will work on circleci.   Maybe that's a worse solution than what you picked.  By my research here:  https://hub.docker.com/_/node/  we are already out of date on our node release.  :-P

docker-compose.yml
jermnelson: I'm not sure either or what the best practices are in this case. In the past, I've tried to share base_container between repos but some of the container configuration info the docker-compose.yml is going to be duplicative.

PR #405, opened 2019-03-11
Use localStorage to manage the authentication status if the user 

### LINKED ISSUE #316: How to set Cognito session from browser's localStorage

jgreben:
Fixes #316 

On the `HomePage` it checks whether the user is authenticated in redux; if not, it will look for the JWT query param and authenticate them. When the component updates it will then save the authentication state to localStorage via the `saveState()` call.

The `LoginPanel`, `Editor`, and other child components then look for the authentication state via the `loadState()` call and set their local state accordingly. 

The HomePage `logOut()` function gets passed to the `LoginPanel` child component (other child components will need it later) and when called it clears the localStorage state and then redirects to cognito for the destruction of the JWT.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -47,7 +47,14 @@ module.exports = {
   },
   overrides: [
     {
-      "files": ["**/*.jsx", "src/index.js", "src/*/*.js", "src/store.js", "src/Config.js", "src/LoginData.js", "src/Logout.js"],
+      "files": ["**/*.jsx",
+                "src/index.js",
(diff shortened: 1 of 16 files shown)


jermnelson:
LGTM; one question, is there a way to set a timeout for the `jwtAuth` stored in LocalStorage?

jgreben:
@jermnelson I added a timeout check to `localStorage.js`. When the loadState() function is checked, it will return `undefined` if the timeout set in the jwtAuth hash has passed .

PR #377, opened 2019-02-28
UI Implementation and Redux Store Changes 

jermnelson:
In the [February 19th Code4Lib Pre-conference](https://ld4p.github.io/code4lib_2019/) session on user interfaces designs  available on the current master branch at https://development.sinopia.io:

screen shot 2019-02-28 at 12 08 55 pm

The Sinopia Linked Data Editor follows the Library of Congress's BFE UI design of externally referenced Resource Templates with a Resource Template are rendered as pop-up modals:

screen shot 2019-02-28 at 12 16 06 pm

Many of the participants  [mentioned](https://github.com/LD4P/sinopia/wiki/Code4Lib-2019-Pre-conference) that after a couple of modals it was very difficult to keep the context what linked-data they were working on. This feedback wasn't new to the team and @astridu had developed a UI that created Property Panels with the externally Referenced Templates being rendered in an outline form:

screen shot 2019-02-28 at 12 18 30 pm

This new design was much preferred by the participants and this PR is the first steps to reaching this UI that not only is easier to keep the context, it turns out that this UI enabls a simpler method for maintaining the Editor's Redux state using [reselect](https://github.com/reduxjs/reselect) selectors. Here is the new `PropertyPanel` components that replace the previous `ResourceTemplateModal` with all of the items collapsed:

screen shot 2019-02-28 at 12 55 40 pm

Here is the UI with the `PropertyResourceTemplate` expanded:
screen shot 2019-02-28 at 12 54 12 pm

- [x] Get all tests to pass and ~~add tests for all of the new components~~ (see [fd4ea90](https://github.com/LD4P/sinopia_editor/pull/377/commits/fd4ea907f39a34efcea0816bcb2950e6fdce03f1).)
- [x] Support expanding/contracting outline in the `OutlineHeader`, see [310443c](https://github.com/LD4P/sinopia_editor/pull/377/commits/310443cb1943fe081377271cab232f9b2e1a9e8d)

## Remaining TODOs in follow-up tickets and PRs
- [x] ~~Click the **Add** button in the `PropertyPanel` adds a new property or is disabled when **repeatable** is false and a value exists..~~ (moved to [PR# 387 PropertyPanel UI -- Milestone 2 requirements](https://github.com/LD4P/sinopia_editor/issues/387))
- [x] ~~Defaults are initially saved the same shape as expected in the `formData` hash in the
[...]

PR #407, opened 2019-03-12
schemaValidation.test: add cognito JWT token passed in

ndushay:
this is necessary for future work on schemaValidation tests (e.g. after import tab redirects to editor tab with data loaded properly - WIP by @jgreben)

@jermnelson this will help with #377 

Note that when `sinopia-devs` email list is approved, we will want to change the token used here (see LD4P/sinopia#190)


jgreben:
LGTM!

Issue #309, opened 2019-01-25
retrieve resourceTemplates from trellis server


ndushay:
depends on templates already being in server:  LD4P/sinopia_server/issues/38

At first, ok if proof of concept shows that it was retrieved somehow (in console?  as initial template???)

jermnelson:
I think if we can pull the currently spoofed `resourceTemplate:bf2:Monograph:Instance` as the initial template in a that would be a great proof-of-concept.

ndushay:
great idea, Jeremy!  i'll make it a priority.

jgreben:
In the editor when you drop a resource template or profile file to be loaded, it will:
- parse the file for contained resource templates
- check if the rtId already exists
- POST the RT data to Trellis
- re-display the table of resource templates including the newly loaded resource templates

jgreben:
Fixed by #453

mjgiarlo:
Extracted #459 from this issue.

PR #417, opened 2019-03-14
Prepare to retrieve resource templates from Sinopia platform

mjgiarlo:
Includes:

* Rename `sinopiaServerSpoof` to `sinopiaServer` in preparation for work in earnest on #309.
* Prefer `import ... from` load syntax to `const ... require()`.
* Use Array functions `map()` and `find()` in `sinopiaServer.js` instead of `forEach()`.
* Add docker-compose configuration so the editor can integrate with the Sinopia platform in dev/test.
* Use CircleCI's node image so we get `dockerize` for free (HT: @ndushay!).
* Use port 8888 for `webpack-dev-server` to prevent conflict with Trellis on port 8080, and reflect this change in integration tests.
* Bump Jest timeout length on integration tests (they fail otherwise on my laptop, and I want :green_heart: before I touch anything more).
* Extend eslint `node/no-unsupported-features/es-syntax` override to all JS files under `src/`, `__tests__`, and `__mocks__`, instead of onesy-twosying it, which should mean we don't have to touch this block again when adding new files.

connects to #309


jgreben:
LGTM!

Issue #388, opened 2019-03-06
PropertyOutline Rendering of Resource Template


jermnelson:
For internally rendering of Resource Templates using the `PropertyResourceTemplate` component, need a way to handling of two resource templates that refer to each other. Getting caught into infinite loops, one that just reaches a recursion limit and hangs the browser tab or worse, when using React component composition brings down the entire web browser, replicated in both Firefox and Chrome.

Will look at moving the render until the user clicks to expand that will make the call out to the server for the resource template.

Blocks #387
Would close the following open tickets:
* #152, #351, #352, #355

PR #419, opened 2019-03-15
Working PropertyTemplateOutline using embedded Resource Template in PropertyPanel Component

jermnelson:
As documented in issue #388, in the current master branch, for embedded resource templates in a `PropertyPanel` is not rendering the Resource Template, like **Has Bibframe Instance** below:
Screen Shot 2019-03-15 at 10 16 00 AM

This PR adds that rendering support:
Screen Shot 2019-03-15 at 3 43 03 PM

Remaining TODOs:
- [x] Add support for `InputLookupQA`
- [x] Deduplicate `this.state.output`
- [x] CSS and/or HTML changes to reduce left margin.
- [x] Unit tests for this expansion

### DIFF
--- a/__tests__/components/editor/PropertyTemplateOutline.test.js
+++ b/__tests__/components/editor/PropertyTemplateOutline.test.js
@@ -2,6 +2,7 @@
 
 import React from 'react'
 import { shallow } from 'enzyme'
+import InputLiteral from '../../../src/components/editor/InputLiteral'
 import PropertyTemplateOutline from '../../../src/components/editor/PropertyResourceTemplate'
 import PropertyTypeRow from '../../../src/components/editor/PropertyTypeRow'
 
@@ -9,7 +10,12 @@ describe('<PropertyTemplateOutline />', () => {
   let propertyRtProps = {
     resourceTemplate: {
       resourceLabel: "Test Schema Thing Template",
-      propertyTemplates: []
+      propertyTemplates: [
+        {
+          propertyLabel: "Test Schema name as a literal",
+          propertyURI: "http://schema.org/name"
+        }
+      ]
     }
   }
   const wrapper = shallow(<PropertyTemplateOutline {...propertyRtProps} />)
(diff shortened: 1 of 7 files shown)


__tests__/components/editor/PropertyTemplateOutline.test.js
```diff
@@ -21,4 +29,54 @@ describe('<PropertyTemplateOutline />', () => {
...
+            ],
+            "useValuesFrom": [],
+            "valueDataType": {},
+            "defaults": []
+          }
+        }
+      ]
+    }
+  }
+  const wrapper = mount(<PropertyTemplateOutline {...propertyRtProps}
```
jgreben: Is it possible to write some tests for the other propertyTemplate types (using different `propertyRtProps`) in order to get better test coverage on this component? Otherwise future PRs will need to address this code coverage (or we will have to keep decreasing the threshold). Coverage is down to 76% on this PR and we currently have 75% as the bottom threshold.

jgreben:
LGTM! I'd just like to see a bit higher coverage if possible on the one test mentioned 👆if possible.

__tests__/components/editor/PropertyTemplateOutline.test.js
jermnelson: Thanks @jgreben! I'll add some tests to increase coverage.

jermnelson: Hi @jgreben, the latest commit puts code coverage back over 77%

src/components/editor/PropertyActionButtons.jsx
```diff
@@ -0,0 +1,25 @@
...
+    </div>)
+  }
+}
+
+PropertyButtonActions.propTypes = {
+  handleAddClick: PropTypes.func,
+  handleMintUri: PropTypes.func
+}
+
+export default PropertyButtonActions;
```
ndushay: does it matter that this is a file called PropertyActionButtons, but it is declared an exporting PropertyButtonActions?    If it matters, I like PropertyActionButtons better ...

ndushay: and if this *is* a mistake, how is it the tests actually passed?????

jermnelson: Good catch! So, it was a mistake and just pushed up a commit that make. Speculating (but haven't confirmed) is that I think when we imported the ProperyActionButtons component in the code and tests, the name of the module was correct and the part of syntax `import ProperyActionButtons` creates a named variable in the calling code.

PR #425, opened 2019-03-19
Adds Stanford, Cornell, and UCSD ShareVDE authority sources

jermnelson:
Adds QA authority lookup sources for Stanford, Cornell, and UCSD Works and Instances. **NOTE** currently QA is returning an URI instead of a text literal in the label field. This will be fixed soon.


ndushay:
Do we really want to keep adding stuff to the spoof?  When will we be close enough to real to stop fussing with the spoof?

jgreben:
@ndushay when https://github.com/LD4P/qa_server/issues/51 gets resolved...

Issue #412, opened 2019-03-13
InputLiteral, InputListLOC, and InputLookupQA use common reducers


jermnelson:
`InputLiteral`, `InputListLOC`, and `InputLookupQA` components to all use the same Redux reducer with selectors. Accomplishing this item should address these bugs:
  Should close 
  - [#390 when a property template with NO default value follows another property template that DOES have a default value, the second property template incorrectly displays a default value](https://github.com/LD4P/sinopia_editor/issues/390) 
  - [#391 entered value is mistakenly duplicated across all literal fields in Profile](https://github.com/LD4P/sinopia_editor/issues/391)

Blocks #411 
Blocks #414

jgreben:
Blocks #414

Issue #421, opened 2019-03-18
Editor components Integration test


jermnelson:
Given a specific Resource Template, insure the editor generates the expected RDF from the redux store.

- [x] An authenticated user starts from the home page and clicks on the Linked Data Editor tab
- [ ] User loads a Resource Template into the Editor Tab; Resource Template should be moderately
   complex and include:
   - [x] `literal`, `lookup`, and `resource` propertyTemplates along with defaults.
   - [x] At least one `InputLiteral` with non-roman script
   - [ ] A least one `InputLiteral` with diacritic in latin script
    - [ ] A least one `InputLiteral`  transliterated title with ligatures

ndushay:
@jermnelson how does this relate to #423?   Is this a more contained set of tests checking more narrowly focused things?   B/C I'm not sure the tests being written are checking the RDF produced ...  (or is the title of this ticket misleading)

jermnelson:
Hi @ndushay, this list of items comes directly from [UAT ticket](https://github.com/LD4P/sinopia/issues/167) for Milestone 4 and the title was adopted from that description. I think #423 should be a separate integration test that does check the RDF output and I just updated #423 to reflect that.

ndushay:
Thanks @jermnelson .  I also just slacked with Aaron and he knows he's not yet checking the produced RDF, so he won't check the boxes here until he is.  :-)

aaron-collier:
Yes, my assumption here is that the output needs to be checked in order for these to be "complete"... if that is an incorrect assumption, I think the title should be updated.

jermnelson:
👍

justinlittman:
Current test coverage is adequate.

PR #426, opened 2019-03-20
Working selector and defaults for InputLiteral using common reducers

### LINKED ISSUE #390: when a property template with NO default value follows another property template that DOES have a default value, the second property template incorrectly displays a default value
Attached profile has one resource template with two property templates, both type literal. the first has a default value (1976), the second does not. When I load the profile into the Linked Data Editor, both "fields" show the default value 1976. Only the first field should show this.

JSON profile attached and screenshot of unexpected results when loading profile in Linked Data Editor

### LINKED ISSUE #391: entered value is mistakenly duplicated across all literal fields in Profile
I have a profile with one resource template and 4 literal fields. I open in Linked Data Editor, enter a value in one field, hit enter, and the value I entered now shows up beneath all 4 fields. (No default values are specified in this profile). Profile JSON and video attached.

### LINKED ISSUE #428: Input Literal Uses selectorReducer
Blocks #412

jermnelson:
On the way to finishing ticket #412, this PR uses Redux to set defaults based on Resource Template ID and URI of property using more generic reducers and a single `getProperty` selector to map the redux state to the InputLiteral's props. 

Fixes #390
Fixes #391
Fixes #428

### DIFF
--- a/__tests__/components/editor/InputLiteral.test.js
+++ b/__tests__/components/editor/InputLiteral.test.js
@@ -1,5 +1,5 @@
 // Copyright 2018, 2019 Stanford University see Apache2.txt for license
-
+import 'jsdom-global/register'
 import React from 'react'
 import { shallow } from 'enzyme'
 import { InputLiteral } from '../../../src/components/editor/InputLiteral'
@@ -176,16 +176,33 @@ describe('when there is a default literal value in the property template', () =>
   const mockRemoveItem = jest.fn()
 
   it('sets the default values according to the property template if they exist', () => {
-    plProps.propertyTemplate['valueConstraint'] = valConstraintProps
-
-    const wrapper = shallow(<InputLiteral {...plProps} id={12}
+    const plProps =  {
+      "propertyTemplate":
+        {
+          "propertyLabel": "Instance of",
+          "propertyURI": "http://id.loc.gov/ontologies/bibframe/instanceOf",
+          "type": "literal",
+          "mandatory": "",
+          "repeatable": "",
+          "valueConstraint": valConstraintProps
+        }
+    }
+    const wrapper = shallow(<InputLiteral propertyTemplate={plProps.propertyTemplate} id={12}
                           blankNodeForLiteral={{ termType: 'BlankNode', value: 'n3-0'}}
                           handleMyItemsChange={mockMyItemsChange}
-                          rtId={'resourceTemplate:bf2:Monograph:Instance'}
-    />)
-    wrapper.setProps({ formData: { items: valConstraintProps.defaults } })
+                          rtId={'resourceTemplate:bf2:Monograph:Instance'} />)
+    // Mocking a call to the Redux store
+    const items = [{
+      "uri": "http://id.loc.gov/vocabulary/organizations/dlc",
+      "content": "DLC"
+    }]
+    wrapper.setProps({
+      formData: {
+        items: items
+      }
+    })
     wrapper.instance().forceUpdate()
-    expect(wrapper.find('#userInput')).toBeTruthy()
+    expect(wrapper.find('#userInput').text()).toMatch(items[0].content)
   })
 
   describe('when repeatable="false"', () => {
--- a/__tests__/components/editor/InputLiteral.test.js
+++ b/__tests__/components/editor/InputLiteral.test.js
@@ -172,27 +172,20 @@ describe('When the user enters input into field', ()=>{
 })
 
 describe('when there is a default literal value in the property template', () => {
+  const mockMyItemsChange = jest.fn()
+  const mockRemoveItem = jest.fn()
+
   it('sets the default values according to the property template if they exist', () => {
     plProps.propertyTemplate['valueConstraint'] = valConstraintProps
 
-    const setDefaultsForLiteralWithPayLoad = jest.fn()
-    const defaultsForLiteral = jest.fn()
-    defaultsForLiteral.mockReturnValue({
-      content: "content",
-      id: 0,
-      bnode: { termType: 'BlankNode', value: 'n3-0'},
-      propPredicate: "predicate"
-    })
-
     const wrapper = shallow(<InputLiteral {...plProps} id={12}
                           blankNodeForLiteral={{ termType: 'BlankNode', value: 'n3-0'}}
+                          handleMyItemsChange={mockMyItemsChange}
                           rtId={'resourceTemplate:bf2:Monograph:Instance'}
-                          setDefaultsForLiteralWithPayLoad={setDefaultsForLiteralWithPayLoad}
-                          defaultsForLiteral={defaultsForLiteral}
     />)
-    expect(setDefaultsForLiteralWithPayLoad).toHaveBeenCalledTimes(1)
-    expect(defaultsForLiteral).toHaveBeenCalledTimes(1)
-    expect(wrapper.instance().lastId).toEqual(0)
+    wrapper.setProps({ formData: { items: valConstraintProps.defaults } })
+    wrapper.instance().forceUpdate()
+    expect(wrapper.find('#userInput')).toBeTruthy()
   })
 
   describe('when repeatable="false"', () => {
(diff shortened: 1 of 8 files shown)


__tests__/components/editor/InputLiteral.test.js
```diff
@@ -172,27 +172,20 @@ describe('When the user enters input into field', ()=>{
...
                           rtId={'resourceTemplate:bf2:Monograph:Instance'}
-                          setDefaultsForLiteralWithPayLoad={setDefaultsForLiteralWithPayLoad}
-                          defaultsForLiteral={defaultsForLiteral}
     />)
-    expect(setDefaultsForLiteralWithPayLoad).toHaveBeenCalledTimes(1)
-    expect(defaultsForLiteral).toHaveBeenCalledTimes(1)
-    expect(wrapper.instance().lastId).toEqual(0)
+    wrapper.setProps({ formData: { items: valConstraintProps.defaults } })
+    wrapper.instance().forceUpdate()
+    expect(wrapper.find('#userInput')).toBeTruthy()
```
ndushay: is this adequately testing what it is supposed to?  It doesn't seem to be checking for default values at all ...

__tests__/reducers/literal.test.js
```diff
@@ -1,88 +1,136 @@
...
+        uri:'http://schema.org/name',
+        items: [ { id: 0, content: 'Run the tests'} ]
+      }
+    })
+    expect(literalSetItems).toEqual({
+      "resourceTemplate:Monograph:Instance": {
+        'http://schema.org/name': {
+          items: [{ id: 0, content: 'Run the tests'}]
+        }
+      }
```
ndushay: I'm a bit muzzy on the diff between the old and the new, but the OLD way showed items as an empty array in all cases, whereas this new one shows items populated ...  in both cases, I'm not clear on what SET_ITEMS is supposed to be doing.  "handle SET_ITEMS" is pretty vague as a description.

I note the new test code certainly seems no worse than the old, in that regard ...

__tests__/reducers/literal.test.js
```diff
@@ -1,88 +1,136 @@
...
-      ]}, {
-        type: 'REMOVE_ITEM',
-        payload: {id: 0, label: "Statement"}
-      })
-    ).toEqual({
-      "formData": [
-        {"id": 1, "uri": "Test", "items": [{content: "test content", id: 0}]},
-        {"id": 2, "uri": "Statement", "items": []}
-      ]
+    expect(removeMyItem({
```
ndushay: this seems like a test showing that removing an item that doesn't match one we have already results in a no-op.  perhaps it warrants its own test description?

__tests__/reducers/literal.test.js
```diff
@@ -1,88 +1,136 @@
...
-        {content: "more content", id: 1}
-        ]}]}, {
-        type: 'REMOVE_ITEM',
-        payload: {id: 0, label: "Test"}
-      })
-    ).toEqual({
-      "formData": [{
-        id: 1, uri: "Test", "items": [{content: "more content", id: 1}]
-      }]
+    expect(removeMyItem({
```
ndushay: this seems to show that removing the item in the payload, when it matches one we have, results in that item being removed.  Perhaps the description for this ONE expects could be better?  "REMOVE_ITEM removes the item when it is found" ?

__tests__/reducers/literal.test.js
```diff
@@ -1,88 +1,136 @@
...
+      'http://schema.org/name': { items: [] }
+    }}, {
+      type: 'SET_ITEMS',
+      payload: {
+        rtId: 'resourceTemplate:Monograph:Instance',
+        uri:'http://schema.org/name',
+        items: [ { id: 0, content: 'Run the tests'} ]
+      }
+    })
+    expect(literalSetItems).toEqual({
```
ndushay: this expect seems to be about "replaces empty items with the payload items"

__tests__/reducers/literal.test.js
```diff
@@ -1,88 +1,136 @@
...
+      "resourceTemplate:Monograph:Instance": {
+        'http://schema.org/name': {
+          items: [{ id: 0, content: 'Run the tests'}]
+        }
+      }
     })
 
     expect(
```
ndushay: this expect seems to warrant its own description -- perhaps "SET_ITEMS appends new items to the existing array of items" 

which makes me wonder ... what happens if you pass in one that is already there?

src/components/App.jsx
```diff
@@ -45,6 +45,8 @@ class App extends Component{
 
     const defaultRtId = 'resourceTemplate:bf2:Monograph:Instance'
 
+    // const defaultRtId = 'resourceTemplate:bf2:Identifiers:Barcode'
+
```
ndushay: time to remove this line?

src/components/editor/InputLiteral.jsx
```diff
@@ -169,7 +146,9 @@ export class InputLiteral extends Component {
 
   makeAddedList = () => {
     let formInfo = this.props.formData
-    if (formInfo == undefined) return
+    // console.log(`makeAddedList ${this.props.propertyTemplate.propertyLabel}`)
+    // console.warn(formInfo)
```
ndushay: do we want to keep these console.log statements?

src/components/editor/InputLiteral.jsx
```diff
@@ -169,7 +146,9 @@ export class InputLiteral extends Component {
...
+    if (formInfo == undefined || formInfo.items == undefined) return
```
ndushay: this looks funny to me ... would we want the "return" on the next line, along with indenting the following lines appropriately?   Not a biggie, just asking.

__tests__/reducers/literal.test.js
jermnelson: The description should now make it clear that the test is about changing the Redux state based on the SET_ITEMS action being passed to the reducer (this should be the most common use case for the reducer).

__tests__/reducers/literal.test.js
jermnelson: Storing the InputLiterals state in the Redux is now changed to support selector pattern; changed the test description from the vague to hopefully a little more specific on what calling **SET_ITEMS** does to the Redux state for a specific InputLiteral.

__tests__/reducers/literal.test.js
jermnelson: Changed description to better reflect the change in state.

__tests__/reducers/literal.test.js
jermnelson: Changed description per comment.

src/components/editor/InputLiteral.jsx
jermnelson: This just follows the syntax for a `if` statement one-liner; changed to separate line to improve readability.

src/components/editor/InputLiteral.jsx
jermnelson: Removed in latest commit.

src/reducers/index.js
```diff
@@ -1,48 +1,60 @@
...
+    items = []
+  }
+  return items
+}
 
 export const getProperty = createSelector(
-  [resourceTemplateSelector, propertySelector],
-  (resourceTemplate, propertyURI)  => {
-    return resourceTemplate[propertyURI]
+  [ resourceTemplateSelector ],
```
jgreben: I assume you will be discussing exactly how this works next week, so after some smoke testing I will be able to merge this.

ndushay:
thanks for changing those test descriptions - I much prefer the new ones;  hope it wasn't too time consuming.  My question about that one test still stands -- *is* it testing what it says it's testing?

I'll leave it up to osh to merge;  I'm not going to the mat over my remaining question.

__tests__/components/editor/InputLiteral.test.js
jermnelson: Changed tests so we're actually testing the value set earlier; got sidetracked in spoofing strategies with Redux store on this issue, for now the .setProps is noted as a mock attempt to trigger the right action. From my research, [#421 RDF integration tests](https://github.com/LD4P/sinopia_editor/issues/421) should be used to actually test setting defaults in the `IndexLiteral` components based on state of the Redux store.

src/reducers/index.js
jermnelson: Cool! Yes, this will be a core part of how this works next week. Thanks for creating tickets [429]( https://github.com/LD4P/sinopia_editor/issues/429) and [430](https://github.com/LD4P/sinopia_editor/issues/430) for follow-up.

PR #433, opened 2019-03-21
List of resource templates retrieved from sinopia_server

### LINKED ISSUE #418: Display a List of Available Resource Templates
In the Import Resource Template tab, display a list of resource templates from Sinopia Server.

jgreben:
Fixes #418 
Make calls to sinopia_server using the swagger generated API code. If the calls come back with an error message, display on the page, otherwise display the returned data in a bootstrap table:

**No connection to sinopia_server, or no data retrieved from calls:**

Screen Shot 2019-03-21 at 4 34 58 PM

**sinopia_server is connected and returns the expected data from `repository/ld4p`:**

Screen Shot 2019-03-28 at 3 29 02 PM

### DIFF
--- a/__tests__/components/editor/SinopiaResourceTemplates.test.js
+++ b/__tests__/components/editor/SinopiaResourceTemplates.test.js
@@ -21,35 +21,43 @@ describe('<SinopiaResourceTemplates />', () => {
     expect(wrapper.find(BootstrapTable).dive().length).toEqual(1)
   })
 
-  describe('gettiong data from the sinopia_server', () => {
+  describe('getting data from the sinopia_server', () => {
 
-    it('sets the state with group data (as Array) from sinopia_server', async() => {
+    const mockResponse = (status, statusText, response) => {
+      return new Response(response, {
+        status: status,
+        statusText: statusText,
+        headers: {
+          'Content-type': 'application/json'
+        }
+      }).body
+    }
 
-      const mockResponse = (status, statusText, response) => {
-        return new Response(response, {
-          status: status,
-          statusText: statusText,
-          headers: {
-            'Content-type': 'application/json'
-          }
-        }).body
-      }
-
-      const bodyContains = {
-        response: {
-          body: {
-            contains: [
-              'ld4p', 'pcc'
-            ]
-          }
+    const bodyContains = {
+      response: {
+        body: {
+          contains: [
+            'ld4p', 'pcc'
+          ]
         }
       }
+    }
+
+    it('sets the state with group data (as Array) from sinopia_server', async() => {
 
       const promise = Promise.resolve(mockResponse(200, null, bodyContains))
       const wrapper2 = shallow(<SinopiaResourceTemplates />)
       await wrapper2.instance().fulfillGroupPromise(promise)
       wrapper2.update()
       expect(wrapper2.state('groupData')).toEqual(['ld4p', 'pcc'])
+    })
+
+    it('sets a message if there is no server response', async() => {
+      const promise = Promise.resolve(mockResponse(200, null, undefined))
+      const wrapper2 = shallow(<SinopiaResourceTemplates />)
+      await wrapper2.instance().fulfillGroupPromise(promise)
+      await wrapper2.update()
+      expect(wrapper2.state('message')).toBeTruthy()
     })
 
     it('sets the state with a list of resource templates from the server', async() => {
--- a/__tests__/components/editor/SinopiaResourceTemplates.test.js
+++ b/__tests__/components/editor/SinopiaResourceTemplates.test.js
@@ -47,6 +47,7 @@ describe('<SinopiaResourceTemplates />', () => {
       const promise = Promise.resolve(mockResponse(200, null, bodyContains))
       const wrapper2 = shallow(<SinopiaResourceTemplates />)
       await wrapper2.instance().fulfillGroupPromise(promise)
+      wrapper2.update()
       expect(wrapper2.state('groupData')).toEqual(['ld4p'])
     })
 
@@ -69,6 +70,7 @@ describe('<SinopiaResourceTemplates />', () => {
       //TODO: figure out how to mock and test a nested promise...
       const spy = jest.spyOn(wrapper3.instance(), 'fulfillGroupDataPromise')
       await wrapper3.instance().fulfillGroupDataPromise(promise)
+      wrapper3.update()
       expect(spy).toHaveBeenCalled()
     })
 
--- a/__tests__/components/editor/SinopiaResourceTemplates.test.js
+++ b/__tests__/components/editor/SinopiaResourceTemplates.test.js
@@ -18,7 +18,7 @@ describe('<SinopiaResourceTemplates />', () => {
   })
 
   it('has a bootstrap table that displays the results from the calls to sinopia_server', () => {
-    expect(wrapper.find(BootstrapTable).dive().length).toEqual(1)
+    expect(wrapper.find('BootstrapTable').length).toEqual(1)
   })
 
   describe('getting data from the sinopia_server', () => {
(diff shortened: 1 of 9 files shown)


__tests__/components/editor/SinopiaResourceTemplates.test.js
```diff
@@ -0,0 +1,79 @@
...
+      }
+
+      const promise = Promise.resolve(mockResponse(200, null, bodyContains))
+      const wrapper3 = shallow(<SinopiaResourceTemplates />)
+      //TODO: figure out how to mock and test a nested promise...
+      const spy = jest.spyOn(wrapper3.instance(), 'fulfillGroupDataPromise')
+      await wrapper3.instance().fulfillGroupDataPromise(promise)
+      wrapper3.update()
+      expect(spy).toHaveBeenCalled()
+    })
```
ndushay: in a perfect world, would we not want a test for the error case, where the templates are NOT populated?

src/components/editor/SinopiaResourceTemplates.jsx
```diff
@@ -0,0 +1,115 @@
...
+    })
+  }
+
+
+  fulfillGroupPromise = (promise) => {
+    promise.then((data) => {
+      const contains = [].concat(data.response.body.contains)
+      this.setState({groupData: contains})
+    }).catch((error) => {
+      this.setState({message: error})
```
ndushay: this is the line that may not be tested ... and does display something to end user.

src/components/editor/SinopiaResourceTemplates.jsx
```diff
@@ -0,0 +1,115 @@
...
+      isLoading: false,
+      groupData: [],
+      templatesForGroup: []
+    }
+  }
+
+  async componentDidMount() {
+    const groupPromise = new Promise((resolve) => {
+      resolve(instance.getBaseWithHttpInfo())
+    }).then((data) => data).catch(() => {})
```
ndushay: (trying to understand what I always screw up:  

- so this new promise is necessary b/c we need to silence the error from instance.getBaseWithHttpInfo() with the empty catch block, but still pass to fulfillGroupPromise() regardless?

src/components/editor/SinopiaResourceTemplates.jsx
```diff
@@ -0,0 +1,115 @@
...
+  constructor(props) {
+    super(props)
+    this.state = {
+      isLoading: false,
+      groupData: [],
+      templatesForGroup: []
+    }
+  }
+
+  async componentDidMount() {
```
mjgiarlo: Take this with a grain of salt, since I'm the newest to JavaScript, but I am finding lines 19 to 68 difficult to understand. Probably because of all the promise machinery. I wonder how much work it would take to rework this into a more declarative style using `async`/`await`, like the following:

```javascript
async componentDidMount() {
  try {
    const data = await instance.getBaseWithHttpInfo()
    const contains = [].concat(data.response.body.contains)
    this.setState({groupData: contains})
  } catch(error) {
    this.setState({message: error})
  }

  const groupData = this.state.groupData.map(async group => {
    const name = this.resourceToName(group)
    await instance.getGroupWithHttpInfo(name)
  })

  groupData.forEach(data => {
    try {
      let joined = []
      const groupName = this.resourceToName(data.response.body['@id'])
      data.response.body.contains.map(async c => {
        const name = this.resourceToName(c)
        const response_and_data = await instance.getResourceWithHttpInfo(groupName, name, { acceptEncoding: 'application/json' })
        joined.push({name: name, uri: c, id: response_and_data.response.body.id, group: 'ld4p'})
        this.setState({templatesForGroup: joined})
      })
    } catch(error) {
      this.setState({message: error})
    }
  })
}
```

I'm not sure this is exactly right; it's probably off by a fair bit, actually. But I put it forward as a comparison re: readability.

Again, grain of salt!

__tests__/components/editor/SinopiaResourceTemplates.test.js
```diff
@@ -0,0 +1,79 @@
...
+
+  it('has a header for the area where the table of resource templates for the groups are displayed', () => {
+    expect(wrapper.find('div > h4').last().text()).toEqual('Available Resource Templates in Sinopia')
+  })
+
+  it('has a bootstrap table that displays the results from the calls to sinopia_server', () => {
+    expect(wrapper.find(BootstrapTable).dive().length).toEqual(1)
+  })
+
+  describe('gettiong data from the sinopia_server', () => {
```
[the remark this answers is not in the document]
astridu: @jermnelson You are correct-- The user would click on the Resource Template to select it. I updated my mock up to make the links look blue so it's clear.

src/components/editor/SinopiaResourceTemplates.jsx
jgreben: Yes to the former. we need to silence the error from instance.getBaseWithHttpInfo() with the empty catch block, but still pass to fulfillGroupPromise() regardless. Otherwise the integration test fails.

src/components/editor/SinopiaResourceTemplates.jsx
jgreben: @mjgiarlo I tried a little of what you were getting at, but this ends up being a promise that needs to be resolved:
```
const groupData = this.state.groupData.map(async group => {
      const name = this.resourceToName(group)
      await instance.getGroupWithHttpInfo(name)
    })
```

If you want to pair on Monday and see if we can work it out further, i'd be down for that!

__tests__/components/editor/SinopiaResourceTemplates.test.js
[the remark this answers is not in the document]
jgreben: Fixed typo, and I think adding a link should be a follow-on PR for sure.

src/components/editor/SinopiaResourceTemplates.jsx
jgreben: test added

__tests__/components/editor/SinopiaResourceTemplates.test.js
jgreben: I added a test for that.

__tests__/components/editor/SinopiaResourceTemplates.test.js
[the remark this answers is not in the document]
jermnelson: Created ticket [#436](https://github.com/LD4P/sinopia_editor/issues/436) to add links in the `SinopiaResourceTemplate` component.

src/components/editor/SinopiaResourceTemplates.jsx
mjgiarlo: Sorry not to drop this message earlier but LDCX intervened. I have time this afternoon and tomorrow, if still helpful to chat.

jgreben: Thanks @mjgiarlo I hacked on this a little more, and it seems to me while it is possible to make the code more readable, it becomes much less testable. I got testability out of making those promises messages that could be passed into the parsing functions. That way I could mock the same promise/message as: 

`const promise = Promise.resolve(mockResponse(200, null, bodyContains))`

Is readability more important than testability? Maybe I'm making a false distinction because I'm just not familiar with how to mock and test a const in `componentDidMount()` ¯\_(ツ)_/¯

mjgiarlo: Naw, that's debateable, and like I said... grain of salt! I am not blocking this PR, given my lack of familiarity with React and Javascript software design norms.

mjgiarlo: Re: testability, my druthers are usually to create small methods/functions and then to mock those out as needed in tests. Since what you have already works though... :slightly_smiling_face:

jgreben:
@jermnelson made issue https://github.com/LD4P/sinopia_editor/issues/438

jgreben:
@jermnelson added the start of a link to the Editor component in a way supported by bootstrap-table. Would something like this be ok re: design @astridu?

Screen Shot 2019-03-28 at 10 10 32 AM

astridu:
It's not a typical UI pattern, but it will work. We may need to put some instruction at the top of the table: "Click the radio button to select the resource template you want to use."

jgreben:
@astridu Thanks, but I think I found a way to make it work with normal links. Stay tuned...

Issue #338, opened 2019-02-07
Finalize Help & Resources links on Sinopia home page


michelleif:
1) Users: check that all necessary resources are listed, and in desired order / grouping
2) Users: check that sure all items link to proper URL
3) Users: make sure Help & Resources sidebar is accessible from any Sinopia page
4) dev team: fix anything identifed in #1-3
(sticking this in Milestone 4)

michelleif:
@jermnelson I just realized that the "Help & Resources" tab on the Profile Editor is *not* pointing to the latest Help & Resources content. If you compare the sidebar that comes up from sinopia.io or sinopia.io/editor to the one that comes up from profile-editor.sinopia.io you'll see the difference. Make a new ticket for this?

jermnelson:
Yes, I think we should create a new ticket in the PE repo at https://github.com/ld4p/sinopia_profile_editor/issues.

michelleif:
menu is working and I verified I can easily make changes to the menu by editing these files: https://github.com/LD4P/sinopia/tree/master/help_and_resources

PR #434, opened 2019-03-22
revised sidebar links

michelleif:
-revised sidebar links according to latest user designs
-made the link headings non-clickable


michelleif:
connects to #338

Issue #348, opened 2019-02-08
lookups to multiple sources per field


michelleif:
PropertyType = lookup can be configured to have more than one "Value" (selection from the dropdown list of QA configs). If Profile specifies that lookup should search more than one QA config, then Editor needs to search both of those sources and present results from both. - [ ] Search more than one QA source at a time if specified in Profile.

See LC BFE Monograph/Work/Subject/SearchSubjects/Search LCSH/LCNAF
The "Search LCSH/LCNAF" field searches LCNAF and LCSH, and presents results in two sets.

michelleif:
screenshot showing LC Profile Editor with two lookup sources for the field:
![Profilewithmultiplelookupsinfield.png](https://waffleio-direct-uploads-production.s3.amazonaws.com/uploads/581116c7f17930130007d3eb/125516c66e82c728ace21e0d46d88695258b8fac98e0ab4ba948fc8020562b76250bac206595f021b1447204030b5cb053181d1ba6acda6fb4bc2e75d85d0ae8d66b0e22c63f16f8ba821fbe52e8268f4a87b3ae965e82b96371b8d6d6056b5c17129bfe553fa19c4fc15dff4be14553a01a3671ab.png)

screenshot showing how this looks when profile is used in BFE and user does a search
![multiplelookup.png](https://waffleio-direct-uploads-production.s3.amazonaws.com/uploads/581116c7f17930130007d3eb/125516c66e82c728ace21e0d46e58196378b93a58ae5b04caf48e0d665043761260ea62864c7b96aea402a505a5848fd17585d43a8fb892ba7e56e768e5909e8c3714d36976345f3e5db55ea43a82a994e8dbcbd8f48c9ed3425e783894a3e1e1755.png)

jgreben:
The way to handle this (I think) will be to federate the query and combine the responses.

jermnelson:
Closing with merge to master

PR #439, opened 2019-03-28
Multiplelookups 348

hudajkhan:
Addresses #348 .
Updates to resource template form enable passing back the array of lookup configurations (and not just one) so more than one lookup can be supported.
Options for menu now include an array of objects where each object includes the label/uri for the lookup as well as a "body" element that is the array of the results for that lookup.
Promises.all used to generate a promise per lookup requested, with catch blocks attached to each promise so that if one fails, the results of the others will still be passed along.
In the case where more than one lookup is used, titles are attached to the dropdown (using Menu.Header) from the lookup configuration. 
renderMenu is overridden customized for the typeahead component.
When a lookup returns no results or if there is an error that occurs, a message is appended under the appropriate heading (if more than one lookup source) or just directly (if only one lookup source).
Custom styling added to menu.css for dropdown header and some custom classes for error/empty result situations. 
Added one very basic test that just checks multiple values are available in state on search once saved there.  Would appreciate more input regarding tests!


hudajkhan:
Added some more tests

src/styles/main.css
```diff
@@ -171,3 +171,13 @@ input.rbt-input-main {
 .editor-navtabs a {
     color: #2F2424;
 }
+
+/** For lookup menu headers **/
+.dropdown-header {
+	font-weight:800;
+	color:#333;
```
jgreben: I think we should add a background color, perhaps `background-color: lightgray;`?

jgreben:
Looks good to me! I just think we need a background color to the multiple results header to make it pop a bit more. Also, it wouldn't hurt to add a second lookup ("urn:ld4p:qa:subjects:person") to the MonographInstance spoof file in order to have something readily at hand to see the functionality in action (for demo purposes):

src/styles/main.css
hudajkhan: Added background-color #d3d3d3 and updated the spoof file.

hudajkhan:
I "resolved" the conversation.  I don't think I've ever done that before. Cool : )

__tests__/components/editor/InputLookupQA.test.js
```diff
@@ -82,4 +114,38 @@ describe('<InputLookup />', () => {
...
+	   expect(multipleWrapper.state().options[1]).toEqual(multipleResults[1]) 
+
+  })
+  //Headers expected
+
+  it('should show menu headers with lookup source labels and values in the dropdown when provided results', () => {
+	  console.log("Wrapper instance");
+	  const instance = multipleWrapper.instance();
+	  const menuWrapper = shallow(instance.renderMenuFunc(multipleResults, p2Props));
+	  //console.log(menuWrapper.childAt(0).render().html());
```
jermnelson: I would remove these commented-out log statements.

jermnelson:
Hi @hudajkhan, I made one small comment; looks good to me!

__tests__/components/editor/InputLookupQA.test.js
hudajkhan: I've removed the commented out statements as well the first console.log which was just there for debugging the test.

PR #443, opened 2019-04-04
Resourcelabel 393

hudajkhan:
Shifted the code around that seemed specific to the resource template (i.e. the resource template ID/URI and label) to the resource template component.  At first, I had the preview RDF button still in the Editor component but moved the entire section with that and RDF Modal into ResourceTemplate.  Added a constructor to ResourceTemplate just so it can have state and showRDF (notice these are set in ResourceTemplateForm).  

The placement of preview rdf should match the original, but the label and URI are now being retrieved from the resource template data.

Tests have been updated to reflect shifting of content across components.

### DIFF
--- a/__tests__/components/editor/Editor.test.js
+++ b/__tests__/components/editor/Editor.test.js
@@ -28,10 +28,6 @@ describe('<Editor />', () => {
       expect(wrapper.find(Header).length).toBe(1)
     })
 
-    it('shows resource title', () => {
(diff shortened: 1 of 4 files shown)


hudajkhan:
I originally moved props.rtId to resourceTemplate since it was being used original in the Editor component with  rtId={this.props.rtId} in RDFModal but then I saw rtId being passed to ResourceTemplateForm using the actual data for resource template using rt.id so used that instead.

jermnelson:
Looks good to me; merging into master.

PR #446, opened 2019-04-04
sending multiple lookup config items array to input qa

hudajkhan:
Updating PropertyTemplateOutline to retrieve and send array of lookup config items (instead of just the first one) to the input lookup qa component as expected.  Function for getting just the first lookup (for input list loc) and all lookup configs both employ shared code for retrieving lookup configuration based on a template uri.


jermnelson:
LGTM; merging into master.

PR #447, opened 2019-04-10
Update schema version info message when importing a new RT/profile

### LINKED ISSUE #396: update schema version info message in Linked Data Editor?

mjgiarlo:
Fixes #396

This introduces a new env-overrideable config variable setting the default RT/Profile schema version. 

~~Note that I was unable to determine whether 0.0.9 is the right default for now. #396 specifies 0.0.9 so that is what I went with, but 0.1.0 seems to be the newest version available~~ I... *think* I confirmed on Slack this morning that 0.0.9 is the right default for the time being.

~~FWIW, I attempted to upload a few seemingly valid 0.1.0 and 0.0.9 profiles and I was unable to get *any* to validate. I am not sure whether that should be in scope for this issue but I would at least like to understand why this is the case.~~ Confirmed with @michelleif that https://github.com/LD4P/sinopia_sample_profiles/blob/master/configuration_demo/SinopiaConfigurationDemo.json is the most reliable profile to test, and I can confirm it validates in this branch.

### DIFF
--- a/__tests__/integration/schemaValidation.test.js
+++ b/__tests__/integration/schemaValidation.test.js
@@ -58,7 +58,7 @@ describe('Editor', () => {
           // const dialog = await pupExpect(page).toDisplayDialog(async () => {
           //   await fileInput.uploadFile("__tests__/__fixtures__/lcc_v0.0.2.json")
           // })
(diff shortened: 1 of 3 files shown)


jgreben:
LGTM!

PR #460, opened 2019-04-18
Manage sessions using amplify sdk

jmartin-sul:
This PR reworks Cognito session management in the editor such that Amazon's Amplify SDK (specifically the amazon-cognito-identity-js subpackage) is used for:
* login (via a form directly in the application, instead of redirection through the Cognito hosted login page)
* session persistence between page loads (instead of manual wrangling of local storage)
* session refresh, i.e. obtaining a new valid JWT ID token as needed, as long as a valid refresh token is present for the user (which was not implemented before, and was ultimately the impetus for this work, which would've been much more difficult and messy without the Amplify SDK)

React and Redux have been very pleasant to deal with.  I like the documentation on both projects.

Note that while I've fixed the existing tests that were caused to fail by these changes, I haven't fully implemented all of the new tests that are needed.  In particular, I haven't written unit tests for `CognitoUtils` (which actually does get exercised by the integration test that goes through a full login for real, but integration tests don't count towards reported test coverage, hence the reported drop in coverage).

I will go through this PR and leave some explanations inline, as well as links to tickets I file for follow on work.

closes LD4P/sinopia_server#89

PR #461, opened 2019-04-23
Update resources and warn users

### LINKED ISSUE #454: Warn User when uploading a profile with an existing ID.
When a profile with an existing ID is uploaded, sinopia_server will return a 409 (Conflict) when attempting the `createResourceWithHttpInfo` api call. At this point we will want to:
- Have them confirm whether they want to do an update (overwrite) of the resource
- do a `updateResourceWithHttpInfo(groupID, resourceID, resource, opts)` api call

jgreben:
Update resources when a trellis response indicates a conflict and warn user in rendered modal.

Fixes #454

This PR includes:
 
- Creates `UpdateResourceModal` that appears when a 409/Conflict error is caught from a POST call to trellis with an existing resource ID. The modal captures the 409 response and renders some details about it, and then returns the conflicting template data to a callback function that handles the update (PUT) call instead.
- `ImportResourceTemplate` provides the callback function to the modal and handles the trellis call to do the updates, setting the component state and props as needed
- `SinopiaResourceTemplates` should handle an array of contents as well as a single content as a string. Refactors promise for better testing
- (minor change) Decrease the timeout before being sent to Cognito to login (this will be obviated by forthcoming PR #460) 
- Test changes, additions, and refactoring of components to maintain coverage

### DIFF
--- a/src/components/editor/UpdateResourceModal.jsx
+++ b/src/components/editor/UpdateResourceModal.jsx
@@ -1,0 +1,73 @@
+// Copyright 2019 Stanford University see Apache2.txt for license
+
+import React, {Component} from 'react'
+import PropTypes from 'prop-types'
+import Button from 'react-bootstrap/lib/Button'
+import Modal from 'react-bootstrap/lib/Modal'
+const _ = require('lodash')
+
+class UpdateResourceModal extends Component {
+  constructor(props) {
+    super(props)
+    this.state = {
+      titleMessage: [],
+      rts: []
+    }
+  }
+
+  resourceToName = (resource) => {
+    const idx = resource.lastIndexOf('/')
+    return resource.substring(idx+1)
+  }
+
+  componentDidMount() {
+    let group = ''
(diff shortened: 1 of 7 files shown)


src/components/editor/UpdateResourceModal.jsx
```diff
@@ -0,0 +1,73 @@
...
+        <Modal show={this.props.show}>
+          <Modal.Header>
+            <Modal.Title>{this.state.titleMessage}</Modal.Title>
+          </Modal.Header>
+          <Modal.Body>
+            Do you want to overwrite these resource templates?
+          </Modal.Body>
+          <Modal.Footer>
+            <Button onClick={() => {this.props.update(this.state.rts, this.state.group)}}>Yes, overwite</Button>
+            <Button onClick={this.props.close}>No, get me out of here!</Button>
```
jermnelson: Error message seems a bit too informal but fine for now.

Issue #473, opened 2019-04-26
Determine what to do with linked data reducer


mjgiarlo:
In particular, subsequent to #466, the LD reducer currently **only** works with spoofed resource templates from `sinopiaServer` and will need to use `async`/`await` or promises to deal with resource templates from the server.

Look for related `TODO`s in the code, and look into removing mocks of `Config.spoofSinopiaServer`—*e.g.*, `jest.spyOn(Config, 'spoofSinopiaServer', 'get')...`—as part of this work.

PR #489, opened 2019-04-30
Support in Editor for Server Spoofing

jermnelson:
If the local environment SPOOF_SINOPIA_SERVER variable is set to true, the editor will now return a Promise from reading the local file system instead of doing a network call to either a remote or local version of the Sinopia Server running Trellis. 

As @mjgiarlo noted in ticket #473, the `linkedData` reducer expected a resourceTemplate and not a Promise. Since we're going a different route to generate the RDF based on the Redux state, I just removed the `linkedData` reducer and test.


jgreben:
LGTM!

mjgiarlo:
@jermnelson Nicely done!

PR #491, opened 2019-05-01
Going directly to editor route redirects to the "Resource Templates" tab

### LINKED ISSUE #480: When user clicks "Linked Data Editor" or goes directly to sinopia.io/editor, take them to the "Resource Templates" tab
in order to catalog, user has to first select a template, which is done on the Resource Templates tab, so take the user to this tab, not to Editor tab, as is current behavior.

see also #479

jgreben:
Fixes #480 

- Redirect to Import Templates if there is no resourceTemplateId in state (i.e. no template to edit with yet)
- Spoof all of the local templates in the list of available resource templates (because we no longer will be loading a default RT for testing)

### DIFF
--- a/src/components/editor/SinopiaResourceTemplates.jsx
+++ b/src/components/editor/SinopiaResourceTemplates.jsx
@@ -5,7 +5,8 @@ import PropTypes from 'prop-types'
 import { Link } from 'react-router-dom'
 import {BootstrapTable, TableHeaderColumn} from 'react-bootstrap-table'
 import Config from '../../../src/Config'
-import { getGroups, listReourcesInGroupContainer, getResourceTemplate } from '../../sinopiaServer'
+import { getGroups, listResourcesInGroupContainer,
+         getResourceTemplate, listResourcesInGroupContainer } from '../../sinopiaServer'
 const _ = require('lodash')
 const SinopiaServer = require('sinopia_server')
 const instance = new SinopiaServer.LDPApi()
@@ -31,7 +32,7 @@ class SinopiaResourceTemplates extends Component {
       await this.fulfillGroupPromise(groupPromise).then(async () => {
         this.state.groupData.map(group => {
           const groupName = this.resourceToName(group)
-          listReourcesInGroupContainer(groupName).then((data) => {
+          listResourcesInGroupContainer(groupName).then((data) => {
             this.fulfillGroupData(data)
           }).catch(() => {})
         })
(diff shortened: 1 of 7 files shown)


jgreben:
This PR does not address the concern about non-authenticated users being able to see something without logging in. IMO I think we should still merge this PR and then revisit the issue of when and for what logging in should happen (and I will be glad to add the topic to the storytime doc.

src/sinopiaServer.js
```diff
@@ -95,3 +111,31 @@ export const getResourceTemplate = (templateId, group) => {
...
+        }
+      })
+    })
+
+  return new Promise(resolve => {
+    resolve(instance.getBaseWithHttpInfo())
+  })
+}
+
+export const listReourcesInGroupContainer = (group) => {
```
jermnelson: Misspelling of function name, should this be `listResourcesInGroupContainer` instead?

src/components/editor/SinopiaResourceTemplates.jsx
```diff
@@ -23,18 +24,14 @@ class SinopiaResourceTemplates extends Component {
...
       }).catch(() => {})
 
       await this.fulfillGroupPromise(groupPromise).then(async () => {
-        this.state.groupData.map((group) => {
+        this.state.groupData.map(group => {
           const groupName = this.resourceToName(group)
-          new Promise((resolve) => {
-            resolve(instance.getGroupWithHttpInfo(groupName))
-          }).then((data) => {
+          listReourcesInGroupContainer(groupName).then((data) => {
```
jermnelson: This will need to be updated with corrected function name

src/components/editor/SinopiaResourceTemplates.jsx
```diff
@@ -5,6 +5,7 @@ import PropTypes from 'prop-types'
 import { Link } from 'react-router-dom'
 import {BootstrapTable, TableHeaderColumn} from 'react-bootstrap-table'
 import Config from '../../../src/Config'
+import { getGroups, listReourcesInGroupContainer, getResourceTemplate } from '../../sinopiaServer'
```
jermnelson: `listReourcesInGroupContainer` import will need to change with correct function name.

src/sinopiaServer.js
jgreben: Ha! At least it was internally consistent! Fixed it.

jgreben:
Created issue #495 to address the new authentication needs when this PR is merged.

PR #499, opened 2019-05-03
ResourceTemplateForm.test: skip tests giving false positives

ndushay:
These are just like the false positives I found for PropertyTemplateOutline.test.  I put "expect(1).toBe(2)" in the blocks with expect statements, and these tests passed.  :-(

I also corrected the misspelling of fulfill and got rid of unused imports.

### DIFF
--- a/__tests__/components/editor/ResourceTemplateForm.test.js
+++ b/__tests__/components/editor/ResourceTemplateForm.test.js
@@ -184,7 +184,7 @@ describe('<ResourceTemplateForm /> after fetching data from sinopia server', () 
       //   }
       // }
       // const instance = wrapper.instance()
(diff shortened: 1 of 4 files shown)


__tests__/components/editor/ResourceTemplateForm.test.js
```diff
@@ -2,9 +2,7 @@
 
 import React from 'react'
 import 'jsdom-global/register'
-import ButtonToolbar from 'react-bootstrap/lib/ButtonToolbar'
-import { mount, shallow } from 'enzyme'
-import InputLiteral from '../../../src/components/editor/InputLiteral'
```
ndushay: I got rid of unused imports

Issue #496, opened 2019-05-02
PropertyTemplateOutline, ResourceTemplateForm False Positive Tests Clean-up 


jermnelson:
Remove  or refactor the following **PropertyTemplateOutline.test.js** tests:

All use an approach that is a failed attempted to 'inject' info to an inner method -- attempting to mimic the promises for a particular property template, I believe

#### PropertyTemplateOutline.test.js

- [clicking removes collapsed state](https://github.com/LD4P/sinopia_editor/blob/af0f81ecfad3534ebec3e9cb1484a421f8182a4f/__tests__/components/editor/PropertyTemplateOutline.test.js#L294)
    - "collapsed" isn't a state, it's a prop.  Maybe it should be a state
    - in order to examine STATE after simulate, you must do a fresh "find" from the root component (https://github.com/airbnb/enzyme/issues/1229#issuecomment-462496246)
    - I believe for this to work, mockHandleCollapsed has to mock the implementation ... which leads me to believe that this should be an integration test

- [handles "Add" button click](https://github.com/LD4P/sinopia_editor/blob/af0f81ecfad3534ebec3e9cb1484a421f8182a4f/__tests__/components/editor/PropertyTemplateOutline.test.js#L304)
    - wrapper has OutlineHeader but no PropertyActionButtons, no PropertyTypeRow
    - it needs to be expanded first?  implying an integration test

- [handles "Mint URI" button click](https://github.com/LD4P/sinopia_editor/blob/af0f81ecfad3534ebec3e9cb1484a421f8182a4f/__tests__/components/editor/PropertyTemplateOutline.test.js#L320)
    - wrapper has OutlineHeader but no PropertyActionButtons, no PropertyTypeRow
    - it needs to be expanded first?  implying an integration test

#### ResourceTemplateForm.test.js

- 'renders a lookup component'
- 'renders a list component'
- 'renders InputLiteral nested component (b/c we have a property of type "literal")'
- 'renders error alert box'

PR #503, opened 2019-05-06
Make false positives fail and show the pattern that avoids introducing them

mjgiarlo:
connects to #496 -- this is not a complete fix for #496, just a start to show the pattern.  

Naomi sez:

>Once we all check it off, we can turn this into a real PR and also know what to do for #496 to be done.  

>Note that this is showing a technique and may not be sufficient for the affected tests to pass -- it may only make them fail "correctly".

>Unfortunately I think we have a bunch of tests like this (I wrote some of them!) and we all (me too!) need to understand how to avoid them in the future.

>For example, in trying to get the profile editor build working, i'm referring back to this PR so I can make sure any tests that are problematic (e.g. taking too long or throwing errors in circleci) don't have these problems, too.

- [x] @ndushay 
- [x] @jermnelson 
- [x] @jmartin-sul 
- [x] @jgreben 
- [x] @hudajkhan

### DIFF
--- a/__tests__/components/App.test.js
+++ b/__tests__/components/App.test.js
@@ -51,7 +51,7 @@ describe("#routes", () => {
 
     it('root renders HomePage component', async () => {
       const component = renderRoutes("/")
-      await expect(component.find(HomePage).length).toEqual(1)
+      return await expect(component.find(HomePage).length).toEqual(1)
     })
 
     it('/editor renders Editor component', () => {
--- a/__tests__/components/editor/ImportFileZone.test.js
+++ b/__tests__/components/editor/ImportFileZone.test.js
@@ -85,7 +85,7 @@ describe('<ImportFileZone />', () => {
     })
 
     it('displays an error message when missing required property', async () => {
-      await wrapper.instance().promiseTemplateValidated(template, schemaUrl).catch((err) => {
+      return await wrapper.instance().promiseTemplateValidated(template, schemaUrl).catch((err) => {
         expect(wrapper.state().validTemplate).toBeFalsy()
         expect(err.toString()).toMatch("should have required property")
       })
@@ -98,7 +98,7 @@ describe('<ImportFileZone />', () => {
     })
 
     it('displays an error message when the id is invalid', async () => {
-      await wrapper.instance().promiseTemplateValidated(template, schemaUrl).catch((err) => {
+      return await wrapper.instance().promiseTemplateValidated(template, schemaUrl).catch((err) => {
         expect(wrapper.state().validTemplate).toBeFalsy()
         expect(err.toString()).toMatch('should match pattern')
       })
@@ -117,7 +117,7 @@ describe('<ImportFileZone />', () => {
     })
 
     it('displays an error message', async () => {
-      await wrapper.instance().promiseTemplateValidated(template, schemaUrl).catch((err) => {
+      return await wrapper.instance().promiseTemplateValidated(template, schemaUrl).catch((err) => {
         expect(wrapper.state().validTemplate).toBeFalsy()
         expect(err.toString()).toMatch("Error: error getting json schemas")
       })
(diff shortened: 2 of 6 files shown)


jmartin-sul:
@mjgiarlo, would you be able to give a bit more explanation as to what's going on in principle in the adjusted versions of the tests?  e.g., is it that the `return` is necessary when the expectation is (or is inside of) an `await`ed `Promise`?  does that imply that the test's success or failure is determined by the return value of the test (thus also implying that the last expectation in the test is all that matters as far as test passage is concerned)?  does that also imply that `expect` returns a value in some cases, and a `Promise` in others (e.g. the first change in the PR results in `return await expect(component.find(HomePage).length).toEqual(1)`, but in other parts of the PR, the `await`ed operation is a promise with an expectation in the `catch` block).

apologies if these are all super basic questions, still a noob to both JS promises and Jest.  if there are docs i could be pointed to, i'd also happily do some reading (even if the answer is just something like "read the basic Jest docs").

i read through the PR, but i'll defer checking myself off till i actually understand it 😉

mjgiarlo:
@jmartin-sul :speech_balloon: 
> @mjgiarlo, would you be able to give a bit more explanation as to what's going on in principle in the adjusted versions of the tests? e.g., is it that the `return` is necessary when the expectation is (or is inside of) an `await`ed `Promise`?

Yes, exactly that. https://jestjs.io/docs/en/asynchronous#promises Without the `return` the test will pass everytime. Particularly if we are also eating errors via `.catch()` calls.

> does that imply that the test's success or failure is determined by the return value of the test (thus also implying that the last expectation in the test is all that matters as far as test passage is concerned)? does that also imply that `expect` returns a value in some cases, and a `Promise` in others (e.g. the first change in the PR results in `return await expect(component.find(HomePage).length).toEqual(1)`, but in other parts of the PR, the `await`ed operation is a promise with an expectation in the `catch` block).

The `return` technique is primarily for expectations that occur within a Promise, either a `.then()` call or a `.catch()` call. That doesn't mean you can't have other expectations *before* the promise

I also notice that a number of our tests appear to be using both `await` *and* a promise, which I don't entirely understand. (Those tests begin failing once applying the `return` per Jest documentation recommendations, so I'm wondering whether those are also ripe for simplification.)

> apologies if these are all super basic questions, still a noob to both JS promises and Jest. if there are docs i could be pointed to, i'd also happily do some reading (even if the answer is just something like "read the basic Jest docs").

Not super basic! And I'm more of a noob than any of y'all, so no worries there. With that said, take what I say here with a grain of salt. FWIW, it is consistent with my experience testing promise-based functionality in the indexing pipeline, but beyond that... yeah.
 
> i read through the PR, but i'll defer checking myself off till i actually understand it

Sounds like you do!

jermnelson:
I'm glad that the false positives with Promises in the tests are being surfaced in this PR but I'm troubled with the tone of the PR, particularly the check-off for each developer, as being somewhat patronizing especially the phrase "Please check off after you have learned the lesson here".

ndushay:
@jermnelson that's on me - sorry.  I said it badly -- I was trying to say something more like ... unfortunately I think we have a lot of tests like this (I wrote some of them!) and we all (me too!) need to understand how to avoid them in the future.

ndushay:
I'm finding that I prefer this approach, as I don't have to remember when I do and don't need a return statement.  However, if there is a `.then()` block, you still need `return`.

```js
  it('modal title says Choose Vocab Template', async () => {
    expect.assertions(1)
    const modalTitleSel = `${modalSel} > div.modal-header > h3.modal-title`
    await expect_value_in_sel_textContent(modalTitleSel, 'Choose Vocab Template')
  })
```

Also, Green-Red-Green is always useful.  :-)   I have been using 

```js
expect(2).toBe(1)
```

In spots to ensure test code is executed, and THEN change the code itself  or the test code to deliberately break and look at results for failure I expected.

mjgiarlo:
We've all internalized this, AFAICT. Discussed on Slack today and am closing this issue.

PR #513, opened 2019-05-07
Inputlistloc selector reducer

### LINKED ISSUE #430: InputListLOC uses selectorReducer

jgreben:
Fixes #430 

- Renames `literal` reducer to `inputs` because it is now being used by more than one input component
- Add a redux function to set the selections from the typeahead option list. Selections are overwritten every time with everything in the `selected` array because the typeahead component keeps track of what's in there
- Reorder the properties in the default payload to match what's set when selections are made
- Remove the un-used rtId and add the reduxPath as needed by the new selectorReducer way of setting the redux state
- Whether the typeahead component allows multiple selections should come from the repeatable property of the valueConstraint and not the property template.

### DIFF
--- a/src/reducers/inputs.js
+++ b/src/reducers/inputs.js
@@ -37,7 +37,7 @@ export const setMyItems = (state, action) => {
 }
 
 export const setMySelections = (state, action) => {
-  let newState = Object.assign({}, state)
+  const newState = Object.assign({}, state)
   const reduxPath = action.payload.reduxPath
   let level = 0
   reduxPath.reduce((obj, key) => {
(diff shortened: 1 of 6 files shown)


src/reducers/inputs.js
```diff
@@ -36,6 +36,23 @@ export const setMyItems = (state, action) => {
   return newState
 }
 
+export const setMySelections = (state, action) => {
+  let newState = Object.assign({}, state)
```
jermnelson: I think this can be a `const` instead

jgreben: 👍 Changed it.

PR #515, opened 2019-05-08
Handles missing valueConstraint and valueTemplateRef in PropertyTemplateOutline

### LINKED ISSUE #509: Clicking the plus to expand a property with a nested resource does not work
receiving console error `TypeError: undefined is not an object (evaluating 'e.map')`

Jeremy believes #493 (merged PR) may solve this

jermnelson:
Adds check for missing valueConstraint and valueConstraint.valueTemplateRefs when user clicks on the plus.

Fixes #509

Todo:
- [x] Test for specific Resource Templates that have propertyTemplates missing `valueConstraint` and `valueConstraint.valueTemplateRefs`

### DIFF
--- a/src/components/editor/PropertyTemplateOutline.jsx
+++ b/src/components/editor/PropertyTemplateOutline.jsx
@@ -180,12 +180,7 @@ export class PropertyTemplateOutline extends Component {
   handleClick = (property) => (event) => {
     event.preventDefault()
     let newOutput = this.state.output
-    const templateRefList = []
-    if(hasValueTemplateRef(property)) {
-      property.valueConstraint.valueTemplateRefs.map((row) => {
-        templateRefList.push(row)
-      })
-    }
+    const templateRefList = hasValueTemplateRef(property) ? property.valueConstraint.valueTemplateRefs : []
     this.fulfillRTPromises(this.resourceTemplatePromises(templateRefList))
       .then(() => {
         const propertyJsx = this.propertyComponentJsx(property)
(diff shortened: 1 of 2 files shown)


src/components/editor/PropertyTemplateOutline.jsx
```diff
@@ -180,8 +180,13 @@ export class PropertyTemplateOutline extends Component {
   handleClick = (property) => (event) => {
     event.preventDefault()
     let newOutput = this.state.output
-
-    this.fulfillRTPromises(this.resourceTemplatePromises(property.valueConstraint.valueTemplateRefs))
+    const templateRefList = []
+    if(hasValueTemplateRef(property)) {
+      property.valueConstraint.valueTemplateRefs.map((row) => {
```
mjgiarlo: I'm a bit confused about what this is doing. Is the intent of this PR to add a check before sending a value to `this.resourceTemplatePromises()`?

Is this code essentially creating a new variable to hold an exact copy of an array? If so, I'm not sure `map` is the clearest solution unless there is a function you want to apply to every member of the array. The intent would be clearer, in my judgment, if you simplified this to:

```javascript
const templateRefList = hasValueTemplateRef(property) ? property.valueConstraint.valueTemplateRefs : []
```

Would that work?

mjgiarlo: Or, if you add the safe navigation/"Elvis"/optional chaining feature via babel, you can simplify this further to:

```javascript
const templateRefList = property?.valueConstraint?.valueTemplateRefs || []
```

jermnelson: Thanks! You're right that using `map` over-complicates the `templateRefList` variable setting; I need to read-up on the babel chaining feature you mention but I went ahead and pushed up your first suggestion.

PR #523, opened 2019-05-14
Missing templates

### LINKED ISSUE #496: PropertyTemplateOutline, ResourceTemplateForm False Positive Tests Clean-up

### LINKED ISSUE #517: fix message given when trying to open RT that points to an RT that Trellis doesn't know about
When you try to us (open for cataloging) a RT that points to another RT (property template with type = resource), and Trellis doesn't know about the referenced RT, you get a message like:

"There are missing resource templates required by resource template: http://id.loc.gov/ontologies/bibframe/Work.
Make sure all referenced templates in property template are uploaded first"

It would be very helpful to user if it could tell you the *name of the referenced RT that is missing* (instead of telling you the class of the resource template that is pointing to the missing ones).

so instead of:
"There are missing resource templates required by resource template: http://id.loc.gov/ontologies/bibframe/Work..."

the message would be
"This Resource Template requires the following resource templates, which are not yet in Sinopia: [then list the IDs of the missing Resource Templates]"

jgreben:
Fixes #517 
Fixes #496 for ResourceTemplateFormTest

- Collect ids from resource templates that could not be fetched from server and display them in error message:

Screen Shot 2019-05-14 at 4 19 47 PM

- Fix false positives in ResourceTemplateForm tests and test for missing template message

- There was a function that was needed to be used in three different components to get the resource name out of a URI (everything after the last `/`). Switch to using a one-liner instead of repeating that function (or creating a global function)

### DIFF
--- a/__tests__/components/editor/ResourceTemplateForm.test.js
+++ b/__tests__/components/editor/ResourceTemplateForm.test.js
@@ -8,15 +8,14 @@ import { ResourceTemplateForm } from '../../../src/components/editor/ResourceTem
 
 describe('<ResourceTemplateForm /> functional testing', () => {
 
- const basicRt = { resourceURI: 'http://schema.org/name' }
- const basicWrapper = shallow(<ResourceTemplateForm propertyTemplates={[]}
-    rtId={'resource:schema:Name'}
-    resourceTemplate={ basicRt } />)
+  const basicRt = { resourceURI: 'http://schema.org/name' }
+  const basicWrapper = shallow(<ResourceTemplateForm propertyTemplates={[]}
+                                                     rtId={'resource:schema:Name'}
+                                                     resourceTemplate={ basicRt } />)
 
- const testId = jest.spyOn(shortid, 'generate').mockReturnValue('abcd45')
+  shortid.generate = jest.fn().mockReturnValue('abcd45')
 
- describe('resourceTemplateFields expectations and outputs', () => {
-
+  describe('resourceTemplateFields expectations and outputs', () => {
     it('empty array, null, or undefined resource templates', () => {
       expect(basicWrapper.instance().resourceTemplateFields([])).toEqual([])
       expect(basicWrapper.instance().resourceTemplateFields(null)).toEqual([])
@@ -53,62 +52,8 @@ describe('<ResourceTemplateForm /> functional testing', () => {
         "abcd45",
         "resourceTemplate:bf2:Note"])
     })
-
   })
 })
-
-const rtProps = {
-  "propertyTemplates": [
-    {
-      "propertyLabel": "Literally",
-      "type": "literal"
-    },
-    {
-      "propertyLabel": "Look up, look down",
-      "type": "lookup",
-      "editable": "do not override me!",
-      "repeatable": "do not override me!",
-      "mandatory": "do not override me!",
-      "valueConstraint": {
-        "useValuesFrom": [
-          "urn:ld4p:qa:names:person"
-        ]
-      }
-    },
-    {
-      "propertyLabel": "What's the frequency Kenneth?",
-      "type": "resource",
-      "valueConstraint": {
-        "useValuesFrom": [
-          "https://id.loc.gov/vocabulary/frequencies"
-        ]
-      }
-    },
-    {
-      "propertyLabel": "Chain chain chains",
-      "type": "resource",
-      "valueConstraint": {
-        "valueTemplateRefs": [
-          "resourceTemplate:bf2:Note"
-        ]
-      },
-      "mandatory": "true"
-    },
-    {
-      "propertyLabel": "YAM (yet another modal)",
-      "type": "resource",
-      "valueConstraint": {
-        "valueTemplateRefs": [
-          "resourceTemplate:bf2:Note"
-        ]
-      }
-    },
-    {
-      "propertyLabel": "Non-modal resource",
-      "type": "resource"
-    }
-  ]
-}
 
 const mockResponse = (status, statusText, response) => {
   return new Response(response, {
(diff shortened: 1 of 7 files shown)


src/components/editor/ResourceTemplateForm.jsx
```diff
@@ -222,7 +226,9 @@ export class ResourceTemplateForm extends Component {
     const errMessage = <div className="alert alert-warning">
       There are missing resource templates required by resource template: <strong>{this.props.resourceTemplate.resourceURI}</strong>.
       <br />
-      Make sure all referenced templates in property template are uploaded first
+      Make sure all referenced templates in property template are uploaded first. Missing templates:
```
ndushay: should we say Please?

src/components/editor/ResourceTemplateForm.jsx
```diff
@@ -85,7 +83,13 @@ export class ResourceTemplateForm extends Component {
 
   resourceTemplatePromises = async (templateRefs) => {
     return Promise.all(templateRefs.map(rtId =>
-      getResourceTemplate(rtId)
+      getResourceTemplate(rtId).catch(err => {
+        const errId = decodeURIComponent(/[^/]*$/.exec(err.url)[0])
```
ndushay: while I laud the one-liner, I find this obtuse at best and would not say it passes the "inebriated coder" test.

__tests__/components/editor/ResourceTemplateForm.test.js
```diff
@@ -13,7 +13,7 @@ describe('<ResourceTemplateForm /> functional testing', () => {
     rtId={'resource:schema:Name'}
     resourceTemplate={ basicRt } />)
 
- const testId = jest.spyOn(shortid, 'generate').mockReturnValue('abcd45')
+ jest.spyOn(shortid, 'generate').mockReturnValue('abcd45')
```
mjgiarlo: If I understand correctly, you are using `jest.spyOn()` here not as a spy but as a stub. Given that, I wonder whether you would consider using `jest.mock()` instead.

mjgiarlo: If I understand correctly, you are using jest.spyOn() here not as a spy but as a stub. Given that, I wonder whether you would consider using jest.mock() instead.

src/components/editor/ResourceTemplateForm.jsx
```diff
@@ -85,7 +83,13 @@ export class ResourceTemplateForm extends Component {
...
+        this.setState({tempErrorUrl: errId})
```
mjgiarlo: I have seen this pattern used elsewhere in the editor and I'm a bit puzzled. As far as I can tell, the only code that looks for and uses `tempErrorUrl` in state is two lines beneath. Which makes me wonder: what do we get out of injecting this into application state? In other words, how is this code different from:

```javascript
const joinedErrorUrls = this.state.templateErrors.slice(0)
joinedErrorUrls.push(errId)
this.setState({templateErrors: joinedErrorUrls})
```

:question:

src/components/editor/SinopiaResourceTemplates.jsx
```diff
@@ -30,7 +30,7 @@ class SinopiaResourceTemplates extends Component {
 
       await this.fulfillGroupPromise(groupPromise).then(async () => {
         this.state.groupData.map(group => {
-          const groupName = this.resourceToName(group)
+          const groupName = /[^/]*$/.exec(group)[0]
```
mjgiarlo: I'm OK with moving from a string parsing strategy to a regex matching one, but I wonder whether we should combine strategies here and put the regex strategy **inside** the implementation of `resourceToName` so we don't have an identical regex duplicated several times throughout the code. That will make it less tedious to change later and less error-prone.

src/components/editor/SinopiaResourceTemplates.jsx
```diff
@@ -73,14 +73,9 @@ class SinopiaResourceTemplates extends Component {
   }
 
   groupDataPromise = (groupName, name) => new Promise(async (resolve) => {
-    // await resolve(instance.getResourceWithHttpInfo(groupName, name, { acceptEncoding: 'application/json' }))
```
mjgiarlo: Yay for cruft removal!

src/components/editor/SinopiaResourceTemplates.jsx
```diff
@@ -73,14 +73,9 @@ class SinopiaResourceTemplates extends Component {
...
     await resolve(getResourceTemplate(name, groupName))
   })
 
-  resourceToName = (resource) => {
```
mjgiarlo: Not a fan of removing this method per the above comment. Also don't love that we have this same method defined in multiple components. Combined with my above related comment, what this says to me is that we ought to have `resourceToName` implemented elsewhere (in some utilities class?) and that both the `SinopiaResourceTemplates` and `UpdateResourceModal` components should import that function. **And** I'm fully on board with replacing the implementation so that it uses a regex instead of substrings.

mjgiarlo:
@ndushay :speech_balloon: 
> 
> Also, the tests are a no-go; I believe we still have false positives in the affected test file.

This is a concern of mine as well. Can you confirm that you can tweak the expected values in these tests to make them fail?

jgreben:
The issue has changed to loading a profile with missing RTs instead of messaging. Closing PR for now, but please leave branch up for the time being.

jgreben:
After discussing with Michelle, this is the right issue and solution.

__tests__/components/editor/ResourceTemplateForm.test.js
jgreben: I couldn't find a way to make it work with just `jest.mock()` but the unused const testId was bothering me. I found a different solution which is:
```
const shortid = require('shortid')
 shortid.generate = jest.fn().mockReturnValue('abcd45')
``` 
Do you think this is better?

src/components/editor/ResourceTemplateForm.jsx
jgreben: I suppose...

src/components/editor/ResourceTemplateForm.jsx
jgreben: So given @mjgiarlo comments below I'm not sure if I should use regex in an imported "utilities" function. I ended up not doing that and used substring instead.

src/components/editor/SinopiaResourceTemplates.jsx
jgreben: I ended up using substring inside an imported utilities function.

src/components/editor/SinopiaResourceTemplates.jsx
jgreben: see comments above

src/components/editor/ResourceTemplateForm.jsx
jgreben: That works fine in this case and I fixed it up. I'd have to take a second look at the other instances where I used this pattern. I think here I was blindly following the same pattern I used elsewhere and didn't need to, but I think maybe elsewhere it was used for a reason (maybe because of updating state or awaiting promises). I remember struggling with getting the array in question populated correctly and that pattern is what ended up working for me. In any case, It's definitely work taking a second look at those, and you may already be refactoring the components where those patterns were anyway.

__tests__/components/editor/ResourceTemplateForm.test.js
mjgiarlo: @jgreben Yep! That's even better. How would you feel about using `import ... from` instead of `const ... require()`? It looks like we do use both in the editor. Looks like we use import nearly 400 times and we use require 45 times, so we're inching closer to consistency here. This is minor, admittedly.

src/components/editor/SinopiaResourceTemplates.jsx
mjgiarlo: WFM! Thanks.

src/components/editor/SinopiaResourceTemplates.jsx
mjgiarlo: :man_dancing:

jgreben:
> @ndushay 💬
> 
> > Also, the tests are a no-go; I believe we still have false positives in the affected test file.
> 
> This is a concern of mine as well. @jgreben Can you confirm that you can tweak the expected values in these tests to make them fail?

I was able to confirm that by changing the various expected values in the test, I could make them fail.

__tests__/components/editor/ResourceTemplateForm.test.js
```diff
@@ -148,134 +93,144 @@ const responseBody = [{
...
-      resourceTemplate = {rtTest}
-      handleGenerateRDF = {mockHandleGenerateLD}
-      literals = {lits}
-      lookups = {lups}
-      rtId = {"resourceTemplate:bf2:Monograph:Instance"}
-      parentResourceTemplate = {"resourceTemplate:bf2:Monograph:Instance"}
-      generateLD = { ld }
-    />
-  )
+  const mockAsyncCall = (index) => {
```
ndushay: thanks for the rename - I find it much clearer.

__tests__/components/editor/ResourceTemplateForm.test.js
```diff
@@ -53,63 +52,9 @@ describe('<ResourceTemplateForm /> functional testing', () => {
         "abcd45",
         "resourceTemplate:bf2:Note"])
     })
-
   })
 })
 
-const rtProps = {
```
ndushay: i love that you moved these constants down to where they are actually used.

__tests__/components/editor/ResourceTemplateForm.test.js
```diff
@@ -148,134 +93,144 @@ const responseBody = [{
...
-    },
-    {
-      "@id": "n3-8",
-      "@type": "http://id.loc.gov/ontologies/bibframe/Note",
-      "http://www.w3.org/2000/01/rdf-schema#label": "NOTE"
-    }
-  ]
-}
-
 const rtTest = { resourceURI: "http://id.loc.gov/ontologies/bibframe/Work" }
```
ndushay: I can live w it as is, but .... "rtTest" ?   what is this?  A resource template missing most of the required data?

__tests__/components/editor/ResourceTemplateForm.test.js
```diff
@@ -148,134 +93,144 @@ const responseBody = [{
...
+                "urn:ld4p:qa:names:person"
+              ]
+            }
+          }
+        ]
+      }
+
+      const wrapper = shallow(<ResourceTemplateForm {...rtProps} resourceTemplate = {rtTest}/>)
+
+      expect.assertions(2)
```
ndushay: 👍

__tests__/components/editor/ResourceTemplateForm.test.js
```diff
@@ -305,31 +260,60 @@ describe('<ResourceTemplateForm /> after fetching data from sinopia server', ()
...
-  it.skip('renders error alert box', async () => {
-    // FIXME: this test gives false positive - see github issue #496 - perhaps we need an integration test
-    // await wrapper.instance().fulfillRTPromises(promises).then(() => wrapper.update()).then(() => {
-    //   expect(wrapper.state.errot).toBeTruthy()
-    //   const errorEl = wrapper.find('div.alert')
-    //   expect(errorEl).toHaveLength(1)
-    //   expect(errorEl.text()).toEqual('Sinopia server is offline or has no resource templates to display')
-    // }).catch(e => {})
+
+  const asyncCall = () => {
```
ndushay: whoops - another mockAsyncCall

__tests__/components/editor/ResourceTemplateForm.test.js
```diff
@@ -305,31 +260,60 @@ describe('<ResourceTemplateForm /> after fetching data from sinopia server', ()
...
+          ]
+        }
+      }
+    ]
+  }
+
+  const wrapper = shallow(<ResourceTemplateForm {...rtProps} resourceTemplate = {rtTest}/>)
+
+  it('renders error alert box', async () => {
+    expect.assertions(3)
```
ndushay: 🕺

src/components/editor/ResourceTemplateForm.jsx
```diff
@@ -12,14 +12,12 @@ import PropertyResourceTemplate from './PropertyResourceTemplate'
 import lookupConfig from '../../../static/spoofedFilesFromServer/fromSinopiaServer/lookupConfig.json'
 import { getLD, setItems, removeAllContent } from '../../actions/index'
 import { getResourceTemplate } from '../../sinopiaServer'
+import {resourceToName} from "../../Utilities";
```
ndushay: totally ok to ignore ... all other curly braces nearby have a space.  :-P

src/components/editor/ResourceTemplateForm.jsx
```diff
@@ -85,7 +84,11 @@ export class ResourceTemplateForm extends Component {
 
   resourceTemplatePromises = async (templateRefs) => {
```
ndushay: so now this is returning a single promise, yes?   Would `resourceTemplatesPromise` be more accurate?

src/components/editor/ResourceTemplateForm.jsx
```diff
@@ -222,7 +225,9 @@ export class ResourceTemplateForm extends Component {
     const errMessage = <div className="alert alert-warning">
       There are missing resource templates required by resource template: <strong>{this.props.resourceTemplate.resourceURI}</strong>.
       <br />
-      Make sure all referenced templates in property template are uploaded first
+      Please make sure all referenced templates in property template are uploaded first. Missing templates:
```
ndushay: 🕺

__tests__/components/editor/ResourceTemplateForm.test.js
jgreben: Yes, all it requires for the test is the resourceURI.

__tests__/components/editor/ResourceTemplateForm.test.js
jgreben: Fixed.

src/components/editor/ResourceTemplateForm.jsx
jgreben: I don't think so...It's doing Promise.all and iterating over the collected resource templates that it found in the profile.

src/components/editor/ResourceTemplateForm.jsx
jgreben: I guess people like spaces in their curly braces... :)

src/components/editor/ResourceTemplateForm.jsx
ndushay: yes, but it returns a single promise (I had to look that up) for all of the templates combined.  Merging, but if you agree, feel free to make a tiny PR.

jgreben: ok that makes sense.

PR #525, opened 2019-05-15
Refactor and fix resource template components

### LINKED ISSUE #506: Running on the AWS server, uploading an existing RT does not actually update RT
what should happen:

- user should see a warning that they will overwrite existing template
- when user clicks thru, the uploaded template should be shown in the template list, and clicking on it should have the updates reflected in the editor

Josh believes this may have to do with how we call "update" and "componentDidMount" and such.

~~Related to #510~~ (#510 was a duplicate and is now closed)
Blocked by #507

### LINKED ISSUE #507: (Safari, Firefox) Browser not refreshing list of Resource Templates
Safari:  When uploading a Profile containing one or more Resource Templates, the expected success message is being returned but the list of available Resource Templates is not being update in the display.

Firefox:  When uploading a profile, the success message is not returned the first time;  you have to do it twice.

~~Blocks #506~~

mjgiarlo:
Fixes #506 
Fixes #507 

Includes:
* Bump sinopia_server to 3.0.0-beta4
  * Using the newer version of the sinopia client allows the editor to invoke the updateResource API with a custom content type (application/json). Prior to this change, updates worked (apparently) but what was happening was the editor was issuing PUT requests and getting 204s... but not modifying the NonRDFSource. Instead the RDFSource that Trellis auto-attaches to every NonRDFSource was being modified.
* Use new SSL-based location of the CreativeCommons image
  * This gets rid of the constant mixed content errors in the console.
* Correct two typos that are surfaced in the user-facing UI
* Avoid wrapping promises in other promises (coming from the sinopia client)
* Refactor resource template components
* Add test coverage for refactoring of SinopiaResourceTemplates component
* Add test coverage for refactoring of ImportResourceTemplate component
* Include some fixes to Import component and some fakes to make it easier to test the sinopia client implementation

### DIFF
--- a/__mocks__/SinopiaClientErrorFake.js
+++ b/__mocks__/SinopiaClientErrorFake.js
@@ -1,0 +1,13 @@
+export default class SinopiaClientErrorFake {
+  async createResourceWithHttpInfo() {
+    throw {
+      response: 'i am an error for a created object'
+    }
+  }
+
+  async updateResourceWithHttpInfo() {
+    throw {
+      response: 'i am an error for an updated object'
+    }
+  }
+}
--- a/__tests__/components/editor/SinopiaResourceTemplates.test.js
+++ b/__tests__/components/editor/SinopiaResourceTemplates.test.js
@@ -2,8 +2,10 @@ import 'jsdom-global/register'
 import React from 'react'
 import { shallow } from 'enzyme'
 import SinopiaResourceTemplates from '../../../src/components/editor/SinopiaResourceTemplates'
-import BootstrapTable from 'react-bootstrap-table'
+import { getGroups, listResourcesInGroupContainer, getResourceTemplate } from '../../../src/sinopiaServer'
 import 'isomorphic-fetch'
+
+jest.mock('../../../src/sinopiaServer')
 
 describe('<SinopiaResourceTemplates />', () => {
   const message = [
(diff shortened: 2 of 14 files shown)


src/components/editor/ImportResourceTemplate.jsx
```diff
@@ -43,80 +44,102 @@ class ImportResourceTemplate extends Component {
...
+  handleResponse = (response, operation, profileCount) => {
+    console.log(`handling ${operation} response:`)
+    console.dir(response)
+    // HTTP status 409 == Conflict
+    if (operation == 'update' && response.status !== 409) {
+      const msg = 'The sinopia server is not accepting the request for this resource.'
+      this.state.updateResourceError ? this.setState({message: [`${msg}: ${this.state.updateResourceError}`]}) : this.setState({message: [msg]})
+    }
+
+    // HTTP status 401 == Unauthorized
```
ndushay: mike sez this shouldn't be here - we should surface the error

src/components/editor/ImportResourceTemplate.jsx
```diff
@@ -43,80 +44,102 @@ class ImportResourceTemplate extends Component {
...
+    console.log('received non-error response to update resource request:')
+    console.dir(response)
+    return response.response
   }
 
-  updateAllResources = (rts, group) => {
-    return Promise.all(rts.map(async rt =>
-      await this.fulfillCreateResourcePromise(this.updateResourcePromise(rt, group))
-    ))
+  handleResponse = (response, operation, profileCount) => {
```
ndushay: consider renaming to handleTrellisResponse?  or is that too long?

src/components/editor/ImportResourceTemplate.jsx
```diff
@@ -43,80 +44,102 @@ class ImportResourceTemplate extends Component {
...
+    }
   }
 
   handleUpdateResource = async (rts, group) => {
-    await this.updateAllResources(rts, group).then(() => {
-      const incrementedKey = this.state.updateKey + 1
-      this.setState({updateKey: incrementedKey})
-      this.modalClose()
-      window.location.reload()
+    rts.forEach(async rt => {
```
ndushay: would there be any gain to using Promise.all to wait for all to return here? (prob not)

src/components/editor/ImportResourceTemplate.jsx
ndushay: consider not using "handle" because we use that for things like "handleClick" ... maybe `processTrellisResponse` ??

"It sets the state for the trellis response" is a description

`updateStatePerTrellisResponse`   `setStatePerTrellis`

src/components/editor/SinopiaResourceTemplates.jsx
```diff
@@ -23,63 +23,69 @@ class SinopiaResourceTemplates extends Component {
...
-  fulfillGroupPromise = (promise) => {
-    promise.then(data => {
-      const contains = [].concat(data.response.body.contains)
-      this.setState({groupData: contains})
-    }).catch((error) => {
-      this.setState({message: error})
-    })
-    return promise
-  }
+  handleGroupResponse = data => {
```
ndushay: consider renaming to `setStatePerTrellis` or whatever. ("handle" is often "handleClick" sort of thing)

src/components/editor/SinopiaResourceTemplates.jsx
```diff
@@ -23,63 +23,69 @@ class SinopiaResourceTemplates extends Component {
...
+    if (data.response.body === null)
+      return
 
-  fulfillGroupData = (data) => {
     const groupName = this.resourceToName(data.response.body['@id'])
+    // Array-ify the value of `contains` and filter out falsy values like null and undefined
+    const contains = [].concat(data.response.body.contains).filter(Boolean)
+
+    contains.forEach(child => {
```
ndushay: comment from discussion:   `forEach` for iterating thru without needing the result after;  `map` when you need the result what happens for each step.

There may be some foo around `forEach` vs `map`

ndushay: caution:  don't use "return" with forEach or you'll be surprised!

src/components/editor/SinopiaResourceTemplates.jsx
```diff
@@ -23,63 +23,69 @@ class SinopiaResourceTemplates extends Component {
...
+        this.setState({
+          tempState: {
+            key: rt.id,
+            name: rt.resourceLabel,
+            uri: child,
+            id: rt.id,
+            group: groupName
+          }
+        })
+        const joined = this.state.templatesForGroup.slice(0)
```
ndushay: Jeremy: a bit uncomfy with `slice`.   Suggests `assigned` -- semantically clearer.  Also, with slice, a blank or empty object, it will error.

src/components/editor/ImportResourceTemplate.jsx
mjgiarlo: done in my local branch

src/components/editor/ImportResourceTemplate.jsx
mjgiarlo: Done in my local branch.

src/components/editor/ImportResourceTemplate.jsx
mjgiarlo: I'm trying to avoid overt promises wherever possible. Prolly worth broader team discussion.

src/components/editor/SinopiaResourceTemplates.jsx
mjgiarlo: done in my local branch

src/components/editor/SinopiaResourceTemplates.jsx
mjgiarlo: `return` within `.forEach()` functions like `next` in Ruby or `continue` in Java—it merely terminates the current iteration. It does not return a value to the containing context.

src/components/editor/SinopiaResourceTemplates.jsx
mjgiarlo: `Object.assign()`. Though the spread operator (`...`) is nice too.

mjgiarlo:
Will rebase and deal with merge conflicts in the morning.

mjgiarlo:
This needs more user testing (which I am doing, both locally and in dev), and I need to remove the Config changes.

src/components/editor/ImportResourceTemplate.jsx
```diff
@@ -1,120 +1,123 @@
...
     this.setState( { modalShow: false } )
   }
 
-  //resource templates are set via ImportFileZone and passed to ResourceTemplate via redirect to Editor
+  // Resource templates are set via ImportFileZone and passed to ResourceTemplate via redirect to Editor
   setResourceTemplates = (content, group) => {
     const profileCount = content.Profile.resourceTemplates.length
-    content.Profile.resourceTemplates.map(rt => {
-      this.fulfillCreateResourcePromise(this.createResourcePromise(rt, group, profileCount))
+    content.Profile.resourceTemplates.forEach(async rt => {
```
jermnelson: Since you are using the arrow function in this `forEach` as an anonymous function you could just as easily use the `map` operator.

src/components/editor/ImportResourceTemplate.jsx
```diff
@@ -1,120 +1,123 @@
...
 
-  updateAllResources = (rts, group) => {
-    return Promise.all(rts.map(async rt =>
-      await this.fulfillCreateResourcePromise(this.updateResourcePromise(rt, group))
-    ))
+  updateStateFromServerResponse = (response, operation, profileCount) => {
+    // HTTP status 409 == Conflict
+    if (operation == 'update' && response.status !== 409) {
+      const msg = 'The sinopia server is not accepting the request for this resource.'
+      this.state.updateResourceError ?
```
jermnelson: 👍  on this multiple-line conditional, I think it reads easier than having this on a single line.

src/components/editor/ImportResourceTemplate.jsx
jermnelson: Since you are using the anonymous arrow function in this `forEach`, you could use the `map` operator here as well.

src/components/editor/SinopiaResourceTemplates.jsx
```diff
@@ -1,95 +1,92 @@
...
-    }
-
-  }
+      const templates = [...this.state.resourceTemplates]
 
-  groupDataPromise = (groupName, name) => new Promise(async (resolve) => {
-    await resolve(getResourceTemplate(name, groupName))
-  })
+      // Check if newly retrieved template is already stored in component state. If not, add it.
+      if (!templates.some(template => template.key === retrievedTemplateObject.key))
```
jermnelson: 👍 Had to lookup `some` array method in the [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some), like it's use here.

src/components/editor/SinopiaResourceTemplates.jsx
```diff
@@ -1,95 +1,92 @@
...
 
-  render(){
-    if (this.state.message) {
+  render() {
+    if (this.state.errors.length > 0) {
       return (
         <div className="alert alert-warning alert-dismissible">
           <button className="close" data-dismiss="alert" aria-label="close">&times;</button>
-          No connection to the Sinopia Server is available, or there are no resources for any group
+          No connection to the Sinopia Server is available, or there are no resources for any group.
```
jermnelson: Not a blocking comment but it seems this should be two separate error messages; 
1.  one for when there is no connection to the Sinopia Server
1.  another message if there are no resources for any group.

__mocks__/SinopiaClientErrorFake.js
```diff
@@ -0,0 +1,13 @@
+export default class SinopiaClientErrorFake {
```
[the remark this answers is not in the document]
mjgiarlo: `ImportResourceTemplate` calls these methods on an object called `instance`, so at the very least, it should be an object. Doesn't Babel convert these classes to functions underneath the hood ultimately? FWIW, this is the same fake pattern we've been using in sinopia_acl, sinopia_indexing_pipeline, and sinopia_user_backup.

__tests__/components/editor/SinopiaResourceTemplates.test.js
```diff
@@ -20,73 +22,216 @@ describe('<SinopiaResourceTemplates />', () => {
...
+    it('initializes this.resourceTemplates to empty array', () => {
+      expect(wrapper.state().resourceTemplates).toEqual([])
+    })
+  })
+
+  describe('componentDidUpdate()', () => {
+    const initialUpdateKey = 1
+
+    it('calls fetchResourceTemplatesFromGroups() if updateKey has been incremented', async () => {
+      const wrapper2 = shallow(<SinopiaResourceTemplates message={message} updateKey={initialUpdateKey} />)
```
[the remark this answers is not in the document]
mjgiarlo: OK, good to know.

src/components/editor/SinopiaResourceTemplates.jsx
mjgiarlo: Alas, I wound up replacing this to fix one of the bugs! Ha.

src/components/editor/SinopiaResourceTemplates.jsx
mjgiarlo: I totally agree! would love to see this ticketed and taken on. (for now, inclined not to address this in this already sizable branch since it's not connected to the issues I'm working on.)

jermnelson: Done, now being tracked in issue [352](https://github.com/LD4P/sinopia_editor/issues/532)

__mocks__/SinopiaClientErrorFake.js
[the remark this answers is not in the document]
jermnelson: Yeah, recognize that Babel does convert these classes into functions anyways, I think it easier to see side effects in pure functions vs. in classes.

Issue #442, opened 2019-04-01
InputLookupSinopia React Component


jermnelson:
Create an `InputLookupSinopia` component that uses the Sinopia RDF Elasticsearch for searching and linking to entities in Sinopia. Should use the same Swagger spec used for QA.

PR #546, opened 2019-05-22
Refactor ResourceTemplateForm and PropertyTemplateOutline

### LINKED ISSUE #527: Refactor components and tests for ResourceTemplateForm and PropertyTemplateOutline
While looking at test for PropertyTemplateOutline I came across this case:

```
else {
      // TODO: remove this case
      // if no valueTemplateRef, assume it is a confused propertyTemplate intended to be of type lookup
      // NOTE: this case is due to orig LC resourceTemplates;  sinopia schemas v0.0.9 and up do NOT allow this
      return <InputListLOC propertyTemplate = {property}
                            lookupConfig = {getLookupConfigItem(property)}
                            rtId = {property.rtId} />
    }
```

Removing this case breaks the test and after reviewing the code around this it seems that we need some refactoring to:

- handle the InputListLOC case in the PropertyTemplateOutline properly
- try to create a class to handle the choosing of the component type (lookup, literal, etc) that will work for both the PropertyTemplateOutline and the ResourceTemplateForm

Hopefully this will be able to fix the breaking test and make the code more readable and testable.

jgreben:
Fixes #527 

- ResourceTemplateForm and PropertyTemplateOutline now use a single class component, PropertyComponent, that decides what kind of input form to mount
- Moves some helper functions to Utilities file
- Creates a new class component to render a ResourceProperty used by the PropertyTemplateOutline
- PropertyTemplateOutline separates the concerns of loading the nested resource templates and mounting the added property rows. Fetching of RTs happens the onClick function is called instead.
- OutlineHeader gets a new data-id attribute to make integration tests of nested resource templates possible
- All tests are refactored according to the new component structure and their separation of concerns. 
- Used the "green-red-green" method and expect.assertions on async tests when writing all these tests to verify that there were no false positives


src/Utilities.js
```diff
@@ -1,5 +1,18 @@
 // Copyright 2018, 2019 Stanford University see Apache2.txt for license
 
+export const isResourceWithValueTemplateRef = ( property ) => {
```
jermnelson: Like moving this function to this utilities module

src/components/editor/PropertyComponent.jsx
```diff
@@ -0,0 +1,95 @@
...
+    let config, result
+
+    try {
+      config = this.state.configuration[0].component
+    } catch {
+      // ignore undefined configuration
+    }
+
+    // We do not support mixed list and lookups, so we will just go with the value of the first config item found
+    switch(config) {
```
jermnelson: Not blocking but the more get into the specific Sinopia workflows, I'm wondering if the `type` property could reference the React component instead of us needing to this conditional code to determine the specific React component to use.

src/components/editor/PropertyComponent.jsx
```diff
@@ -0,0 +1,95 @@
...
+        }
+    }
+
+    return result
+  }
+
+  render() {
+    if (this.inputComponentType) {
+      return(
+        <div>
```
jermnelson: Do we need this `div`? Couldn't this be a React fragment?

src/components/editor/PropertyComponent.jsx
```diff
@@ -0,0 +1,95 @@
...
+      return false
+    }
+  }
+}
+
+export const getLookupConfigItems = (property) => {
+  const templateConfigItems = []
+
+  if (_.find([property], 'valueConstraint.useValuesFrom')) {
+    property.valueConstraint.useValuesFrom.forEach(templateUri => {
```
jermnelson: Totally rehashing ongoing discussion but I think this `forEach` should be `map`. This isn't a blocking comment so please resolve/ignore if you want to keep this operator.

src/components/editor/PropertyComponent.jsx
```diff
@@ -0,0 +1,95 @@
...
+      const configItem = getLookupConfigForTemplateUri(templateUri)
+      templateConfigItems.push(configItem)
+    })
+  }
+  return templateConfigItems
+}
+
+export const getLookupConfigForTemplateUri = (templateUri) => {
+  let returnConfigItem = {}
+  lookupConfig.forEach(configItem => {
```
jermnelson: See `forEach` vs `map` comment from above. One reason to favor using `map` is that it is faster than `forEach`.

src/components/editor/PropertyTemplateOutline.jsx
```diff
@@ -97,162 +42,98 @@ export class PropertyTemplateOutline extends Component {
...
-  }
 
-  /**
-   * @private
-   */
-  outlinerClasses = () => {
-    let classNames = "rOutline-property"
-    if (this.state.collapsed) { classNames += " collapse"}
-    return classNames
+    newOutput.forEach((propertyJsx) => {
```
jermnelson: Same `forEach` vs. `map` comment.

src/components/editor/PropertyComponent.jsx
jgreben: I think we decided that `list` would not be one of the type properties (See https://github.com/LD4P/sinopia/blob/0d0d858343725ba11e4ceec59a6c917e237c37c0/schemas/0.2.0/property-template.json#L71) so we need to check the config details of the lookup source to choose the component. However, this is going to change slightly after QA starts hosting the config (see https://github.com/LD4P/qa_server/issues/51) and will hinge ion whether or not the `action` is a `search`. 

If I am mis-reading your comment let's discuss!

jermnelson: Resolving for now, this is a larger conversation as we look towards **M4** support for a [InputLookupSinopia](https://github.com/LD4P/sinopia_editor/issues/442) component to support searching the Sinopia Server's elasticsearch index. Don't think we want to depend on QA to host configurations for non-QA React components (although they are for the `InputListLOC`?) 

There are likely other components coming soon like a [RDFRegistry](https://www.rdaregistry.info/) that QA is not supporting but has already been requested by a cohort member. Another component is a direct Wikidata lookup component that uses the wikidata REST JSON API (I don't think we want to wait if/when QA supports Wikidata when there is already interest in the cohort to use Wikidata directly in Sinopia and a REST JSON api already exists)

jgreben: Point well taken, and we may need to revisit the schema requirements and spefication of different types in property templates

jgreben:
LOL:
```
forEach(instanceofForEach => {
  useMapInstead()
})
```

Maybe we can have a discussion (today?) and come to an agreed resolution?

__tests__/components/editor/PropertyTemplateOutline.test.js
```diff
@@ -280,132 +27,119 @@ describe('<PropertyTemplateOutline /> with propertyTemplate of type resource', (
...
-    // - it needs to be expanded first?  implying an integration test
-    // FIXME: the approach below is a failed attempted to 'inject' info to an inner method
-    // await wrapper.instance().fulfillRTPromises(promises).then(() => wrapper.update()).then(() => {
-    //   const addButton = wrapper.find('div > section > PropertyActionButtons > div > AddButton')
-    //   addButton.handleClick = mockHandleAddClick
-    //   addButton.simulate('click')
-    //   expect(mockHandleAddClick.mock.calls.length).toBe(1) // correct
-    //   expect(mockHandleAddClick.mock.calls.length).toBe(0) // incorrect
-    // }).catch(() => {})
+  it('outline header anchor has an ID based on the property template prpoertyURI', () => {
```
ndushay: typo

__tests__/integration/nestedResources.test.js
```diff
@@ -0,0 +1,41 @@
+// Copyright 2019 Stanford University see Apache2.txt for license
+
+import Config from '../../src/Config'
+import pupExpect from 'expect-puppeteer'
```
ndushay: huzzah!  an integration test!  Yay!

__tests__/Utilities.test.js
```diff
@@ -0,0 +1,46 @@
+// Copyright 2019 Stanford University see Apache2.txt for license
+
+import {isResourceWithValueTemplateRef, resourceToName } from '../src/Utilities'
+
+describe('Utilities', () => {
+  it('can check if a property template is a type of resource with valueTemplateRefs', () => {
```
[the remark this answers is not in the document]
jgreben: I'll submit a follow-on PR.

PR #549, opened 2019-05-23
Update editor to use new default schema 0.0.3

### LINKED ISSUE #548: Handle schema 0.2.0, 0.1.0, 0.0.9 and 0.0.3 profiles/resource templates.
From Naomi: "we want the editor to be able to cope with 0.2.0, 0.1.0, 0.0.9 and 0.0.3 profiles/resource templates. 0.2.0 and 0.1.0 and 0.0.9 resource templates will have the schema url inside them; 0.0.3 templates do not. (Was there a pre-loading of schemas into the ajv object?)"

Investigate this further and make necessary changes.

jgreben:
Fixes #548 
Schema 0.0.3 is the new default schema that requires spaceless resource IDs.

#548 states "we want the editor to be able to cope with 0.2.0, 0.1.0, 0.0.9 and 0.0.3 profiles/resource templates..." but the only place in the editor that specifies a schema is the Config that sets the default schema if there is not one specified in the template. Otherwise it will use whatever schema is in the template. 

See https://github.com/LD4P/sinopia_editor/blob/e5582807716eb38fdb7e90012ee5465f6a05326a/src/components/editor/ImportFileZone.jsx#L70


jgreben:
> Please rename the test files to "_no_schema_specified" or "_no_schema" or "_without_schema" (from, e.g. `__fixtures__/lcc_v0.0.2.json`)
> 
> Also, please add _profile_editor_ ticket for adding tests to _profile_editor_ showing it copes with 0.2.0 and 0.0.3 schemas, and any other appropriate changes.

Ok. I am also changing the `0.1.0` tests and filenames to `0.2.0`.

ndushay:
I'll go ahead and make the profile_editor ticket.  :-)

PR #550, opened 2019-05-23
Prevent upload modal & flash message from showing stale messages

### LINKED ISSUE #534: strangeness with warning and confirmation messages on RT upload
tested in DEV with Firefox with Profiles containing only one RT each

1) attempt to upload existing template friday5, get warning as expected "friday5 already exists. Do you want to overwrite these resource templates?", choose "Yes, overwrite", get blue message as expected "Attempting to update, Updated". Everything works so far.
2) try to upload another RT which is called friday5, but which has some internal changes (resource and property label). get message "friday5 already exists, friday5 already exists" as if the message from #1 didn't get cleared. choose "yes overwrite", blue message is "Attempting to update , Updated , Attempting to update , Unexpected response (500)! , Updated" again as if the message from #1 is still there (and don't know what the 500 error is but the revised RT was successfully uploaded and the old one overwritten). (I checked that the Resource label was updated as per the new version, and opened the RT to make sure the new property label was showing and both those things worked fine.)
3) now I close the blue message (with the little x) to see if the next time I do it, I will get a fresh confirmation message. but now the blue message doesn't appear at al when I try to upload a new or revised resource template (though the new or revised does seem to upload successfully). If I click away from that tab and then back, the blue message comes back the next time I upload, and the blue message is "refreshed", i.e., it doesn't have any of the old confirmation messages.

basically if I keep trying to upload existing Resource Templates, the overwrite warning keeps getting longer and longer, like this (this is on the 3rd time uploading a saturday2 template)
Screen Shot 2019-05-18 at 8 56 33 AM

and the blue confirm message keeps getting longer and longer, like:
Screen Shot 2019-05-18 at 8 56 43 AM

mjgiarlo:
Fixes #534

This branch includes the following changes:
* Reset messages and errors in `ImportResourceTemplate` component when HTTP response is successful
* Reset messages when update handler is finished
* Prevent users from dismissing flash message box, else it never comes back
* Extract code to parse a response object into a human-friendly location into a new function in `ImportResourceTemplate` component
* Add Babel plugin for optional chaining (as dev dependency) via the safe navigation (or "Elvis") operator
  * This lets us safely reach into objects such as HTTP responses without having to check at every level for `null` or `undefined`

### DIFF
--- a/__tests__/components/editor/ImportResourceTemplate.test.js
+++ b/__tests__/components/editor/ImportResourceTemplate.test.js
@@ -124,10 +124,11 @@ describe('<ImportResourceTemplate />', () => {
 
       expect(wrapper.state().message).toEqual([])
 
-      wrapper.instance().updateStateFromServerResponse({ status: 200, headers: {} })
+      wrapper.instance().updateStateFromServerResponse({ status: 200, headers: {} }, 0)
 
       expect(wrapper.state().message).toEqual(['Unexpected response (200)! '])
     })
+
     it('sets modalShow to true when receiving HTTP 409 and errors >= profileCount', () => {
       // Set new wrapper in each of these tests because we are changing state
       wrapper = shallow(<ImportResourceTemplate.WrappedComponent authenticationState={authenticationState}/>)
@@ -135,17 +136,93 @@ describe('<ImportResourceTemplate />', () => {
       expect(wrapper.state().modalShow).toBe(false)
 
       wrapper.instance().updateStateFromServerResponse({ status: 409 , headers: {} }, 0)
+
       expect(wrapper.state().modalShow).toBe(true)
     })
+
     it('sets message in state with any create operation not resulting in HTTP 409', () => {
       // Set new wrapper in each of these tests because we are changing state
       wrapper = shallow(<ImportResourceTemplate.WrappedComponent authenticationState={authenticationState}/>)
 
       expect(wrapper.state().message).toEqual([])
 
-      wrapper.instance().updateStateFromServerResponse({ status: 201 , headers: { location: 'http://foo.bar' } })
+      wrapper.instance().updateStateFromServerResponse({ status: 201 , headers: { location: 'http://sinopia.io/repository/ld4p/myResourceTemplate' } }, 0)
 
-      expect(wrapper.state().message).toEqual(['Created http://foo.bar'])
+      expect(wrapper.state().message).toEqual(['Created http://sinopia.io/repository/ld4p/myResourceTemplate'])
+    })
+
+    it('resets createResourceError and message in component state when response is OK', () => {
+      // Set new wrapper in each of these tests because we are changing state
+      wrapper = shallow(<ImportResourceTemplate.WrappedComponent authenticationState={authenticationState}/>)
+
+      wrapper.setState({ createResourceError: ['Could not update resource!'], message: ['Updated http://sinopia.io/repository/ld4p/myResourceTemplate'] })
+      expect(wrapper.state().createResourceError).toEqual(['Could not update resource!'])
+      expect(wrapper.state().message).toEqual(['Updated http://sinopia.io/repository/ld4p/myResourceTemplate'])
+
+      wrapper.instance().updateStateFromServerResponse({ ok: true, status: 204 }, 0)
+
+      expect(wrapper.state().createResourceError).toEqual([])
+      expect(wrapper.state().message).toEqual(['Updated '])
+    })
+
+    it('leaves createResourceError alone and appends message in component state when response is not OK', () => {
+      // Set new wrapper in each of these tests because we are changing state
+      wrapper = shallow(<ImportResourceTemplate.WrappedComponent authenticationState={authenticationState}/>)
+
+      wrapper.setState({ createResourceError: ['Could not update resource!'], message: ['Updated http://sinopia.io/repository/ld4p/myResourceTemplate'] })
+      expect(wrapper.state().createResourceError).toEqual(['Could not update resource!'])
+      expect(wrapper.state().message).toEqual(['Updated http://sinopia.io/repository/ld4p/myResourceTemplate'])
+
+      wrapper.instance().updateStateFromServerResponse({ ok: false, status: 400 }, 0)
+
+      expect(wrapper.state().createResourceError).toEqual(['Could not update resource!'])
+      expect(wrapper.state().message).toEqual(['Updated http://sinopia.io/repository/ld4p/myResourceTemplate', 'Unexpected response (400)! '])
+    })
+  })
+
+  describe('humanReadableLocation()', () => {
+    it('returns human-readable label when Location header exists', () => {
+      const response = {
+        headers: {
+          location: 'http://sinopia.io/repository/ld4p/myResourceTemplate'
+        },
+        req: {
+          url: 'http://sinopia.io/repository/ld4p',
+          _data: {
+            id: 'myResourceTemplate'
+          }
+        }
+      }
+      expect(wrapper.instance().humanReadableLocation(response)).toEqual('http://sinopia.io/repository/ld4p/myResourceTemplate')
+    })
+
+    it('returns human-readable label when data ID exists', () => {
+      const response = {
+        req: {
+          url: 'http://sinopia.io/repository/ld4p',
+          _data: {
+            id: 'myResourceTemplate'
+          }
+        }
+      }
+      expect(wrapper.instance().humanReadableLocation(response)).toEqual('http://sinopia.io/repository/ld4p/myResourceTemplate')
+    })
+
+    it('returns human-readable label when request URL already includes data ID', () => {
+      const response = {
+        req: {
+          url: 'http://sinopia.io/repository/ld4p/myResourceTemplate',
+          _data: {
+            id: 'myResourceTemplate'
+          }
+        }
+      }
+      expect(wrapper.instance().humanReadableLocation(response)).toEqual('http://sinopia.io/repository/ld4p/myResourceTemplate')
+    })
+
+    it('returns an empty label otherwise', () => {
+      const response = undefined
+      expect(wrapper.instance().humanReadableLocation(response)).toEqual('')
     })
   })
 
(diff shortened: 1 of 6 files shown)


__tests__/components/editor/ImportResourceTemplate.test.js
```diff
@@ -124,44 +124,125 @@ describe('<ImportResourceTemplate />', () => {
...
+
+      expect(wrapper.state().createResourceError).toEqual([])
+      expect(wrapper.state().message).toEqual(['Updated '])
+    })
+
+    it('leaves createResourceError alone and appends message in component state when response is not OK', () => {
+      // Set new wrapper in each of these tests because we are changing state
+      wrapper = shallow(<ImportResourceTemplate.WrappedComponent authenticationState={authenticationState}/>)
+
+      wrapper.setState({ createResourceError: ['foo'], message: ['bar'] })
```
jermnelson: Non-blocking comment, but I find using placeholder variable values like `foo` and `bar` to detract from what you're trying to do here. To me, example values you would expect makes it easier to tie back to what the code is actually trying to do.

package.json
```diff
@@ -22,6 +22,7 @@
     "@babel/core": "^7.4.3",
     "@babel/node": "^7.4.5",
     "@babel/plugin-proposal-class-properties": "^7.4.0",
+    "@babel/plugin-proposal-optional-chaining": "^7.2.0",
```
jermnelson: Is there any indication when the optional chaining plugin will move from being just a proposal? (Granted we're already using the class properties proposal but I'm not familiar with the actual process from moving from a *proposal* state)

src/components/editor/ImportResourceTemplate.jsx
```diff
@@ -85,20 +85,44 @@ class ImportResourceTemplate extends Component {
...
 
+    // `response.ok` is a Boolean that returns true if the HTTP status is in the
+    // 200-299 range, indicating the HTTP request was successful
+    if (response.ok) {
+      // Wipe out past messages and errors if response was successful
+      newState.createResourceError = []
+      newState.message = [newMessage]
+    } else {
+      // Response not OK, so append message to list of current messages
+      newState.message = [...this.state.message, newMessage]
```
jermnelson: Like this use of the spread operator here and the appending of the `newMessage` to the list

src/components/editor/ImportResourceTemplate.jsx
```diff
@@ -85,20 +85,44 @@ class ImportResourceTemplate extends Component {
...
+    // Use ID in request body if it exists, appending it to the request URL
+    if (response?.req?._data?.id) {
+      // But if the request URL already contains the ID, don't bother appending
+      if (response.req.url.endsWith(response.req._data.id))
+        return response.req.url
+      return `${response.req.url}/${response.req._data.id}`
+    }
+
+    // Welp, we tried anyway
+    return ''
```
jermnelson: Any specific reason you're returning an empty string? Does the calling method or function require a string?

__tests__/components/editor/ImportResourceTemplate.test.js
ndushay: you mean 'bar' could be 'I am an error message' or some such?  I'd be in favor of that.

src/components/editor/ImportResourceTemplate.jsx
mjgiarlo: The calling method uses the returned value when building a string (a label for human consumption in the UI). In general, I try to keep return values consistently typed, though it's not always feasible in practice.

__tests__/components/editor/ImportResourceTemplate.test.js
```diff
@@ -124,44 +124,125 @@ describe('<ImportResourceTemplate />', () => {
...
+      }
+      expect(wrapper.instance().humanReadableLocation(response)).toEqual('http://foo.bar/baz')
+    })
+
+    it('returns human-readable label when data ID exists', () => {
+      const response = {
+        req: {
+          url: 'http://foo.bar',
+          _data: {
+            id: 'baz'
```
ndushay: maybe "baz" could be 'data_id' ?  and url could be 'http://example.org'?  I'm getting confused as to what "label" is and what is constituted into it.  (applies to other tests using foo and bar)

src/components/editor/ImportResourceTemplate.jsx
mjgiarlo: @jermnelson Do you think this function could be improved by returning some other value (`null`? `undefined`?) and letting the calling function deal with those values?

package.json
mjgiarlo: I'm not sure. It's in stage 1, though some stage 2 to-dos are already done. It's at least mature enough that there's a babel plugin for us that does our heavy lifting.

__tests__/components/editor/ImportResourceTemplate.test.js
mjgiarlo: Changed and pushed! Resolving.

src/components/editor/ImportResourceTemplate.jsx
```diff
@@ -85,20 +85,44 @@ class ImportResourceTemplate extends Component {
...
+      // Response not OK, so append message to list of current messages
+      newState.message = [...this.state.message, newMessage]
+    }
+
     this.setState(newState)
   }
 
+  // Returns a URL or an empty string
+  humanReadableLocation = response => {
+    // Use Location header if it exists
```
ndushay: superfluous comment?

src/components/editor/ImportResourceTemplate.jsx
```diff
@@ -85,20 +85,44 @@ class ImportResourceTemplate extends Component {
...
+    if (response?.headers?.location)
+      return response.headers.location
+
+    // Use ID in request body if it exists, appending it to the request URL
+    if (response?.req?._data?.id) {
+      // But if the request URL already contains the ID, don't bother appending
```
ndushay: superfluous comment?

src/components/editor/ImportResourceTemplate.jsx
```diff
@@ -121,6 +145,9 @@ class ImportResourceTemplate extends Component {
       // The 0 is the profileCount which is only used for tracking create errors
```
ndushay: ok, "profileCount" confuses me here, and "create errors" too -- the errors when ... something ... is created?  what?

src/components/editor/ImportResourceTemplate.jsx
```diff
@@ -85,20 +85,44 @@ class ImportResourceTemplate extends Component {
...
       newState.modalShow = true
 
+    // `response.ok` is a Boolean that returns true if the HTTP status is in the
+    // 200-299 range, indicating the HTTP request was successful
+    if (response.ok) {
+      // Wipe out past messages and errors if response was successful
+      newState.createResourceError = []
+      newState.message = [newMessage]
+    } else {
+      // Response not OK, so append message to list of current messages
```
ndushay: superfluous comment?

src/components/editor/ImportResourceTemplate.jsx
```diff
@@ -85,20 +85,44 @@ class ImportResourceTemplate extends Component {
...
-      updateKey: updateKey
-    }
+    const newMessage = `${this.humanReadableStatus(response.status)} ${this.humanReadableLocation(response)}`
+    const newState = { updateKey: updateKey }
 
     if (showModal)
       newState.modalShow = true
 
+    // `response.ok` is a Boolean that returns true if the HTTP status is in the
+    // 200-299 range, indicating the HTTP request was successful
```
ndushay: superfluous comment?  Seems sorta SOP, and can be looked up somewhere?

src/components/editor/ImportResourceTemplate.jsx
```diff
@@ -85,20 +85,44 @@ class ImportResourceTemplate extends Component {
...
+    } else {
+      // Response not OK, so append message to list of current messages
+      newState.message = [...this.state.message, newMessage]
+    }
+
     this.setState(newState)
   }
 
+  // Returns a URL or an empty string
+  humanReadableLocation = response => {
```
ndushay: like this method naming.

ndushay:
If you're in a rush to merge, I can change to "approve" but I think the cleanup would help readability of tests and code.

__tests__/components/editor/ImportResourceTemplate.test.js
mjgiarlo: I'll rework this.

src/components/editor/ImportResourceTemplate.jsx
mjgiarlo: Removed.

src/components/editor/ImportResourceTemplate.jsx
mjgiarlo: Removed.

src/components/editor/ImportResourceTemplate.jsx
mjgiarlo: The `profileCount` stores a count of the RTs in the uploaded file, and `createResourceErrors` keeps track of errors that might happen while loading individuals RTs within a profile. I don't think I've touched related code in this branch, so not too inclined to make modifications around this.

src/components/editor/ImportResourceTemplate.jsx
mjgiarlo: removed.

src/components/editor/ImportResourceTemplate.jsx
mjgiarlo: ok.

src/components/editor/ImportResourceTemplate.jsx
jermnelson: Not sure but returning an empty string makes sense as you're building a label in the UI; if the calling function needs to react differently to the `null` then I would return null. If you remove the return statement altogether the return value in the calling function would automatically be `undefined`.

src/components/editor/ImportResourceTemplate.jsx
ndushay: fair enough.  So curiously, "profileCount" is really "resourceTemplateCount" and "createResourceErrors" is "resourceTemplateLoadErrors".  K.

PR #551, opened 2019-05-23
Minor tweaks to refactor

jgreben:
Only use one function for searching lookupConfig, use `map` instead of `forEach`, and fix propTypes


ndushay:
Here's another doc re: testing practices.  Some of it is too rspec specific (and old) but I like much of the stuff:

http://www.betterspecs.org/

This has some nice nuggets as well ("Write your commit message in the imperative: "Fix bug" and not "Fixed bug" or "Fixes bug." This convention matches
up with commit messages generated by commands like git merge and git revert.").  I need to read it carefully again:

https://medium.com/palantir/code-review-best-practices-19e02780015f

PR #555, opened 2019-05-23
Make tests and test descriptions better

jgreben:
Per Naomi's post-merge review https://github.com/LD4P/sinopia_editor/pull/546.


jgreben:
> thank you! I hope you're not just trying to appease me.

Why not?

ndushay:
b/c it's ok if you disagree; stuff like this is sorta opinionated.

PR #569, opened 2019-05-28
Redux state changes to support RDF generation

jermnelson:
The first steps for RDF generation in the Linked Data editor involves adding the following properties to the Redux state:

- `rdfPredicate` property set for propertyTemplates

### DIFF
--- a/src/reducers/index.js
+++ b/src/reducers/index.js
@@ -76,7 +76,7 @@ export const refreshResourceTemplate = (state, action) => {
 export const setResourceTemplate = (state, action) => {
   const rtKey = action.payload.id
   let output = Object.create(state)
-  action.payload.propertyTemplates.forEach((property) => {
+  action.payload.propertyTemplates.map((property) => {
     let propertyAction = {
       payload: {
         reduxPath: [rtKey, property.propertyURI],
--- a/src/reducers/index.js
+++ b/src/reducers/index.js
@@ -38,7 +38,8 @@ export const populatePropertyDefaults = (propertyTemplate) => {
   if (propertyTemplate === undefined || propertyTemplate === null || Object.keys(propertyTemplate).length < 1) {
     return defaults
   }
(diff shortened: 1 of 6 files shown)


src/reducers/index.js
```diff
@@ -38,7 +38,8 @@ export const populatePropertyDefaults = (propertyTemplate) => {
   if (propertyTemplate === undefined || propertyTemplate === null || Object.keys(propertyTemplate).length < 1) {
     return defaults
   }
-  if (propertyTemplate.valueConstraint.defaults && propertyTemplate.valueConstraint.defaults.length > 0) {
+
+  if (propertyTemplate?.valueConstraint?.defaults && propertyTemplate.valueConstraint.defaults.length > 0) {
```
mjgiarlo: Because you're using the safe navigation operator here, you can drop the first condition and shorten this to:

```javascript
if (propertyTemplate?.valueConstraint?.defaults?.length > 0) {
```

Safe navigation will return `null` anywhere along the chain that an attribute isn't defined, and `null` is not greater than 0, so the condition does what we want.

src/reducers/inputs.js
```diff
@@ -21,8 +21,11 @@ export const setMyItems = (state, action) => {
   reduxPath.reduce((obj, key) => {
     level++
     if (level === reduxPath.length) {
-      if ((key in obj) != true || !Object.keys(obj[key]).includes('items')) {
-        obj[key] = { items: [] }
+      if (!(key in obj)) {
+        obj[key] = {}
+      }
+      if (!Object.keys(obj[key]).includes('items')) {
```
mjgiarlo: Hmm, it may be JavaScript newbness, but it seems we're using two different strategies three lines apart to determine if an object contains a key. Would `if (!('items' in obj[key])) {` work here?

jermnelson:
I'm closing this draft PR but leaving the feature branch for now in case anyone wants to reference it later.

PR #570, opened 2019-05-28
Add unit tests for CognitoUtils class

### LINKED ISSUE #561: write unit tests for CognitoUtils.js
has an integration test, since sign-in is exercised, but the individual functions should be given unit tests for their basic success and failure modes.

spawned by #528

mjgiarlo:
Fixes #561

### DIFF
--- a/__tests__/CognitoUtils.test.js
+++ b/__tests__/CognitoUtils.test.js
@@ -1,0 +1,148 @@
+// Copyright 2018 Stanford University see Apache2.txt for license
+
+import { AuthenticationDetails, CognitoUser, CognitoUserPool } from 'amazon-cognito-identity-js'
+import CognitoUtils from '../src/CognitoUtils'
+import Config from '../src/Config'
+
+describe('CognitoUtils', () => {
+  const username = 'testuser'
+  const password = 'testpass'
+
+  describe('authenticationDetails()', () => {
+    it('returns an instance of AuthenticationDetails with passed credentials', () => {
+      const details = CognitoUtils.authenticationDetails(username, password)
+      expect(details).toBeInstanceOf(AuthenticationDetails)
+      expect(details.username).toEqual(username)
+      expect(details.password).toEqual(password)
+    })
+  })
+
+  describe('cognitoUserPool()', () => {
+    it('returns an instance of CognitoUserPool with configured user pool and client IDs', () => {
+      const pool = CognitoUtils.cognitoUserPool()
+      expect(pool).toBeInstanceOf(CognitoUserPool)
+      expect(pool.userPoolId).toEqual(Config.awsCognitoUserPoolId)
+      expect(pool.clientId).toEqual(Config.awsClientID)
+    })
+  })
+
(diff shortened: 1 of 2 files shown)


__tests__/CognitoUtils.test.js
```diff
@@ -0,0 +1,148 @@
...
+  })
+
+  describe('authenticateUser()', () => {
+    const user = CognitoUtils.cognitoUser(username)
+
+    afterEach(() => {
+      jest.restoreAllMocks()
+    })
+
+    it('returns a promise', () => {
```
ndushay: 👍

__tests__/CognitoUtils.test.js
```diff
@@ -0,0 +1,148 @@
...
+      const getSessionSpy = jest.spyOn(user, 'getSession').mockImplementation(callback => {
+        return callback(mockError, mockSession)
+      })
+      return CognitoUtils.getIdTokenString(user)
+        .then(result => {
+          expect(getSessionSpy).toHaveBeenCalledTimes(1)
+          expect(result).toEqual(token)
+        })
+        .catch(error => { throw error })
+    })
```
jmartin-sul: test for the failure case might be nice, but wouldn't hold up merging over it.

__tests__/CognitoUtils.test.js
```diff
@@ -0,0 +1,148 @@
...
+      expect(pool.clientId).toEqual(Config.awsClientID)
+    })
+  })
+
+  describe('cognitoUser()', () => {
+    it('returns an instance of CognitoUser with username and user pool', () => {
+      const user = CognitoUtils.cognitoUser(username)
+      expect(user).toBeInstanceOf(CognitoUser)
+      expect(user.username).toEqual(username)
+      expect(user.pool).toBeInstanceOf(CognitoUserPool)
```
jmartin-sul: i would be slightly inclined to check that the pool has the right details... but that might be overkill?

__tests__/CognitoUtils.test.js
```diff
@@ -0,0 +1,148 @@
...
+    })
+
+    it('returns a promise', () => {
+      const authResult = CognitoUtils.authenticateUser(user, password)
+      expect(authResult).toBeInstanceOf(Promise)
+    })
+
+    it('resolves on successful authentication', () => {
+      const mockSession = {}
+      const authUserSpy = jest.spyOn(user, 'authenticateUser').mockImplementation((_details, callback) => {
```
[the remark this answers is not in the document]
mjgiarlo: Good catch. Yeah, maybe `callbacksObject` to surface that the param is a legit object.

__tests__/CognitoUtils.test.js
mjgiarlo: Would be good to add. What would/should that function do when there's a failure? I don't believe it catches.

jmartin-sul: i think just bubble up the original error, though i hadn't thought about it much. we'll see if a test that effect ends up exposing weird behavior or something.

PR #571, opened 2019-05-28
Fix lookup config bug in InputListLOC

jgreben:
Fixed bug where lookup config is returning an array of objects and InputListLOC expects just a single object.

Also, removes annoying console logging which we do not really need

Also, makes sure that the lack of a setting for one property variable does not break the other variables where the property is actually set.

Adds a test for the presence of a lookupConfig URI.

### DIFF
--- a/__tests__/components/editor/InputListLOC.test.js
+++ b/__tests__/components/editor/InputListLOC.test.js
@@ -115,12 +115,17 @@ describe('<Typeahead /> component', () => {
       jest.restoreAllMocks()
     })
 
+    const defaults = [{
+      "defaultLiteral": "volume",
+      "defaultURI": "http://id.loc.gov/vocabulary/carriers/nc"
+    }]
+
     it('sets the default values according to the property template if they exist', () => {
-      const defaults = [{
-        "defaultLiteral": "volume",
-        "defaultURI": "http://id.loc.gov/vocabulary/carriers/nc"
-      }]
       expect(wrapper.instance().props.propertyTemplate.valueConstraint.defaults).toEqual(defaults)
+    })
+
+    it('sets the defaults state as the defaults array', () => {
+      expect(wrapper.state('defaults').length).toEqual(1)
     })
 
     it('logs an error when no defaults are set', () => {
--- a/__tests__/components/editor/InputListLOC.test.js
+++ b/__tests__/components/editor/InputListLOC.test.js
@@ -150,17 +150,10 @@ describe('<Typeahead /> component', () => {
         }
       }
       const infoSpy = jest.spyOn(console, 'info').mockReturnValue(null)
-      const wrapper2 = shallow(<InputList.WrappedComponent {...plProps} handleSelectedChange={mockFormDataFn} />)
+      const wrapper2 = shallow(<InputListLOC.WrappedComponent {...plProps} handleSelectedChange={mockFormDataFn} />)
 
       expect(wrapper2.state('defaults')).toEqual([])
       expect(infoSpy).toBeCalledWith(`no defaults defined in property template: ${JSON.stringify(plProps.propertyTemplate)}`)
-// =======
-//       const errorSpy = jest.spyOn(console, 'error').mockReturnValue(null)
-//       const wrapper2 = shallow(<InputListLOC.WrappedComponent {...plProps} handleSelectedChange={mockFormDataFn} />)
-//
-//       expect(wrapper2.instance().props.propertyTemplate.valueConstraint.defaults).toBeUndefined()
-//       expect(errorSpy).toBeCalledWith(`no defaults defined in property template: ${JSON.stringify(plProps.propertyTemplate)}`)
-// >>>>>>> Fix bug where lookup config is returning an array and InputListLOC expects an object
     })
   })
 
--- a/__tests__/components/editor/InputListLOC.test.js
+++ b/__tests__/components/editor/InputListLOC.test.js
@@ -3,7 +3,7 @@
 import React from 'react'
 import { shallow } from 'enzyme'
 import { Typeahead } from 'react-bootstrap-typeahead'
-import InputList from '../../../src/components/editor/InputListLOC'
+import InputListLOC from '../../../src/components/editor/InputListLOC'
 
 const plProps = {
   "propertyTemplate":
@@ -31,10 +31,52 @@ const plProps = {
     }
 }
 
-describe('<InputList />', () => {
+const mockLookupConfig = [
+  {
+    "label": "carriers",
+    "uri": "https://id.loc.gov/vocabulary/carriers",
+    "component": "list"
+  },
+  {
+    "label": "frequency",
+    "uri": undefined,
+    "component": "list"
+  }
+]
+
+describe('<InputListLOC /> configuration', () => {
   // our mock formData function to replace the one provided by mapDispatchToProps
   const mockFormDataFn = jest.fn()
-  const wrapper = shallow(<InputList.WrappedComponent {...plProps} handleSelectedChange={mockFormDataFn} />)
+
+  beforeEach(() => {
+    global.alert = jest.fn()
+  })
+
+  afterEach(() => {
+    global.alert = alert
+  })
+
+  it('expects a single lookupConfig object', () => {
+    shallow(<InputListLOC.WrappedComponent {...plProps} handleSelectedChange={mockFormDataFn} lookupConfig={mockLookupConfig[0]} />)
+    expect(global.alert.mock.calls.length).toEqual(0)
+  })
+
+  it('displays a browser alert if the lookupConfig is undefined', () => {
+    shallow(<InputListLOC.WrappedComponent {...plProps} handleSelectedChange={mockFormDataFn} lookupConfig={mockLookupConfig[1]} />)
+    expect(global.alert.mock.calls.length).toEqual(1)
+  })
+
+  it('displays a browser alert if the lookupConfig is an array of objects and not a single object', () => {
+    shallow(<InputListLOC.WrappedComponent {...plProps} handleSelectedChange={mockFormDataFn} lookupConfig={mockLookupConfig} />)
+    expect(global.alert.mock.calls.length).toEqual(1)
+  })
+
+})
+
+describe('<Typeahead /> component', () => {
+  // our mock formData function to replace the one provided by mapDispatchToProps
+  const mockFormDataFn = jest.fn()
+  const wrapper = shallow(<InputListLOC.WrappedComponent {...plProps} handleSelectedChange={mockFormDataFn} lookupConfig={mockLookupConfig[0]} />)
 
   it('contains a placeholder with the value of propertyLabel', () => {
     expect(wrapper.find(Typeahead).props().placeholder).toMatch("Frequency (RDA 2.14)")
@@ -75,11 +117,10 @@ describe('<InputList />', () => {
 
     it('sets the default values according to the property template if they exist', () => {
       const defaults = [{
-        id: 'http://id.loc.gov/vocabulary/carriers/nc',
-        uri: 'http://id.loc.gov/vocabulary/carriers/nc',
-        label: 'volume'
+        "defaultLiteral": "volume",
+        "defaultURI": "http://id.loc.gov/vocabulary/carriers/nc"
       }]
-      expect(wrapper.state('defaults')).toEqual(defaults)
+      expect(wrapper.instance().props.propertyTemplate.valueConstraint.defaults).toEqual(defaults)
     })
 
     it('logs an error when no defaults are set', () => {
@@ -108,6 +149,13 @@ describe('<InputList />', () => {
 
       expect(wrapper2.state('defaults')).toEqual([])
       expect(infoSpy).toBeCalledWith(`no defaults defined in property template: ${JSON.stringify(plProps.propertyTemplate)}`)
+// =======
+//       const errorSpy = jest.spyOn(console, 'error').mockReturnValue(null)
+//       const wrapper2 = shallow(<InputListLOC.WrappedComponent {...plProps} handleSelectedChange={mockFormDataFn} />)
+//
+//       expect(wrapper2.instance().props.propertyTemplate.valueConstraint.defaults).toBeUndefined()
+//       expect(errorSpy).toBeCalledWith(`no defaults defined in property template: ${JSON.stringify(plProps.propertyTemplate)}`)
+// >>>>>>> Fix bug where lookup config is returning an array and InputListLOC expects an object
     })
   })
 
--- a/__tests__/components/editor/InputLookupQA.test.js
+++ b/__tests__/components/editor/InputLookupQA.test.js
@@ -14,6 +14,7 @@ const plProps = {
       "type": "lookup",
       "resourceTemplates": [],
       "valueConstraint": {
+        "repeatable": "true",
         "valueTemplateRefs": [],
         "useValuesFrom": [
           'lookupQaLocNames'
--- a/src/components/editor/InputListLOC.jsx
+++ b/src/components/editor/InputListLOC.jsx
@@ -5,7 +5,7 @@ import { Typeahead } from 'react-bootstrap-typeahead'
 import PropTypes from 'prop-types'
 import PropertyRemark from './PropertyRemark'
 import { connect } from 'react-redux'
-import { changeSelections } from '../../actions/index'
+import { changeSelections  } from '../../actions/index'
 import { defaultValuesFromPropertyTemplate, booleanPropertyFromTemplate } from '../../Utilities'
 import shortid from 'shortid'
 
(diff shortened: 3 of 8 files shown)


src/components/editor/InputListLOC.jsx
```diff
@@ -14,9 +14,9 @@ class InputListLOC extends Component {
     this.state = {
       isLoading: false,
       options: [],
-      defaults: []
+      defaults: [],
+      lookupUri: this.props.lookupConfig.uri
```
mjgiarlo: If we're not modifying this value via `this.setState()`, what value do we get out of adding it to component state? Can we use `this.props.lookupConfig.uri` throughout the component instead?

mjgiarlo: Alternatively, if this component only needs the URI and does not need the rest of `lookupConfig`, could the parent component inject just the URI?

src/components/editor/InputListLOC.jsx
```diff
@@ -14,9 +14,9 @@ class InputListLOC extends Component {
...
     }
-    this.hasPropertyRemark = this.hasPropertyRemark.bind(this)
```
mjgiarlo: Big :+1: 	to this change and the related one on line 44!

src/components/editor/InputListLOC.jsx
```diff
@@ -41,7 +41,7 @@ class InputListLOC extends Component {
     this.props.handleSelectedChange(payload)
   }
 
-  hasPropertyRemark(propertyTemplate) {
+  hasPropertyRemark = (propertyTemplate) => {
```
mjgiarlo: Parens not needed in this context, but they don't break anything and we're already using both styles throughout this codebase, so no biggie here.

src/components/editor/InputListLOC.jsx
```diff
@@ -50,13 +50,22 @@ class InputListLOC extends Component {
   }
 
   render() {
-    let lookupUri, isMandatory, isRepeatable
+    let isMandatory, isRepeatable
+
+    if (this.state.lookupUri === undefined) {
+      console.error("lookup URI was not defined!")
```
mjgiarlo: If the lookup URI is undefined, what is the ultimate effect? Is it rendering the the `Typeahead` component with an `onFocus` value of `false`? Could/should we surface that error condition earlier rather than just logging an error?

src/components/editor/InputListLOC.jsx
```diff
@@ -50,13 +50,22 @@ class InputListLOC extends Component {
...
+    }
+
     try {
       isMandatory = JSON.parse(this.props.propertyTemplate.mandatory)
+    } catch (error) {
+      //We do not really care about this error. If there is an error the variable will be undefined and hence false
```
mjgiarlo: What causes the errors on lines 62 and 68? I wonder if we want to use a different strategy here, lest we eat errors that we don't anticipate. I'm a bit leery of using `try`/`catch` for flow control. For instance, if the situation we're coding around here is a property template that hasn't set the mandatory attribute, perhaps we should replace this with:

```javascript
const isMandatory = this.props.propertyTemplate.mandatory !== undefined ? this.props.propertyTemplate.mandatory : false
```

src/components/editor/InputListLOC.jsx
```diff
@@ -50,13 +50,22 @@ class InputListLOC extends Component {
...
+    }
+
+    try {
       isRepeatable = JSON.parse(this.props.propertyTemplate.valueConstraint.repeatable)
-      lookupUri = this.props.lookupConfig.uri
     } catch (error) {
-      console.error(`Some properties were not defined in the json file: ${error}`)
+      //We do not really care about this error. If there is an error the variable will be undefined and hence false
```
mjgiarlo: See comment on line 62.

src/components/editor/InputListLOC.jsx
ndushay: in a perfect world, would this be an alert to the user?

__tests__/components/editor/InputListLOC.test.js
```diff
@@ -95,4 +100,9 @@ describe('<InputList />', () => {
   it('sets the formData store with the total number or objects sent to "selected" when a change event happens', () => {
     expect(mockFormDataFn.mock.calls.length).toBe(2)
   })
+
+  it('gets the LOC lookup URL from the config based on the URI in useValuesFrom ', () => {
```
mjgiarlo: This test would fail w/o the changes you made to the two source files below, I assume.

src/components/editor/PropertyComponent.jsx
```diff
@@ -33,7 +33,7 @@ export class PropertyComponent extends Component {
         break
       case "list":
         result = (<InputListLOC key = {this.props.index} reduxPath={[this.props.rtId, property.propertyURI]}
-                                propertyTemplate = {property} lookupConfig = {this.state.configuration}/>)
+                                propertyTemplate = {property} lookupConfig = {this.state.configuration[0]}/>)
```
mjgiarlo: It would be ideal to add a test covering this change.

__tests__/components/editor/InputListLOC.test.js
```diff
@@ -3,38 +3,43 @@
...
+  }
+}
+
+const mockLookupConfig = {
+  "label": "frequency",
+  "uri": "https://id.loc.gov/vocabulary/frequencies",
+  "component": "list"
 }
 
 describe('<InputList />', () => {
```
ndushay: InputListLOC?

__tests__/components/editor/InputListLOC.test.js
jgreben: Indeed:
```
 Expected: "https://id.loc.gov/vocabulary/frequencies"
    Received: undefined

      103 | 
      104 |   it('gets the LOC lookup URL from the config based on the URI in useValuesFrom ', () => {
    > 105 |     expect(wrapper.state('lookupUri')).toEqual(mockLookupConfig.uri)
          |                                        ^
```

src/components/editor/InputListLOC.jsx
jgreben: Yeah, not sure what I was thinking. I changed it to `this.props.lookupConfig.uri`.

mjgiarlo: @jgreben :speech_balloon: 
> Yeah, not sure what I was thinking. 

:laughing: I'm always trying stuff in PRs and then forgetting to back it out. Part of the process!

> I changed it to `this.props.lookupConfig.uri`.

Thank you, sir.

src/components/editor/InputListLOC.jsx
jgreben: I personally like the parens, but if we all wanted to agree on a style and make it so throughout the code, I would stick to it.

__tests__/components/editor/InputListLOC.test.js
mjgiarlo: @jgreben 	Most excellent!

src/components/editor/InputListLOC.jsx
mjgiarlo: Agreed. Ideally we would lean on eslint to do this for us, but I'm loath to suggest we make nitpicky style changes at this point.

src/components/editor/InputListLOC.jsx
jgreben: So theoretically what if they set the mandatory: "foo" and/or repeatable: "bar"? They are not undefined. The Typeahead component does not break but throws a warning: 

 ```
Warning: Failed prop type: Invalid prop `multiple` of type `string` supplied to `TypeaheadContainer(WrappedTypeahead)`, expected `boolean`.
```

The schema validation probably checks that those values are "true" or "false", but we also have clever librarians who like to hand-edit their resource templates.

I like the syntax you provided and I'm happy to make that change but I'd like to get your thoughts first 👆.

File: src/components/editor/InputListLOC.jsx (hunk 6 above)
Comment (jgreben) [reply src/components/editor/InputListLOC.jsx]: 👍 👆

File: src/components/editor/PropertyComponent.jsx (hunk 8 above)
Comment (jgreben) [reply src/components/editor/PropertyComponent.jsx]: Should this satisfy it? 
https://github.com/LD4P/sinopia_editor/blob/4c2925b7e2bb23c1b64303158d4a8977ec4963e7/__tests__/components/editor/PropertyComponent.test.js#L30

File: src/components/editor/InputListLOC.jsx (hunk 5 above)
Comment (mjgiarlo) [reply src/components/editor/InputListLOC.jsx]: @jgreben for added robustness, do you think we ought to add a new function to the `Utilities` class? something like:

```javascript
export const booleanPropertyFromTemplate = (template, key, defaultvalue) => {
  // Use safe navigation for dynamic properties: https://github.com/tc39/proposal-optional-chaining#syntax
  const propertyValue = template?.[key]

  if (!propertyValue)
    return defaultValue

  parsedValue = JSON.parse(propertyValue)
  
  if (parsedValue !== true && parsedValue !== false)
    return defaultValue

  return parsedValue
}
```

:question:

src/components/editor/PropertyComponent.jsx
mjgiarlo: I was thinking a typecheck test, yeah. What I had in mind was a test that made sure that a child `InputListLOC` component had a `lookupConfig` prop that looked how we expected it.

__tests__/components/editor/InputListLOC.test.js
jgreben: If I change it to look at props instead of state I think this test should be moved to the parent component.

jgreben: nevermind. I figured out a better way. stay tuned.

src/components/editor/InputListLOC.jsx
jgreben: You would just see this:

Screen Shot 2019-05-29 at 8 27 55 AM

jgreben: I added an alert.

src/components/editor/InputListLOC.jsx
jgreben: I like this, but how it will work with (e.g.) `this.props.propertyTemplate.valueConstraint.repeatable`?

mjgiarlo: @jgreben In that context, I reckon it could be invoked thusly:

```javascript
booleanPropertyFromTemplate(this.props.propertyTemplate.valueConstraint, 'repeatable', false)
```

So maybe the function and first arg could be renamed so it's clear that a property template is not required, but rather any ol' object?

jgreben: Ah nevermind, I was thinking that `template` was always the whole property template, but in the second case it would actually be the valueConstraint object. Maybe template is not the right var name. I will add this to the next commit and you can judge.

mjgiarlo: awesome! thanks, @jgreben

jgreben: I don't like that the optional chaining syntax makes my Webstorm editor complain...but I guess that's my own problem!

mjgiarlo: Emacs complains at me too. Ha!

__tests__/components/editor/InputListLOC.test.js
```diff
@@ -3,38 +3,44 @@
...
-describe('<InputList />', () => {
+const mockLookupConfig = {
+  "label": "frequency",
+  "uri": undefined,
+  "component": "list"
+}
+
+describe('<InputListLOC />', () => {
   // our mock formData function to replace the one provided by mapDispatchToProps
+  global.alert = jest.fn()
```
mjgiarlo: Does this leak into other tests? Should we restore mocks after this test?

__tests__/components/editor/InputListLOC.test.js
```diff
@@ -3,38 +3,44 @@
...
   const mockFormDataFn = jest.fn()
-  const wrapper = shallow(<InputList.WrappedComponent {...plProps} handleSelectedChange={mockFormDataFn} />)
+  const wrapper = shallow(<InputListLOC.WrappedComponent {...plProps} handleSelectedChange={mockFormDataFn} lookupConfig={mockLookupConfig} />)
```
mjgiarlo: Dig these changes!

__tests__/components/editor/InputListLOC.test.js
```diff
@@ -95,4 +101,9 @@ describe('<InputList />', () => {
   it('sets the formData store with the total number or objects sent to "selected" when a change event happens', () => {
     expect(mockFormDataFn.mock.calls.length).toBe(2)
   })
+
+  it('displays a browser alert if the lookupConfig is undefined', () => {
+    expect(global.alert.mock.calls.length).toBeGreaterThanOrEqual(1)
```
mjgiarlo: Hmm, is the number of call times not deterministic?

__tests__/components/editor/InputLookupQA.test.js
```diff
@@ -13,6 +13,7 @@ const plProps = {
       "type": "lookup",
       "resourceTemplates": [],
       "valueConstraint": {
+        "repeatable": "true",
```
mjgiarlo: Not a big deal, but am curious why this was added.

src/components/editor/InputListLOC.jsx
```diff
@@ -50,13 +49,22 @@ class InputListLOC extends Component {
   }
 
   render() {
-    let lookupUri, isMandatory, isRepeatable
+    let isMandatory, isRepeatable
+
+    if (this.props.lookupConfig.uri === undefined) {
+      alert(`There is no configured list lookup for ${this.props.propertyTemplate.propertyURI}`)
```
mjgiarlo: :+1:

src/components/editor/InputListLOC.jsx
```diff
@@ -130,3 +138,4 @@ const mapDispatchtoProps = dispatch => ({
 })
 
 export default connect(mapStatetoProps, mapDispatchtoProps)(InputListLOC)
+export { InputListLOC }
```
mjgiarlo: I haven't seen this much in our components. What does this do?

src/components/editor/InputLookupQA.jsx
```diff
@@ -72,106 +72,111 @@ class InputLookupQA extends Component {
...
-            isLoading: this.state.isLoading,
-            options: this.state.options,
-            selected: this.state.selected,
-            delay: 300
-        }
-
-        return (
-            <div>
-                <AsyncTypeahead id="lookupComponent"
+    return (
```
mjgiarlo: I'm guessing this is going to be... annoying to rebase on `master` now. Let me know if I can do anything to help.

src/components/editor/InputListLOC.jsx
jgreben: Opps, that was an experiment I forgot to remove. It was supposed to export the "undecorated" version if the component (i.e. the unwrapped component). I'm not sure if it actually did or not, but I discarded trying to use that.

__tests__/components/editor/InputListLOC.test.js
jgreben: I moved the test to the top and then restored the mock, and now it it just a single call.

src/components/editor/InputLookupQA.jsx
jgreben: I'm afraid to find out what you mean....

__tests__/components/editor/InputLookupQA.test.js
jgreben: See line 58 comment on `InputListLOC.jsx` below

src/components/editor/InputListLOC.jsx
```diff
@@ -50,15 +50,13 @@ class InputListLOC extends Component {
...
-      isRepeatable = JSON.parse(this.props.propertyTemplate.valueConstraint.repeatable)
-      lookupUri = this.props.lookupConfig.uri
-    } catch (error) {
-      console.error(`Some properties were not defined in the json file: ${error}`)
+    if (this.props.lookupConfig.uri === undefined) {
+      alert(`There is no configured list lookup for ${this.props.propertyTemplate.propertyURI}`)
     }
 
+    const isMandatory = booleanPropertyFromTemplate(this.props.propertyTemplate, 'mandatory', false)
+    const isRepeatable = booleanPropertyFromTemplate(this.props.propertyTemplate.valueConstraint, 'repeatable', true)
```
jgreben: @mjgiarlo the test was wrong as it was referencing the wrong repeatable value.

__tests__/components/editor/InputListLOC.test.js
mjgiarlo: excellent

src/components/editor/InputListLOC.jsx
mjgiarlo: ah, OK

jermnelson:
When doing a smoke check on this PR, noticed that the defaults are not showing up for `InputListQA`.

mjgiarlo:
related to #576

jgreben:
> When doing a smoke check on this PR, noticed that the defaults are not showing up for `InputListLOC`.

@jermnelson this should be fixed now.

src/components/editor/InputListLOC.jsx
```diff
@@ -118,14 +112,14 @@ InputListLOC.propTypes = {
   }).isRequired
 }
 
-const mapStatetoProps = (state) => {
+const mapStateToProps = (state) => {
```
mjgiarlo: Good catch!

PR #574, opened 2019-05-29
Extract and pass along property template defaults from InputLookupQA component to typeahead

### LINKED ISSUE #568: default value does not show up for type = lookup with QA lookups
for Property Template with type = lookup, the profile creator can include one or both of default URI and default value.

when I test this with a type = lookup that uses id.loc.gov, the default shows up in the Linked Data Editor as expected

but with a QA Lookup, the default value does not show up in the Linked Data Editor.

I know lookups are still being fixed so perhaps this is related, but making this issue now so we can test after lookups are fixed.

Reosurce Template to use for testing (2 property templates both type = lookup, both with defaults, one to a QA lookup, one to a LC lookup)

- QA lookup has default value `coffee`, it doesn't show up in Editor
- loc lookup has default value `color` and it appears properly

[test of defaults with lookups.json.txt](https://github.com/LD4P/sinopia_editor/files/3218997/test.of.defaults.with.lookups.json.txt)

mjgiarlo:
Fixes #568

Includes some refactoring of PropertyComponent and InputListLOC components

### DIFF
--- a/__tests__/components/editor/PropertyComponent.test.js
+++ b/__tests__/components/editor/PropertyComponent.test.js
@@ -5,9 +5,7 @@ import { shallow } from 'enzyme'
 import PropertyComponent from "../../../src/components/editor/PropertyComponent"
 
 describe('<PropertyComponent />', () => {
-
   describe('sets the configuration state based on values from the property template', () => {
-
     describe('for property templates configured as list', () => {
       const template = {
         "propertyURI": "http://id.loc.gov/ontologies/bibframe/issuance",
@@ -62,31 +60,28 @@ describe('<PropertyComponent />', () => {
       it('renders the lookup component', () => {
         expect(wrapper.find('Connect(InputLookupQA)').length).toEqual(1)
       })
-
     })
   })
 
   describe('when there are no configuration values from the property template', () => {
-
     describe('for unconfigured templates of type:literal', () => {
       const template = {
-        "propertyURI": "http://id.loc.gov/ontologies/bibframe/heldBy",
+        "propertyURI": "http//:id.loc.gov/ontologies/bibframe/heldBy",
         "type": "literal"
       }
 
       const wrapper = shallow(<PropertyComponent propertyTemplate={template} />)
 
-      it('will return an empty array', () => {
+      it('returns an empty array', () => {
         expect(wrapper.state('configuration').length).toEqual(0)
       })
 
       it('renders an InputLiteral component', () => {
         expect(wrapper.find('Connect(InputLiteral)').length).toEqual(1)
       })
-
     })
 
-    it('if the property type is not literal (i.e. resource) an empty div is returned', () => {
+    it('returns an empty div if the property type is not literal (i.e. resource)', () => {
       const template = {
         "propertyURI": "http://id.loc.gov/ontologies/bibframe/note",
         "type": "resource",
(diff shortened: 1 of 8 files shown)


src/Utilities.js
```diff
@@ -1,18 +1,30 @@
...
-  return Boolean(
-    ( property.hasOwnProperty('type') && property.type === 'resource' ) &&
-    ( property.hasOwnProperty('valueConstraint') &&
-      ( property.valueConstraint.hasOwnProperty('valueTemplateRefs')  &&
-       property.valueConstraint.valueTemplateRefs.length > 0))
-  )
+const _ = require('lodash')
+
+export const isResourceWithValueTemplateRef = property => {
+  return property?.type === 'resource' &&
```
mjgiarlo: Use the safe navigation operator to simplify and shorten this function.

src/Utilities.js
```diff
@@ -1,18 +1,30 @@
...
+    property?.valueConstraint?.valueTemplateRefs?.length > 0
 }
 
-export const resourceToName = (uri) => {
-  try{
-    return uri.substr(uri.lastIndexOf('/') + 1)
-  } catch {
+export const resourceToName = uri => {
```
mjgiarlo: Rather than a bare `catch`, explicitly check if passed URI is a string.

src/Utilities.js
```diff
@@ -1,18 +1,30 @@
...
+  if (!_.isString(uri))
     return undefined
-  }
+
+  return uri.substr(uri.lastIndexOf('/') + 1)
+}
+
+
+export const defaultValuesFromPropertyTemplate = propertyTemplate => {
```
mjgiarlo: Extract this function from input components to utilities class.

src/components/editor/InputListLOC.jsx
```diff
@@ -6,29 +6,27 @@ import PropTypes from 'prop-types'
...
-      const defaultValue = this.props.propertyTemplate.valueConstraint.defaults[0]
-      const defaults = [{
-        id: defaultValue.defaultURI,
-        label: defaultValue.defaultLiteral,
-        uri: defaultValue.defaultURI
-      }]
-      this.state.defaults = defaults
+    const defaults = defaultValuesFromPropertyTemplate(this.props.propertyTemplate)
+
+    if (defaults.length > 0) {
```
mjgiarlo: Instead of using try/catch for flow control, use `if`.

src/components/editor/InputLookupQA.jsx
```diff
@@ -6,197 +6,195 @@ import Swagger from 'swagger-client'
 import { connect } from 'react-redux'
 import { getProperty } from '../../reducers/index'
 import { changeSelections } from '../../actions/index'
+import { defaultValuesFromPropertyTemplate } from '../../Utilities'
 
-const AsyncTypeahead = asyncContainer( Typeahead )
+const AsyncTypeahead = asyncContainer(Typeahead)
 
 class InputLookupQA extends Component {
-    constructor( props ) {
```
mjgiarlo: This component, for some reason, was using 4-space indentation instead of 2-space indentation. That is making this diff look a bit larger than it actually is. Suggest toggling "No Whitespace" button in GitHub UI.

__tests__/components/editor/PropertyComponent.test.js
```diff
@@ -62,31 +60,28 @@ describe('<PropertyComponent />', () => {
...
-
     })
   })
 
   describe('when there are no configuration values from the property template', () => {
-
     describe('for unconfigured templates of type:literal', () => {
       const template = {
-        "propertyURI": "http://id.loc.gov/ontologies/bibframe/heldBy",
+        "propertyURI": "http:id.loc.gov/ontologies/bibframe/heldBy",
```
jcoyne: Is this intentional?

mjgiarlo: no! I didn't even see that. Will change. Thanks!

mjgiarlo: done. pushed. resolving.

__tests__/components/editor/PropertyComponent.test.js
```diff
@@ -62,31 +60,28 @@ describe('<PropertyComponent />', () => {
...
-
     })
   })
 
   describe('when there are no configuration values from the property template', () => {
-
     describe('for unconfigured templates of type:literal', () => {
       const template = {
-        "propertyURI": "http://id.loc.gov/ontologies/bibframe/heldBy",
+        "propertyURI": "http//:id.loc.gov/ontologies/bibframe/heldBy",
```
ndushay: what????  it's `http://` isn't it?

mjgiarlo: @ndushay 	hahaha. small patch coming. :roll_eyes:

PR #577, opened 2019-05-29
Add reduxPath to PropertyComponent

### LINKED ISSUE #575: defaults for type = literal are not appearing on DEV
Using this profile https://github.com/LD4P/sinopia_sample_profiles/blob/master/configuration_demo/ConfigurationDemoFuller.json
on DEV
fields 1, 4, 6, and 7 are fields of type = literal with default values

the default value should be presented to the user

this is working as expected in stage (and was working in DEV recently?)
Screen Shot 2019-05-28 at 10 28 28 PM

but currently it's not working in DEV
Screen Shot 2019-05-28 at 10 29 12 PM

jermnelson:
As @michelleif noted in #575, the defaults were not showing up because the `reduxPath` was not set  in the PropertyComponent component. 

Fixes #575

### DIFF
--- a/__tests__/components/editor/PropertyComponent.test.js
+++ b/__tests__/components/editor/PropertyComponent.test.js
@@ -17,7 +17,7 @@ describe('<PropertyComponent />', () => {
         }
       }
 
-      const wrapper = shallow(<PropertyComponent propertyTemplate={template} />)
+      const wrapper = shallow(<PropertyComponent propertyTemplate={template} reduxPath={["http://id.loc.gov/ontologies/bibframe/issuance"]} />)
 
       it('should find a configuration object in the lookup config', () => {
         expect(typeof wrapper.state('configuration')).toEqual("object")
@@ -46,7 +46,7 @@ describe('<PropertyComponent />', () => {
         }
       }
 
-      const wrapper = shallow(<PropertyComponent propertyTemplate={template} />)
+      const wrapper = shallow(<PropertyComponent propertyTemplate={template} reduxPath={["http://id.loc.gov/ontologies/bibframe/contribution"]}/>)
 
       it('finds a configuration object in the lookup config', () => {
         expect(typeof wrapper.state('configuration')).toEqual("object")
@@ -70,7 +70,8 @@ describe('<PropertyComponent />', () => {
         "type": "literal"
       }
 
-      const wrapper = shallow(<PropertyComponent propertyTemplate={template} />)
+      const wrapper = shallow(<PropertyComponent propertyTemplate={template}
+                                reduxPath={["http://id.loc.gov/ontologies/bibframe/heldBy"]} />)
 
       it('returns an empty array', () => {
         expect(wrapper.state('configuration').length).toEqual(0)
@@ -93,7 +94,8 @@ describe('<PropertyComponent />', () => {
         }
       }
 
-      const wrapper = shallow(<PropertyComponent propertyTemplate={template} />)
+      const wrapper = shallow(<PropertyComponent propertyTemplate={template}
+                               reduxPath={["http://id.loc.gov/ontologies/bibframe/note"]}/>)
       expect(wrapper.find('Connect(InputListLOC)').length).toEqual(0)
       expect(wrapper.find('Connect(InputLookupQA)').length).toEqual(0)
       expect(wrapper.find('Connect(InputLiteral)').length).toEqual(0)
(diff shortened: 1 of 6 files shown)


__tests__/components/editor/PropertyComponent.test.js
```diff
@@ -98,11 +99,32 @@ describe('<PropertyComponent />', () => {
...
-      const wrapper = shallow(<PropertyComponent propertyTemplate={template} />)
+      const wrapper = shallow(<PropertyComponent propertyTemplate={template}
+                               reduxPath={["http://id.loc.gov/ontologies/bibframe/note"]}/>)
       expect(wrapper.find('Connect(InputListLOC)').length).toEqual(0)
       expect(wrapper.find('Connect(InputLookupQA)').length).toEqual(0)
       expect(wrapper.find('Connect(InputLiteral)').length).toEqual(0)
     })
   })
 
+  it('error if <PropertyComponent /> is missing reduxPath props', () => {
```
mjgiarlo: I think this test description is missing the initial verb, like:

```javascript
it('logs an error if <PropertyComponent /> is missing the reduxPath property', () => {
```

__tests__/components/editor/PropertyComponent.test.js
```diff
@@ -98,11 +99,32 @@ describe('<PropertyComponent />', () => {
...
+    const template = {
```
mjgiarlo: Does this test require its own `template` or can it use the `template` defined above? It'd help keep this test super focused on testing the `reduxPath` prop. No biggie either way, though.

__tests__/components/editor/PropertyComponent.test.js
```diff
@@ -98,11 +99,32 @@ describe('<PropertyComponent />', () => {
...
+      "propertyURI": "http://id.loc.gov/ontologies/bibframe/note",
+      "type": "resource",
+      "valueConstraint": {
+        "valueTemplateRefs": [
+          "resourceTemplate:bf2:Note"
+        ],
+        "useValuesFrom": []
+      }
+    }
+    console.error = jest.fn()
```
mjgiarlo: Does this leak outside of this test? If so, we should restore this mock when this test finishes.

__tests__/components/editor/PropertyComponent.test.js
```diff
@@ -98,11 +99,32 @@ describe('<PropertyComponent />', () => {
...
+    const wrapper = shallow(<PropertyComponent index={1}
```
mjgiarlo: If we're only using the `wrapper` variable to appease eslint on line 125, what about removing the variable entirely? So, like:

```javascript
shallow(<PropertyComponent index={1} propertyTemplate={template} rtId={'resourceTemplate:test'} />)
expect(console.error).toHaveBeenCalledTimes(1)
```

src/components/editor/PropertyComponent.jsx
```diff
@@ -93,7 +93,7 @@ PropertyComponent.propTypes = {
       useValuesFrom: PropTypes.oneOfType([ PropTypes.string, PropTypes.array])
     })
   }).isRequired,
-  reduxPath: PropTypes.array,
+  reduxPath: PropTypes.array.isRequired,
```
mjgiarlo: :+1:

__tests__/components/editor/PropertyComponent.test.js
jermnelson: Good idea; fixed in commit.

__tests__/components/editor/PropertyComponent.test.js
jermnelson: Fixed in latest commit; stored original `console.error` and then restore it to prevent leakage as `console.error.mockRestore()` didn't seem to work.

__tests__/components/editor/PropertyComponent.test.js
jermnelson: Would need to pull out the previous `template` into the enclosing `describe` block; I personally like the the template here to provide context but maybe this particular test should be in it's `describe` block.

__tests__/components/editor/PropertyComponent.test.js
jermnelson: Fixed in commit.

__tests__/components/editor/PropertyComponent.test.js
mjgiarlo: ok, no problem!

PR #579, opened 2019-05-29
Add instructions about building the distribution

jcoyne:
Without this step you get a syntax error when running the server


README.md
```diff
@@ -18,6 +18,10 @@ Technical documentation specific to the Sinopia BIBFRAME Editor may also be foun
 4.  Get latest npm: `npm install -g npm@latest`
 5.  Run `npm install`. This installs everything needed for the build to run successfully.
 
+## Build the distribution
+
+`npm run-script build`
```
mjgiarlo: @jcoyne I think `npm run build` works too and we already reference that elsewhere in the README (see https://github.com/LD4P/sinopia_editor#building-with-webpack). Mind tweaking?

PR #585, opened 2019-05-30
Fix skipped app tests

### LINKED ISSUE #539: fix tests skipped by amplify session management PR
spawned by #528 

- [x] fix skipped tests in `__tests__/components/App.test.js` (which are about `App` component detecting whether user is logged in)
- [x] ~~fix skipped tests in `__tests__/components/editor/ImportResourceTemplate.test.js` (which are about the component hitting Sinopia Server, which requires mocking some Cognito functionality)~~ (had a fix for this, but it was obviated by some concurrent refactoring work from @mjgiarlo)

jmartin-sul:
this fixes the `App` component tests skipped to get #460 merged more quickly, and in the process of that, does a bit of restructuring (in hope that the new organization of the tests is a bit simpler and more regular).
* providing only the supporting components needed to get `App` to have a working redux store and a working react router allowed for faking specific redux states and URI visits (to exercise particular authentication-related behavior in combo with particular URI renderings).  this meant using `Provider` with a fake store to wrap `App`, instead of relying on `RootContainer` (which will already have its own redux store).
* a couple of components included by `App` are also mocked, to prevent their implementation details from interfering with specific unit testing of `App` (since the `App` tests are more about whether the components are rendered or not in given conditions, and not about e.g. the interactions of those components with the back end).
  * i was tempted to mock all the components included by `App`, but then i didn't.

the second commit makes the touchups mentioned in #570.

fixes #539

### DIFF
--- a/__tests__/CognitoUtils.test.js
+++ b/__tests__/CognitoUtils.test.js
@@ -32,6 +32,7 @@ describe('CognitoUtils', () => {
       expect(user).toBeInstanceOf(CognitoUser)
       expect(user.username).toEqual(username)
       expect(user.pool).toBeInstanceOf(CognitoUserPool)
+      expect(user.pool).toEqual(CognitoUtils.cognitoUserPool())
     })
   })
 
@@ -49,8 +50,8 @@ describe('CognitoUtils', () => {
 
     it('resolves on successful authentication', () => {
       const mockSession = {}
-      const authUserSpy = jest.spyOn(user, 'authenticateUser').mockImplementation((_details, callback) => {
-        return callback.onSuccess(mockSession)
+      const authUserSpy = jest.spyOn(user, 'authenticateUser').mockImplementation((_details, callbacksObject) => {
+        return callbacksObject.onSuccess(mockSession)
       })
       return CognitoUtils.authenticateUser(user, password)
         .then(authResult => {
@@ -62,8 +63,8 @@ describe('CognitoUtils', () => {
 
     it('rejects on failed authentication', () => {
       const mockError = {}
-      const authUserSpy = jest.spyOn(user, 'authenticateUser').mockImplementation((_details, callback) => {
-        return callback.onFailure(mockError)
+      const authUserSpy = jest.spyOn(user, 'authenticateUser').mockImplementation((_details, callbacksObject) => {
+        return callbacksObject.onFailure(mockError)
       })
       return CognitoUtils.authenticateUser(user, password)
         .then(result => { throw result })
@@ -89,8 +90,8 @@ describe('CognitoUtils', () => {
     it('resolves with session data when successful', () => {
       const mockError = null
       const mockSession = {}
-      const getSessionSpy = jest.spyOn(user, 'getSession').mockImplementation(callback => {
-        return callback(mockError, mockSession)
+      const getSessionSpy = jest.spyOn(user, 'getSession').mockImplementation(callbackFn => {
+        return callbackFn(mockError, mockSession)
       })
       return CognitoUtils.getSession(user)
         .then(result => {
@@ -102,8 +103,8 @@ describe('CognitoUtils', () => {
 
     it('rejects with error info when failing', () => {
       const mockError = 'uh oh!'
-      const getSessionSpy = jest.spyOn(user, 'getSession').mockImplementation(callback => {
-        return callback(mockError)
+      const getSessionSpy = jest.spyOn(user, 'getSession').mockImplementation(callbackFn => {
+        return callbackFn(mockError)
       })
       return CognitoUtils.getSession(user)
         .then(result => { throw result })
@@ -134,8 +135,8 @@ describe('CognitoUtils', () => {
           jwtToken: token
         }
       }
-      const getSessionSpy = jest.spyOn(user, 'getSession').mockImplementation(callback => {
-        return callback(mockError, mockSession)
+      const getSessionSpy = jest.spyOn(user, 'getSession').mockImplementation(callbackFn => {
+        return callbackFn(mockError, mockSession)
       })
       return CognitoUtils.getIdTokenString(user)
         .then(result => {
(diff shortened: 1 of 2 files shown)


__tests__/components/App.test.js
```diff
@@ -4,14 +4,18 @@ import 'jsdom-global/register'
...
+import LoginPanel from '../../src/components/LoginPanel'
 import HomePage from '../../src/components/HomePage'
 import Editor from '../../src/components/editor/Editor'
 import Browse from '../../src/components/editor/Browse'
+import CanvasMenu from '../../src/components/CanvasMenu'
 import Footer from '../../src/components/Footer'
-import ImportResourceTemplate from "../../src/components/editor/ImportResourceTemplate";
+import ImportResourceTemplate from '../../src/components/editor/ImportResourceTemplate';
+
+jest.mock('../../src/components/editor/Editor')
```
jmartin-sul: wasn't clear to me what exactly, but seems like `Editor` component wants some props that our test mounted `App` doesn't provide, and if it doesn't get them, the `/templates` route gets rendered.  so just mock the component implementation since `Editor`'s tests should confirm that it works correctly, and we can just make these `App` unit tests about whether `App`'s routing is set up as expected.

if we want to make sure `App` and `Editor` are wired up correctly, i'm guessing that it'd be both easier and more confidence inspiring to write a fuller integration test using puppeteer.

__tests__/components/App.test.js
```diff
@@ -4,14 +4,18 @@ import 'jsdom-global/register'
...
+jest.mock('../../src/components/editor/ImportResourceTemplate')
```
jmartin-sul: similarly, `ImportResourceTemplate` ultimately makes a call out to the server (via the `SinopiaResourceTemplates` component) to see if the server has new info for display.  these `App` unit tests are just about whether `ImportResourceTemplate` displays iff the user is logged in *and* the user goes to `/templates`.  we can write full on headless browser integration tests with a backend if we want to make sure that things work from UI to server and back.

__tests__/components/App.test.js
```diff
@@ -20,63 +24,70 @@ describe('<App />', () =>{
...
+    expect(wrapper.find(LoginPanel).length).toEqual(1)
   })
 
+  it('renders the Footer component', () => {
+    expect(wrapper.find(Footer).length).toBe(1)
+  })
 })
 
 describe("#routes", () => {
+  // pattern cribbed/adapted from https://stackoverflow.com/a/54807560
```
jmartin-sul: an alternative might be to use https://redux.js.org/api/createstore.  i just sort of happened on this solution first, and it works well and feels intelligible/reasonable.  totally open to using `createStore` or whatever.

__tests__/components/App.test.js
```diff
@@ -20,63 +24,70 @@ describe('<App />', () =>{
     expect(wrapper.find('div#app').length).toEqual(1)
   })
 
-  it('renders <Footer />', () => {
-    expect(wrapper.find(Footer).length).toBe(1)
+  it('renders the LoginPanel component', () => {
```
jmartin-sul: always present now

__tests__/CognitoUtils.test.js
```diff
@@ -49,8 +50,8 @@ describe('CognitoUtils', () => {
 
     it('resolves on successful authentication', () => {
       const mockSession = {}
-      const authUserSpy = jest.spyOn(user, 'authenticateUser').mockImplementation((_details, callback) => {
-        return callback.onSuccess(mockSession)
+      const authUserSpy = jest.spyOn(user, 'authenticateUser').mockImplementation((_details, callbacksObject) => {
```
jmartin-sul: https://github.com/LD4P/sinopia_editor/pull/570#discussion_r288252222

__tests__/CognitoUtils.test.js
```diff
@@ -144,5 +145,19 @@ describe('CognitoUtils', () => {
         })
         .catch(error => { throw error })
     })
+
+    it('bubbles up the error if obtaining a session fails', () => {
```
jmartin-sul: https://github.com/LD4P/sinopia_editor/pull/570#discussion_r288292938

__tests__/CognitoUtils.test.js
```diff
@@ -32,6 +32,7 @@ describe('CognitoUtils', () => {
       expect(user).toBeInstanceOf(CognitoUser)
       expect(user.username).toEqual(username)
       expect(user.pool).toBeInstanceOf(CognitoUserPool)
+      expect(user.pool).toEqual(CognitoUtils.cognitoUserPool())
```
jmartin-sul: https://github.com/LD4P/sinopia_editor/pull/570#discussion_r288258570

__tests__/components/App.test.js
```diff
@@ -4,14 +4,18 @@ import 'jsdom-global/register'
...
 import App from '../../src/components/App'
-import Config from '../../src/Config'
+import LoginPanel from '../../src/components/LoginPanel'
 import HomePage from '../../src/components/HomePage'
 import Editor from '../../src/components/editor/Editor'
 import Browse from '../../src/components/editor/Browse'
+import CanvasMenu from '../../src/components/CanvasMenu'
 import Footer from '../../src/components/Footer'
-import ImportResourceTemplate from "../../src/components/editor/ImportResourceTemplate";
+import ImportResourceTemplate from '../../src/components/editor/ImportResourceTemplate';
```
mjgiarlo: Trivial: You didn't add it, but we can ditch the line-ending semi-colon, which is inconsistent w/ the above `import`s.

__tests__/components/App.test.js
mjgiarlo: Agree that this makes sense, and that testing integration between these components best fits an integration test than a unit test.

__tests__/components/App.test.js
mjgiarlo: `/shrug` works for me

__tests__/components/App.test.js
```diff
@@ -20,63 +24,70 @@ describe('<App />', () =>{
...
-        </MemoryRouter>
+        <Provider store={unauthenticatedStoreFake}>
+          <MemoryRouter initialEntries={[path]}>
+            <App />
+          </MemoryRouter>
+        </Provider>
       )
 
-    it('root renders HomePage component', async () => {
+    it('root renders HomePage component', () => {
```
mjgiarlo: For the tests in this block, I might expect to see `test()` instead of `it()`, but since they're aliases... this is merely stylistic. (When I see `it()`, I want the test description to begin with a verb, but it makes sense for these tests to start with a noun to contextualize them.)

jmartin-sul: will keep the `it` style since that seems consistent with other tests, and will verbify and regularize the titles accordingly.

PR #588, opened 2019-05-30
Extract action creators and selectors so they can be tested.

### LINKED ISSUE #563: test new `mapStateToProps` and `mapDispatchToProps` functions from session management work as expected
i grouped these together as one ticket, since they're all small and similar.

- [x] test `mapStateToProps` and `mapDispatchToProps` in `LoginPanel`
- [x] test `Editor`'s `mapStateToProps`
- [x] unit test for `ImportResourceTemplate.mapStateToProps`

spawned from #528

jcoyne:
Fixes #563


src/selectors.js
```diff
@@ -0,0 +1,8 @@
+
```
ndushay: "selectors" seems an odd name choice, esp as it's overloaded per CSS selectors.  Maybe these could just be in CognitoUtils?

__tests__/actionCreators.test.js
```diff
@@ -0,0 +1,37 @@
...
+import { authenticationFailed, authenticationSucceeded, signedOut } from '../src/actionCreators';
+
+describe('authenticationFailed', () => {
+  const currentUser = { hello: 'world' }
+  const errInfoauthenticate = { foo: 'bar' }
+
+  const authResult = {
+    currentUser: currentUser,
+    authenticationError: errInfoauthenticate
+  };
```
ndushay: OMG - the dreaded semicolon!   (slightly prefer without them; not a biggie)

src/selectors.js
jcoyne: Selectors is a term of art in Redux: https://medium.com/@matthew.holman/what-is-a-redux-selector-a517acee1fe8

__tests__/actionCreators.test.js
jcoyne: Should we add that to es-lint?

jcoyne: I added a change to eslint for this: #590

jcoyne:
@jermnelson I've renamed that file as you suggested.

PR #592, opened 2019-05-30
589 handle subauthorities

### LINKED ISSUE #589: Ensure QA lookups refer to subauthorities and make correct call
Right now the InputLookupQA component uses the following call structure: "client.apis.SearchQuery .GET_searchAuthority" and passes the authority value in the subauthority field.  When calling the subauthority, the code should use "GET_searchSubauthority" instead.  The apidoc currently uses a space in the tag "Search Query", which cannot be converted to the format for "client.apis.SearchQuery" (spaces don't appear to work).  I'll be updating the apidoc to replace spaces in the tags.  Although this will affect how the documentation looks (since the tags are used as headings), this seems the appropriate method to use to get this code to work.  I'll be making a related issue on QA server to suggest updates to the apidoc there.

hudajkhan:
Fixes #589 

Updated apidoc to enable calling all methods (spaces needed to be removed from tags) using current pattern for calls.  If subauthority present, subauthority API call is made. Otherwise authority API is used.


src/components/editor/InputLookupQA.jsx
```diff
@@ -118,21 +119,43 @@ class InputLookupQA extends Component {
...
-                                })
-                                .catch(function(err) {
-                                  console.error("Error in executing lookup against source", err)
-                                  //return information along with the error in its own object
-                                  return { isError: true, errorObject: err }
-                                })
+                              //Subauthorities require a different API call than authorities so need to check if subauthority is available
+                              //The only difference between this call and the next one is the call to Get_searchSubauthority instead of 
+                              //Get_searchauthority.  If there is a way to somehow pass that in a variable name/dynamically, that would be better
+                              if(subauthority) {
```
mjgiarlo: @hudajkhan Re: your comment, how about:

```javascript
const actionFunction = subauthority === undefined ? client.apis.SearchQuery.GET_searchAuthority : client.apis.SearchQuery.GET_searchSubauthority

return actionFunction({
  q: query,
  vocab: authority,
  subauthority: subauthority,
  maxRecords: maxRecords,
  lang: language
})
// ...
```

src/components/editor/InputLookupQA.jsx
```diff
@@ -84,6 +84,7 @@ class InputLookupQA extends Component {
   render() {
     let authority, subauthority, language
     const lookupConfigs = this.props.lookupConfig
+    const maxRecords = 8
```
mjgiarlo: Can you add this to `src/Config.js` as `maxRecordsForQALookups`?

hudajkhan: Like this? 
static get maxRecordsForQALookups() {
	  return 8
 }

mjgiarlo: @hudajkhan 	How about:

```javascript
static get maxRecordsForQALookups() {
  return process.env.MAX_RECORDS_FOR_QA_LOOKUPS || 8
}
```

This lets us configure the default via environment variables. Thanks!

mjgiarlo:
@hudajkhan 	Can you also amend the test suites for `InputLookupQA` and `Config` (once changed) to reflect the changes you've made?

src/components/editor/InputLookupQA.jsx
hudajkhan: followed this pattern and now in code (tested and works)

src/components/editor/InputLookupQA.jsx
hudajkhan: Added this to Config.js

src/components/editor/InputLookupQA.jsx
hudajkhan: I'm going to resolve this conversation.  I'm taking the initiative or something but feel free to unresolve if it is not... resolved.

src/components/editor/InputLookupQA.jsx
hudajkhan: Also resolving because I followed instructions

hudajkhan:
Added a test for Config.  Mulling over test for input qa.  The changes to inputLookupQA were around identifying if a subauthority is passed from lookup config and then calling the right method from the apidoc.

src/components/editor/InputLookupQA.jsx
```diff
@@ -7,6 +7,7 @@ import { connect } from 'react-redux'
 import { getProperty } from '../../reducers/index'
 import { changeSelections } from '../../actions/index'
 import { booleanPropertyFromTemplate, defaultValuesFromPropertyTemplate } from '../../Utilities'
+import Config from '../../../src/Config'
```
mjgiarlo: This component is already in the `src/` directory, so I think the path should be `'../../Config'`

src/Config.js
```diff
@@ -63,6 +63,10 @@ class Config {
   static get cognitoTestUserPass() {
     return process.env.COGNITO_TEST_USER_PASS
   }
+  
+  static get maxRecordsForQALookups() {
+	  return process.env.MAX_RECORDS_FOR_QA_LOOKUPS || 8
```
mjgiarlo: Awesome, thanks!

src/components/editor/InputLookupQA.jsx
mjgiarlo: awesome!

hudajkhan:
Having an issue with the test I intended to evaluate the function for selecting the API.  I'm commenting it out for now until I figure out how to do this properly (I'm also running into interesting challenges on being able to run tests on any of my local environments but that is a different story).  thanks @jgreben for helping me step through some of it!

hudajkhan:
Down to 0 errors and 12 warnings but the cutoff is 11 warnings.

mjgiarlo:
Hi, @hudajkhan! Yesterday, I believe @jermnelson said that your Sinopia time was running low. If you don't have time to knock out the last warning, I'd be happy to pick up this branch, work on the build, rebase on `master`, and help finish. How would you like to proceed?

hudajkhan:
Hi @mjgiarlo ! If you could help me understand how to knock out that last warning (and maybe how to properly write the test), that would be great.   I can then update the code.  (fyi I slacked you too)

PR #593, opened 2019-05-30
Add the skeleton of an RDF graph builder

jcoyne:


src/graph_builder.js
```diff
@@ -0,0 +1,27 @@
+import Graph from './rdf/graph'
```
mjgiarlo: Can you rename the file to match the case of the class? so: `GraphBuilder.js`

mjgiarlo: Also, please add the copyright statement at the top of the file

src/graph_builder.js
```diff
@@ -0,0 +1,27 @@
...
+import Statement from './rdf/statement'
+/**
```
mjgiarlo: Would like a single blank line after imports. (Would be ideal if eslint did this for us. I will look into that.)

src/graph_builder.js
```diff
@@ -0,0 +1,27 @@
...
+    console.log(state)
+    this.state = state
+  }
+
+  /**
+   * @return {Graph} an RDF graph that represents the data in the state
+   */
+  get graph() {
+    const graph = new Graph()
+    for(var key in this.state) {
```
mjgiarlo: Prefer `const` to `var`. (Will be adding some eslint coverage of this.)

src/rdf/graph.js
```diff
@@ -0,0 +1,24 @@
+/*
+ * Represents an RDF graph
+ */
+export default class Graph {
+  constructor() {
+    this.statements = []
+  }
+
+  /**
+   * Return a serialization of the graph, maybe in n-triples?
```
mjgiarlo: We've got JSON-LD in Trellis. Not sure what is best for this context. @jermnelson ?

src/graph_builder.js
jcoyne: Doesn't it get reassigned each time through the loop?

src/rdf/graph.js
jcoyne: That's fine, but JSON-LD is a little harder to write an exporter for, so I'd defer that.

src/graph_builder.js
mjgiarlo: Er, yes it does. Sorry, meant `let`!

jcoyne:
@jermnelson No reason in particular.  I figured if we wanted to be purely functional we would have chosen elm or clojurescript.

jermnelson:
> @jermnelson No reason in particular. I figured if we wanted to be purely functional we would have chosen elm or clojurescript.

True, however I've been trying to move Sinopia to use more functional Javascript conventions for a couple of reasons. The first is that Redux in particular is easier to understand through a functional lens and switching between OO and functional code seems clunky to me. The second is that I'm concerned about the extra technical debt surrounding Javascript classes that increases the size and complexity of codebase without necessarily all the benefits of a true OO language.

jcoyne:
@jermnelson This part of the code has nothing to do with Redux.  All of React is written using classes:
```
class ResourceTemplate extends Component
```

Furthermore, I find that FP is very hard to use effectively without a strong type system. 

Can you explain what "technical debt surrounding Javascript classes" you are talking about?  They are a well supported and widely adopted feature of Javascript.

jermnelson:
> @jermnelson This part of the code has nothing to do with Redux. All of React is written using classes:
> 
> ```
> class ResourceTemplate extends Component
> ```
> 
> Furthermore, I find that FP is very hard to use effectively without a strong type system.
> 
> Can you explain what "technical debt surrounding Javascript classes" you are talking about? They are a well supported and widely adopted feature of Javascript.

You're right, having a strong type system would be great (my little bit of exploration with Haskell was enlightening) and that React is built with classes. My comment about technical debt is more about the  number of lines of code needed to support classes with their corresponding tests compared to a FP approach. I'll cease discussion on this topic as this isn't a blocker for the PR.

src/GraphBuilder.js
```diff
@@ -0,0 +1,33 @@
...
+   */
+  get graph() {
+    const dataset = rdf.dataset()
+
+    for (const key in this.state) {
+      /*
+       * TODO: this is just a simple statement insert to show this works. We'll
+       *       want to add more here.
+       */
+      dataset.add(rdf.quad(rdf.namedNode(''),
```
jgreben: Following on to the comments about the outdated `src/rdf/graph.js`... The base rdf library seems to use quads as the default, and then requires parsers to translate into JSON-LD or other serializations, so at this stage we don't really need to worry about storing JSON-LD because we are just displaying it in the modal. Also, quads will be one of the serializations that the catalogers will want to see. 👍

PR #590, opened 2019-05-30
Configure eslint to error on unnecessary semicolins

jcoyne:

PR #596, opened 2019-05-30
Make eslint more opinionated

mjgiarlo:
In the spirit of @jcoyne's [earlier PR](https://github.com/LD4P/sinopia_editor/pull/590), which added an eslint rule avoid extra semi-colons, I submit the following PR which makes eslint more opinionated about indentation, whitespace, variable declaration, and quotation marks.

# Why?

1. We already have eslint in our toolchain, and we lean on it both in CI and when doing local development;
1. Leaning on tools to handle style for us, like we do in our Ruby development where our practices have had more time to mature, means less time spent on nitpicky concerns in code review and less back-and-forth about our style conventions;
1. Using community-based rule sets mean our code conventions are better aligned with open-source peers';
1. Using a tool to handle style, particularly when the `--fix` option can be deployed, means less time guessing about how to style our code when working on PRs;
1. `eslint` shows us that the editor already has a very high degree of inconsistency.

# OK, but why now?

1. We are doubling our engineering capacity, so it's timely to solidify our style conventions and rules now before further drift/divergence happens, especially for developers who are less experienced in the Javascript realm;
1. It reflects that our Javascript dev practices are maturing, in much the way that our Ruby practices matured several years ago;
1. We are in the home stretch of the project, and we know we'll need to maintain this code once the current work cycle ends, so having a more consistent codebase will be easier to pick up later.


.eslintrc.js
```diff
@@ -42,24 +43,96 @@ module.exports = {
     "es6": true
   },
   rules: {
+    "node/no-unsupported-features/es-syntax": "off",
```
mjgiarlo: I moved this up from the overrides below. Since this rule applies to the entire codebase, it didn't make sense to me to have it expressed as an override.

.eslintrc.js
```diff
@@ -15,7 +15,8 @@ module.exports = {
     "plugin:import/errors",
     "plugin:import/warnings",
     "plugin:jsx-a11y/recommended",
-    "plugin:security/recommended"
+    "plugin:security/recommended",
+    "airbnb-base/whitespace"
```
mjgiarlo: Use Airbnb's widely used ruleset, but only the whitespace stuff (bc our other rules may not mesh well with Airbnb's, and I don't want to unilaterally change all the rules already declared below).

jermnelson:
Looks good to me and makes sense to push formatting normalization to the tool chain instead of relying on code reviews for noticing and fixing. I would be open to a larger DEV discussion of what other Airbnb JS styles we might want to adopt.

PR #612, opened 2019-06-03
Adds save and cancel buttons to RDF Modal

jgreben:
Also adds skeleton server function to handle save


jgreben:
> Quick question, we should see the **Cancel** and **Save and Publish** button when opening the the RDF Modal correct? I've tried building using `npm run dev-build` and then running the server and I'm still seeing the old RDFModal:

That's strange. This is what I see:

Screen Shot 2019-06-04 at 9 20 48 AM

Screen Shot 2019-06-04 at 9 21 10 AM

PR #613, opened 2019-06-03
Fix issue when clicking templates that reference missing templates

### LINKED ISSUE #602: error when you try to open a nested resource template that has another resource template nested within it
example: 
sinopia:resourceTemplate:bf2:Monograph:Work
Under "Title information"
Click plus sign by "Note on Title"
Sinopia window goes blank

or under Creator of Work,
Primary Contribution,
click + next to "Primary Contributor"

I checked the Resource Template by opening it in the Profile Editor and nothing looks amiss with the set up.

mjgiarlo:
Fixes #602

Includes:
* Make `ImportFileZone` component validate that all referenced identifiers in an imported resource template are either 1) defined within the resource template, or 2) available via the Sinopia server. Else bubble up an error message about using an undefined resource tempalte.
* Fix assumption in reducer that all property templates have `.valueConstraint.defaults` defined; some resource templates lack a `valueConstraint`
* Add new Sinopia server function to check if an identified RT exists
* Add `jsonpath-plus` dependency to quickly gather up nodes in a JSON document
* Allow `arrow-body-style` ESLint violations in tests to permit Jest idiomatic style

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -137,6 +137,13 @@ module.exports = {
   },
   overrides: [
     {
+      // Allow tests to include block statements (idiomatic Jest style)
+      "files": ["__tests__/**"],
+      "rules": {
+        "arrow-body-style": "off"
+      }
+    },
+    {
       // Allow integration tests to use `page` global variable defined by puppeteer
       "files": ["__tests__/integration/**"],
       "rules": {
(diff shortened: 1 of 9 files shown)


mjgiarlo:
Going to have another crack at this PR.

PR #615, opened 2019-06-04
Allows re-editing of literals.

### LINKED ISSUE #540: Handle the re-editing of existing input values and preserve order of multiple input values
The LOC version of the editor has functionality where you can edit a pre-existing input value. It pops it up in a modal, but we want to probably do it differently by pushing the value back into the input field, preserving the order of multiple input values.

justinlittman:
closes #540

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -164,6 +164,13 @@ module.exports = {
         "react/display-name": "warn",
         "react/prop-types": "warn"
       }
-    }
+    },
+    {
+      // Remove limits on number of lines in test files
+      "files": ["__tests__/**"],
+      "rules": {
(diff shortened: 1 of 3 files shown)


.eslintrc.js
```diff
@@ -164,6 +164,13 @@ module.exports = {
         "react/display-name": "warn",
         "react/prop-types": "warn"
       }
-    }
+    },
+    {
+      // Remove limits on number of lines in test files
+      "files": ["__tests__/**"],
+      "rules": {
+        "max-lines": "off"
```
jermnelson: Any particular reason you decided to turn of the `max-lines` instead of upping the limit to say 500 lines?

Issue #614, opened 2019-06-03
Child resource templates should have rdfClass in the state


jcoyne:
Currently if you select `resourceTemplate:bf2:Monograph:Work`.  Then the title sub-template does not have a rdrClass in the Redux state.

jcoyne:
The problem with this is that, the resourceURI doesn't seem to be present in the propertyTemplate.

PR #618, opened 2019-06-04
Omit type assertion since the permission template doesn't support it

jcoyne:
This removes an invalid Literal set as the rdf:type

Ref #614


justinlittman:
Can you provide some more context for this ticket? I'm not understanding the motivation for the change.

jcoyne:
This was just setting a dummy value for rdf:type.  Rather than a URI it was setting a literal.  A literal is not a valid range for rdf:type.  This removes that.

PR #619, opened 2019-06-04
Add rdf:type for child resource templates

### LINKED ISSUE #614: Child resource templates should have rdfClass in the state

jcoyne:
Fixes #614

### DIFF
--- a/src/GraphBuilder.js
+++ b/src/GraphBuilder.js
@@ -24,7 +24,7 @@ export default class GraphBuilder {
 
     // Is there ever more than one base node?
     Object.keys(this.state).forEach((key) => {
-      this.buildTriplesForNode(baseURI, rdf.namedNode(this.state[key].rdfClass), this.getPredicateList(key))
+      this.buildTriplesForNode(baseURI, rdf.namedNode(this.state[key].rdfClass), this.getPredicateList(key, this.state))
     })
     return this.dataset
   }
@@ -32,14 +32,15 @@ export default class GraphBuilder {
   /**
    * Filter out the non-predicate values (e.g. rdfClass)
    * @param {string} key
-   * @return {Array.<Object>} the predicate InputListLOC
+   * @param {Object} tree
+   * @return {Object} the predicate InputListLOC
    */
-  getPredicateList(key) {
+  getPredicateList(key, tree) {
     const predicateList = {}
 
-    Object.keys(this.state[key]).forEach((predicateKey) => {
+    Object.keys(tree[key]).forEach((predicateKey) => {
       if (predicateKey != 'rdfClass') {
-        predicateList[predicateKey] = this.state[key][predicateKey]
+        predicateList[predicateKey] = tree[key][predicateKey]
       }
     })
     return predicateList
@@ -49,15 +50,17 @@ export default class GraphBuilder {
    * @param {rdf.Term} baseURI
    * @param {Object} value looks something like this:
    *   { 'resourceTemplate:bf2:Note':
-   *     { '67Bm64T0p2s': { 'http://www.w3.org/2000/01/rdf-schema#label': [Object] },
+   *     { 'rdfClass: 'http://id.loc.gov/ontologies/bibframe/Instance',
+   * '67Bm64T0p2s': { 'http://www.w3.org/2000/01/rdf-schema#label': [Object] },
    *       '555m64M0f3z': { 'http://www.w3.org/2000/01/rdf-schema#label': [Object] } } }
    */
   buildTriplesForNestedObject(baseURI, value) {
     // Is there ever more than one base node?
     for (const key in value) {
       Object.keys(value[key]).forEach((predicateListKey) => {
-        // we don't have a resourceURI to send as the second argument in the propertyTemplates yet.
-        this.buildTriplesForNode(baseURI, null, value[key][predicateListKey])
+        if (predicateListKey !== 'rdfClass') {
+          this.buildTriplesForNode(baseURI, rdf.namedNode(value[key].rdfClass), value[key][predicateListKey])
+        }
       })
     }
   }
--- a/src/components/editor/ResourceTemplateForm.jsx
+++ b/src/components/editor/ResourceTemplateForm.jsx
@@ -57,7 +57,7 @@ export class ResourceTemplateForm extends Component {
     if (rtIds === null || rtIds === undefined) {
       return rtProperties
     }
-    rtIds.map((rtId, i) => {
+    rtIds.forEach((rtId, i) => {
       const rt = this.rtForPt(rtId)
 
       if (rt !== undefined) {
--- a/__tests__/reducers/inputs.test.js
+++ b/__tests__/reducers/inputs.test.js
@@ -1,4 +1,5 @@
 // Copyright 2018 Stanford University see LICENSE for license
+/* eslint max-lines: 0 */ // --> OFF
 
 import {
   removeAllContent, removeMyItem, setMyItems, setMySelections,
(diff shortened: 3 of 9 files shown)


src/GraphBuilder.js
```diff
@@ -49,15 +50,17 @@ export default class GraphBuilder {
...
+   * '67Bm64T0p2s': { 'http://www.w3.org/2000/01/rdf-schema#label': [Object] },
    *       '555m64M0f3z': { 'http://www.w3.org/2000/01/rdf-schema#label': [Object] } } }
    */
   buildTriplesForNestedObject(baseURI, value) {
     // Is there ever more than one base node?
     for (const key in value) {
       Object.keys(value[key]).forEach((predicateListKey) => {
-        // we don't have a resourceURI to send as the second argument in the propertyTemplates yet.
-        this.buildTriplesForNode(baseURI, null, value[key][predicateListKey])
+        if (predicateListKey !== 'rdfClass') {
```
justinlittman: `rdfClass` is being filtered out in 2 locations. Are both necessary? And if both necessary, can filtering be moved into separate method to avoid repeating?

jcoyne: I believe they are necessary, because the structure of the redux state is not consistent between the ResourceTemplate and PropertyTemplate levels.

jcoyne: See here: https://github.com/LD4P/sinopia_editor/pull/619/files#diff-115c58c1c8d1c22ebd6a017e5b91efe3R78. that key is not present at the ResourceTemplate level.  If they had similar structures, then we could use the same method.

__tests__/components/editor/ImportFileZone.test.js
```diff
@@ -19,7 +19,10 @@ describe('<ImportFileZone />', () => {
   describe('schema valid', () => {
     describe('schema url in JSON', () => {
       const wrapper = shallow(<ImportFileZone />)
-      let schemaUrl; let
+
+      let schemaUrl
+
+      let
```
jgreben: You might want to rebase off of the PR @mjgiarlo submitted (that should get merged first). it fixes this weirdness of the `let`s.

src/components/editor/PropertyComponent.jsx
```diff
@@ -27,7 +27,9 @@ export class PropertyComponent extends Component {
   }
 
   inputComponentType = (property) => {
-    let config; let
+    let config
+
+    let
```
jgreben: Not sure if this shows up as being on separate lines in your editor...

src/components/editor/PropertyTemplateOutline.jsx
```diff
@@ -82,8 +82,13 @@ export class PropertyTemplateOutline extends Component {
...
   }
 
+  /*
+   * This is fired when the expand/collapse button is clicked
+   */
   addPropertyTypeRows = (property) => {
-    let existingJsx; let
+    let existingJsx
+
+    let
```
jgreben: ditto ☝️

src/GraphBuilder.js
jermnelson: @jcoyne and @justinlittman, the current Redux hierarchical structure has been evolving over the project and how best to represent Resource Templates and the relationships with their propertyTemplates including referencing other Resource Templates. Property Templates just represent a triple to the Resource Template and don't have a `rdf:Class`.

src/components/editor/PropertyTemplateOutline.jsx
```diff
@@ -113,7 +121,7 @@ export class PropertyTemplateOutline extends Component {
           addButtonDisabled={this.props.addButtonDisabled}
           propertyTemplate={property}>
           {propertyJsx}
-        </PropertyTypeRow>,
+]        </PropertyTypeRow>,
```
jgreben: This `]` looks out of place here.

src/components/editor/PropertyTypeRow.jsx
```diff
@@ -4,6 +4,9 @@ import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 import shortid from 'shortid'
 
+/**
+ * Represents a single input field. E.g. "Prefered title for work"
```
jgreben: 👍 to adding the comment

src/components/editor/ResourceTemplateForm.jsx
```diff
@@ -63,7 +63,7 @@ export class ResourceTemplateForm extends Component {
     if (rtIds === null || rtIds === undefined) {
       return rtProperties
     }
-    rtIds.map((rtId, i) => {
+    rtIds.forEach((rtId, i) => {
```
jgreben: We had the `map` vs. `forEach` debate before the rest of the infa-team joined. We never collectively came to an agreement about which was preferred. Supposedly `map` is faster ¯\_(ツ)_/¯

src/components/editor/PropertyTemplateOutline.jsx
jcoyne: This stuff was from doing `eslint --fix`. I don't know why it fixed it this way.  Bug in eslint?

src/components/editor/ResourceTemplateForm.jsx
jcoyne: My understanding is that `map` is for returning a value, and `forEach` is just for looping over the elements.

jermnelson: Yes, after checking [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) on `map`, I'm softening my stance on the `forEach` vs. `map`. Good change.

justinlittman:
In local testing, I'm getting an error.

Steps to reproduce:
1. Fill out a Bibframe Instance as shown below.
Screen Shot 2019-06-05 at 4 56 14 PM
2. Hit enter.

Results in:
Screen Shot 2019-06-05 at 4 59 06 PM

jcoyne:
@justinlittman yeah, the problem here is that the state is all messed up.

I'm expecting a structure like:
```
   { 'resourceTemplate:bf2:Note':
      { '67Bm64T0p2s': { 'http://www.w3.org/2000/01/rdf-schema#label': [Object] },
   }
```

 but in this case I get a structure like:
```
    { 'resourceTemplate:bf2:Monograph:Work':
       { FrUjeDCcku: { 'resourceTemplate:bf2:WorkTitle': [Object] } } }
```

jcoyne:
@justinlittman It looks like that bug appears in master too. I opened https://github.com/LD4P/sinopia_editor/issues/648 for it.

jermnelson:
Hi @jcoyne and @justinlittman I'll add my comments to the #648 bug report

jermnelson:
My first pass at supporting repeatable=true for Resource Template is causing this mismatch in the state expectations through the prop `reduxPath` not being set correctly. Using this example, initially the `reduxPath` for the Preferred title's PropertyTemplateOutline reduxPath is set incorrectlly:
```javascript
["resourceTemplate:bf2:Monograph:Instance", "http://id.loc.gov/ontologies/bibframe/instanceOf", 
"Ok71o5G-Ni",  "resourceTemplate:bf2:Monograph:Work", "FrUjeDCcku"]
```
The last element should be fixed to use the propertyTemplate `http://id.loc.gov/ontologies/bibframe/title` as the key instead of "FrUjeDCcku".  

We could roll this fix into #649 and get this PR merged, @justinlittman?

PR #620, opened 2019-06-04
Create a modal to get user input for choosing a sinopia group

### LINKED ISSUE #597: Choose Group Modal when saving RDF
When the user submits an entity's RDF to save, they should be prompted to select which Group from a drop-down list that they want to save the entity into before the payload is sent to the server. 

Here is @astridu design:
Screen Shot 2019-05-30 at 4 43 36 PM

jgreben:
Fixes #597 

Screen Shot 2019-06-04 at 1 34 03 PM
Screen Shot 2019-06-04 at 1 34 15 PM
Screen Shot 2019-06-04 at 1 34 36 PM
Screen Shot 2019-06-04 at 1 34 54 PM

Granted, on the original design, the Preview RDF was a full page instead of a modal.

### DIFF
--- a/src/Config.js
+++ b/src/Config.js
@@ -71,6 +71,22 @@ class Config {
   static get maxRecordsForQALookups() {
     return process.env.MAX_RECORDS_FOR_QA_LOOKUPS || 8
   }
+
+  static get groupsInSinopia() {
+    return [
+      ['ld4p', 'LD4P'],
+      ['pcc', 'PCC'],
+      ['cub', 'University of Colorado Boulder'],
+      ['cornell', 'Cornell University'],
+      ['harvard', 'Harvard University'],
+      ['nlm', 'National Library of Medicine'],
+      ['stanford', 'Stanford University'],
+      ['ucsd', 'University of California, San Diego'],
+      ['penn', 'University of Pennsylvania'],
+      ['hrc', 'Harry Ransom Center, University of Texas at Austin'],
+      ['wau', 'University of Washington'],
+    ]
+  }
 }
 
 export default Config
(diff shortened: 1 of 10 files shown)


jcoyne:
Are you following some design for this?  I'm curious as to why you chose to combine this with the RDF preview.

jgreben:
> Are you following some design for this? I'm curious as to why you chose to combine this with the RDF preview.

@jcoyne https://github.com/LD4P/sinopia_editor/issues/597 shows this as a modal on top of the preview.

jermnelson:
> > Are you following some design for this? I'm curious as to why you chose to combine this with the RDF preview.
> 
> @jcoyne #597 shows this as a modal on top of the preview.

I think moving the **Cancel** and **Save and Publish** button back down to the Editor Tab instead of being part of the **Preview RDF** modal would be a better user experience, maybe @astridu can weigh in and I'll make a ticket?

src/Config.js
```diff
@@ -71,6 +71,22 @@ class Config {
   static get maxRecordsForQALookups() {
     return process.env.MAX_RECORDS_FOR_QA_LOOKUPS || 8
   }
+
```
jermnelson: For the MVP I think it is fine to hard-code these values; we'll need to add all 17 members of the cohort in this list. In future if we have 100s of different institutions, this method seems not scalable. Not sure what the best method for this, maybe an Elasticsearch query when we have the RDF data?  Not blocking.

jgreben: Maybe a trellis api call to `/repository` using https://github.com/LD4P/sinopia_editor/blob/ff314f1b36015b7b4a927e7eabb1e3e914c321e5/src/sinopiaServer.js#L33 ? Initially adding the groups would be handled via `sinopia_acl`

PR #621, opened 2019-06-04
Adds load timestamp to graph for retrieving resource template by mome…

### LINKED ISSUE #75: refactor: dry up tests by making functions/matchers avail
currently, multiple tests have functions such as `expect_value_in_selector_textContent` added.  It would be nice to dry this up.

possible approaches:

- from https://github.com/jest-community/jest-extended#setup
```
If you are already using another test framework, like jest-chain, then you should create a test setup file and require each of the frameworks you are using.
```
I don't know if we can use node's `require` to load a simple file with functions.
- create a node module, akin to jest-extended (https://github.com/jest-community/jest-extended)

#Requires #74

justinlittman:
…nto.

closes #75

### DIFF
--- a/src/GraphBuilder.js
+++ b/src/GraphBuilder.js
@@ -24,8 +24,12 @@ export default class GraphBuilder {
       // If the resourceURI is not in the state, then this is an unsaved resource and we want a relative URI to use as the base
       const resourceURI = rdf.namedNode(this.state[key].resourceURI || '')
 
+      // Assuming that loadTimestamp will be close enough for included template
+      // and so only will record for main template.
+      this.buildTriplesForTemplate(resourceURI, key, this.state[key].loadTimestamp)
       this.buildTriplesForNode(resourceURI, this.state[key].rdfClass, this.getPredicateList(this.state[key]))
     })
+
     return this.dataset
   }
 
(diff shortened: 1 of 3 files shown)


src/reducers/index.js
```diff
@@ -112,7 +112,9 @@ export const setResourceTemplate = (state, action) => {
     output = refreshResourceTemplate(output, propertyAction)
   })
 
+  console.log(action.payload.loadTimestamp)
```
justinlittman: Cruft.

src/GraphBuilder.js
```diff
@@ -22,10 +22,13 @@ export default class GraphBuilder {
     // A relative URI to use as the base
     const baseURI = rdf.namedNode('')
 
-    // Is there ever more than one base node?
-    Object.keys(this.state).forEach((key) => {
-      this.buildTriplesForNode(baseURI, rdf.namedNode(this.state[key].rdfClass), this.getPredicateList(key))
+    Object.keys(this.state).forEach((rtId) => {
+      // Assuming that loadTimestamp will be close enough for included template
+      // and so only will record for main template.
```
jermnelson: Not blocking but what would it take to record this timestamp for all of the embedded resources in the  main template's propertyTemplates?

justinlittman: Yes, but going to hold pending #619.

justinlittman:
Closing because going to use resource create timestamp from Trellis.

PR #628, opened 2019-06-05
Split property component tree into its own directory

jcoyne:
This makes it easier to understand what components are related to properties and which operate at a higher level.


jermnelson:
Could you provide more context and why you're suggesting this change in the PR's description?

.eslintrc.js
```diff
@@ -29,6 +29,7 @@ module.exports = {
   settings: {
     "import/resolver": {
       "node": {
+        "paths": ['src'],
```
mjgiarlo: Can you explain why this branch needs this change?

mjgiarlo: Nvm, I think I get it now. It provides a base resolution path so we can get away from the `..` nonsense. Cool! Good change.

src/components/editor/property/InputLang.jsx
```diff
@@ -4,7 +4,7 @@ import React, { Component } from 'react'
 import { Typeahead } from 'react-bootstrap-typeahead'
 import PropTypes from 'prop-types'
 import { connect } from 'react-redux'
-import { setLang } from '../../actions/index'
+import { setLang } from '../../../actions/index'
```
mjgiarlo: Could the path here not be `'components/actions/index'`? I assumed that's what the changes to `package.json` and `.eslintrc.js` would enable. No?

mjgiarlo: Or is it more that you're not sanguine about committing to nuking each and every `'../'` import in the codebase? That'd be reasonable. Maybe we could get eslint to auto-fix that for us?

.eslintrc.js
jcoyne: yes, that's right.

src/components/editor/property/InputLang.jsx
jcoyne: Yes, I was trying only to change the `__test__` imports in this PR, but I think we could change the imports in `src` too.

jcoyne:
@jermnelson conflict resolved.

PR #640, opened 2019-06-05
[Refactor] rename getResourceTemplatePromise to resolveResourceTempla…

jcoyne:
…tePromise

This makes the intention of the method a bit clearer.  I also removed resourceTemplatePromise() as it
added unnecessary indirection

### DIFF
--- a/src/components/editor/ResourceTemplate.jsx
+++ b/src/components/editor/ResourceTemplate.jsx
@@ -20,11 +20,12 @@ class ResourceTemplate extends Component {
     }
   }
 
+  // Called immediately after the component is mounted
   componentDidMount() {
-    this.getResourceTemplatePromise(this.resourceTemplatePromise())
+    this.resolveResourceTemplatePromise(getResourceTemplate(this.props.resourceTemplateId))
   }
 
-  getResourceTemplatePromise = (promise) => {
+  resolveResourceTemplatePromise = (promise) => {
     promise.then((responseAndBody) => {
       this.setState({ rtData: responseAndBody.response.body })
       this.props.handleResourceTemplate(this.state.rtData)
@@ -32,8 +33,6 @@ class ResourceTemplate extends Component {
       this.setState({ error })
     })
   }
-
-  resourceTemplatePromise = () => getResourceTemplate(this.props.resourceTemplateId)
 
   renderRtData = () => (
     <div className="ResourceTemplate">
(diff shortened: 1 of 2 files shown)


src/components/editor/ResourceTemplate.jsx
```diff
@@ -20,11 +20,12 @@ class ResourceTemplate extends Component {
     }
   }
 
+  // Called immediately after the component is mounted
   componentDidMount() {
-    this.getResourceTemplatePromise(this.resourceTemplatePromise())
+    this.resolveResourceTemplatePromise(getResourceTemplate(this.props.resourceTemplateId))
```
mjgiarlo: Do you think this function is adding enough to be worth remaining on its own? I wonder about that vs. inlining it as follows:

```javascript
  // Called immediately after the component is mounted
  componentDidMount() {
    getResourceTemplate(this.props.resourceTemplateId)
      .then((response_and_body) => {
      this.setState({ rtData: response_and_body.response.body })
      this.props.handleResourceTemplate(this.state.rtData)
    }).catch((error) => {
      this.setState({ error })
    })
  }
```

jcoyne: I tried that, but it broke the testing that calls `resolveResourceTemplatePromise`

mjgiarlo: My other question is about a pattern sometimes used in the editor where we call `this.setState()` to set a property and then immediately after expect it to be updated in `this.state`. This seems to run counter to the [React documentation](https://reactjs.org/docs/react-component.html#setstate):

> ... setState() does not always immediately update the component. It may batch or defer the update until later. This makes reading this.state right after calling setState() a potential pitfall. ... 

Bearing that in mind, how about this:

```javascript
  // Called immediately after the component is mounted
  componentDidMount() {
    getResourceTemplate(this.props.resourceTemplateId)
      .then((response_and_body) => {
	    const templateBody = response_and_body.response.body
        this.setState({ rtData: templateBody })
        this.props.handleResourceTemplate(templateBody)
      })
      .catch((error) => {
        this.setState({ error })
      })
  }
```

jgreben: 👍

jermnelson: @mjgiarlo are you okay with merging this PR?

jgreben: Also (@mjgiarlo) The redux state is getting set with the template data as expected the way it is currently, and we discussed refactoring this kind of component state setting and moving the source of RT data to redux, so this component would be refactored away from the way it currently works.

mjgiarlo: OK, @jgreben.

@jermnelson 	As long as someone approves it, most definitely. Thanks for asking!

PR #642, opened 2019-06-05
Fixes language input (you know, so that you can input a language) and…

### LINKED ISSUE #275: Setting Language for Input type literal: Hitting cancel should empty out the current redux store pertaining to that field.
- Could be related to #256
- If you want to cancel setting a language, make sure Cancel will remove or reset the redux store for that Input Literal field.

- Currently, when you add a field to a literal, then click on the language to update it, but hit cancel, the language still gets se in redux

### LINKED ISSUE #529: Fix displayLang in InputLiteral
The `InputLiteral` component's `displayLang` needs to update to the specific item object in Redux state.,

justinlittman:
… adds lang to state.

closes #529
closes #275


src/components/editor/InputLang.jsx
```diff
@@ -15,47 +14,38 @@ class InputLang extends Component {
...
-        })
-
-        return { id: uri, uri, label }
-      } catch (error) {
-        // ignore
-      }
-    }
+    return json.reduce((result, item) => {
+      // Object.getOwnPropertyDescriptor is necessary to handle the @
+      const uri = Object.getOwnPropertyDescriptor(item, '@id').value
```
mjgiarlo: Can also use the safe navigation operator here, which we use in a couple other spots in the editor:

```javascript
const uri = item?.['@id'].value
```

I am not sure which is more idiomatic or more readable.

mjgiarlo: Or `item['@id']` if you don't need safe navigation.

src/components/editor/InputLang.jsx
```diff
@@ -15,47 +14,38 @@ class InputLang extends Component {
...
+      const labelArrayDescr = Object.getOwnPropertyDescriptor(item, 'http://www.loc.gov/mads/rdf/v1#authoritativeLabel')
+      // There are some odd entries, so ignoring if don't have labels.
+      if (!labelArrayDescr) return result
+      const labelArray = labelArrayDescr.value
+
+      // Looking for English label
+      let label = null
+      labelArray.forEach((langItem) => {
+        if (langItem['@language'] == 'en') {
```
mjgiarlo: It would be better to get eslint to enforce this for us, but generally I think we should prefer `===` over `==` since the latter does type conversions. (Which can be useful if we want type conversions...)

src/reducers/inputs.js
```diff
@@ -42,6 +42,31 @@ export const setMyItems = (state, action) => {
...
+  const reduxPath = action.payload.reduxPath
+  let level = 0
+
+  reduxPath.reduce((obj, key) => {
+    level++
+    if (level === reduxPath.length) {
+      if ((key in obj) != true) {
+        obj[key] = { items: [] }
+      }
+      obj[key].items.forEach(item => {
```
mjgiarlo: I wonder if this should use `map()` instead since you're modifying values in the array?

src/reducers/inputs.js
```diff
@@ -42,6 +42,31 @@ export const setMyItems = (state, action) => {
...
+        if (item.id == action.payload.id) {
```
mjgiarlo: Is this only going to match one item in the array or multiple? (Not sure what this data tends to look like.) If only the first, we may want a different iterator function above.

src/reducers/inputs.js
mjgiarlo: On second thought, no, I don't think `map()` is a good fit here since you're looking to mutate a single property in a single object in the array (I think?).

src/components/editor/InputLang.jsx
justinlittman: Agree with you that these are more idiomatic. However, for some unknown reason they don't work with `@id`.

src/reducers/inputs.js
justinlittman: Yes, only one. Open to suggestions.

src/components/editor/InputLang.jsx
mjgiarlo: @justinlittman That's surprising! What kind of thing is `item` here? I created a dummy object in the console that contained a key of `@id` and it worked fine. So clearly you're not dealing with a dummy object here. :)

src/reducers/inputs.js
mjgiarlo: OK, cool. How about:

```javascript
const payloadItem = obj[key].	items.find(item => item.id === action.payload.id)
payloadItem.lang = { items: action.payload.items }
```

justinlittman: Thanks for the suggestion. Updated.

PR #647, opened 2019-06-06
Remove properties that do not conform to the schema

jcoyne:


static/spoofedFilesFromServer/fromSinopiaServer/resourceTemplates/MonographInstance.json
```diff
@@ -140,8 +140,6 @@
           }
         ],
         "editable": "true",
-        "languageLabel": "LANGUAGE LABEL",
```
mjgiarlo: I have to defer to @jermnelson on this PR. It's not clear to me what schema versions we support currently, and which schema this RT was created with (it's missing `schema:` declarations). And this isn't a Profile, I don't think, so we can't even import via the editor.

jermnelson: As this spoofed resource template predates much of the JSON validation work, I think removing these properties is fine.

PR #652, opened 2019-06-06
Only generate the RDF when the button is pressed to preview

jcoyne:
Also changed the UI a bit.  Before:

Screen Shot 2019-06-06 at 10 05 56 AM

After:

Screen Shot 2019-06-06 at 10 20 57 AM


jcoyne:
@justinlittman Yeah, found that issue. I think I switched a function to use `{ }` when I had debug and I forgot to revert that when I removed the debug.

PR #663, opened 2019-06-06
Clean up ESLINT rules and update code for consistency

aaron-collier:
This cleans up several of the eslint ignores that lead to inconsistent code and pushes in more idiomatic use of javascript.

There is more work to do, but in the interest of avoiding excessive need for rebasing, removing draft now.


mjgiarlo:
@aaron-collier digging this!

server.js
```diff
@@ -30,13 +30,13 @@ app.all('/profile-edit/server/whichrt', (req, res) => {
   const reqUri = req.query.uri
 
   if (reqUri !== null) {
-    // console.debug(`DEBUG: got server/whichrt for ${reqUri}.`)
+    // Console.debug(`DEBUG: got server/whichrt for ${reqUri}.`)
```
jermnelson: We should just delete this line.

aaron-collier: Yeah for sure, i'm continuing the clean up ;-)

PR #671, opened 2019-06-07
Simple integration test for unauthenticated users

aaron-collier:
This is just a simple (and possibly not useful) pass at an integration test to verify that a logged out user cannot access the resource templates that require authentication. This is part of #421 which is has some testing already complete in other PRs.

### DIFF
--- a/__tests__/integration/unauthenticatedNavigation.test.js
+++ b/__tests__/integration/unauthenticatedNavigation.test.js
@@ -7,8 +7,10 @@ describe('When an unauthenticated user tries to access resource templates', () =
     jest.setTimeout(60000) // this seems to take around 10s in practice, but be generous, just in case
     await page.goto('http://127.0.0.1:8888/templates')
 
-    // Attempt to logout
-    // This try block is necessary to avoid failing after logout
+    /*
+     * Attempt to logout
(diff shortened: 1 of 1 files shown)


__tests__/integration/unauthenticatedNavigation.test.js
```diff
@@ -0,0 +1,25 @@
...
+    try {
+      await page.waitForSelector('button.signout-btn')
+      await page.click('button.signout-btn')
+    } catch (error) {
+      console.info(error)
+    }
+  })
+
+  it('does not display "Available Resource Templates in Sinopia"', async () => {
+    await expect(page).not.toMatch('Available Resource Templates in Sinopia')
```
justinlittman: What does it display?

aaron-collier: The existing page. I think the login form displays, but if it also displays for a logged in user (I'll verify) we can't really test for that here.

aaron-collier: Added a test for the login form.

jermnelson: Is this resolved @justinlittman?

jgreben:
LGTM!

PR #676, opened 2019-06-10
Change default display label for Typeahead component

jgreben:
The default Typeahead component `emptyLabel` is "no matches found." Sometimes the list lookup to LOC takes a couple of seconds, so changing the label to something more user friendly might help avoid confusion.


jgreben:
Per @justinlittman improve by making list retrieve when modal opens instead of onFocus

PR #684, opened 2019-06-11
Fix undefined label in the InputLookupQA component

### LINKED ISSUE #660: clicking on some nested resource templates still causing blank screen bug
related to #602 

with this template on stage: 

Name: RT RDA Manifestation for monographs
ID: rt:rda:manifestation:monograph

in the 'Administrative Metadata for Manifestation' section, try to access 'record status' 

blank screen

(I'm sorry I don't have ready access to the RT here, can you get from server?)

jgreben:
Fixes #660 

Adds some new spoofed resource templates for testing this fix.

If the defaultLiteral value is undefined, use the defaultURI instead.

Screen Shot 2019-06-11 at 11 44 46 AM
Screen Shot 2019-06-11 at 11 43 21 AM

### DIFF
--- a/__tests__/Utilities.test.js
+++ b/__tests__/Utilities.test.js
@@ -162,7 +162,7 @@ describe('Utilities', () => {
     })
 
     it('returns an array with an object otherwise', () => {
-      const propertyTemplate = {
+      const propertyTemplateWithDefaultsAndLabel = {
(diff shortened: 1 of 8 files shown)


jermnelson:
LGTM; I don't see anything at this point that would prevent future work.

PR #685, opened 2019-06-11
Fixed add buttons.

### LINKED ISSUE #675: Adding new Resource Template creates PropertyTemplateOutline instead of PropertyTypeRow
For a `PropertyResourceTemplate.PropertyTemplateOutline.PropertyTypeRow` with a repeatable Resource Template, clicking the Add button creates a new `PropertyTemplateOutline` instead of a `PropertyTypeRow`.

In the example below, clicking on the **Add** button in the blue circle create a `PropertyTemplateOutline` outlined in the red rect, should be a `PropertyTypeRow` outlined in the green rec.t 
bug-propTypeOutline

justinlittman:
closes #675

### DIFF
--- a/__tests__/GraphBuilder.test.js
+++ b/__tests__/GraphBuilder.test.js
@@ -41,7 +41,19 @@ describe('GraphBuilder', () => {
                   },
                   {
                     content: 'Sparkly',
-                    id: '3TzRpgv65',
+                    id: '5TzRpgv72',
+                  },
+                ],
+              },
+            },
+          },
+          gdndfCHlqQ4z: {
+            'resourceTemplate:bf2:Note': {
+              'http://www.w3.org/2000/01/rdf-schema#label': {
+                items: [
+                  {
+                    content: 'Shiney',
+                    id: '4dzRpgv42',
                   },
                 ],
               },
@@ -71,12 +83,23 @@ describe('GraphBuilder', () => {
 
       expect(graph.has(propertyTriple)).toBeTruthy()
 
-      let result = graph.filter(quad => quad.object.equals(rdf.literal('Very colorful')))
+      // Multiple items
+      const result1 = graph.filter(quad => quad.object.equals(rdf.literal('Very colorful'))).toArray()
 
-      expect(result.toArray().length).toEqual(1)
+      expect(result1.length).toEqual(1)
 
-      result = graph.filter(quad => quad.object.equals(rdf.literal('Sparkly')))
-      expect(result.toArray().length).toEqual(1)
+      const result2 = graph.filter(quad => quad.object.equals(rdf.literal('Sparkly'))).toArray()
+
+      expect(result2.length).toEqual(1)
+
+      // Multiple resources (notes)
+      const result3 = graph.filter(quad => quad.object.equals(rdf.literal('Shiney'))).toArray()
+
+      expect(result3.length).toEqual(1)
+      // Literals from the same note share blank node.
+      expect(result1[0].subject).toEqual(result2[0].subject)
+      // Literals from different notes have different blank node.
+      expect(result1[0].subject).not.toEqual(result3[0].subject)
     })
   })
 
--- a/__tests__/integration/repeatableResources.test.js
+++ b/__tests__/integration/repeatableResources.test.js
@@ -36,7 +36,7 @@ describe('Adding new embedded Resource Templates', () => {
   it('clicks on a nested property with an embedded Note resource template and then clicks on the AddButton for a second resource template', async () => {
     await pupExpect(page).toClick('a[data-id=\'note\']')
     await pupExpect(page).toClick('.col-sm-4 > div:nth-child(1) > button:nth-child(1)')
-    await pupExpect(page).toMatchElement('div.panel:nth-child(1) > div:nth-child(2) > div:nth-child(1) > div:nth-child(2) > div:nth-child(3) > div:nth-child(2) > h5:nth-child(2)',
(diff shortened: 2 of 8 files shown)


__tests__/integration/repeatableResources.test.js
```diff
@@ -36,7 +36,7 @@ describe('Adding new embedded Resource Templates', () => {
   it('clicks on a nested property with an embedded Note resource template and then clicks on the AddButton for a second resource template', async () => {
     await pupExpect(page).toClick('a[data-id=\'note\']')
     await pupExpect(page).toClick('.col-sm-4 > div:nth-child(1) > button:nth-child(1)')
```
ndushay: might it be possible to have a better selector for the add button that is more obviously ... the add button for the template?  I think we should avoid "nth-child" and classes that are rendering instructions when possible ...

__tests__/integration/repeatableResources.test.js
```diff
@@ -36,7 +36,7 @@ describe('Adding new embedded Resource Templates', () => {
...
-    await pupExpect(page).toMatchElement('div.panel:nth-child(1) > div:nth-child(2) > div:nth-child(1) > div:nth-child(2) > div:nth-child(3) > div:nth-child(2) > h5:nth-child(2)',
+    await pupExpect(page).toMatchElement('div.panel-body > div > div:nth-child(2) > div:nth-child(3) > div.rOutline-property > div:nth-child(2) > div.row > section.col-sm-8> h5',
```
ndushay: this selector is improved but still pretty yucky - can it be further improved to avoid "nth-child" and "col-sm-8" as a class name?

src/components/editor/property/PropertyResourceTemplate.jsx
```diff
@@ -21,12 +21,17 @@ class PropertyResourceTemplate extends Component {
     const existingOutputs = [...this.state.output]
 
     existingOutputs.push(<h4 key={shortid.generate()}>{this.props.resourceTemplate.resourceLabel}</h4>)
-    const result = this.populatePropertyTemplates()
+
+    const newReduxPath = [...this.props.reduxPath]
+
+    // Replace the generated id so that this is a new resource.
+    newReduxPath[newReduxPath.length - 2] = shortid.generate()
```
ndushay: I know I'm not looking at the whole here, but ... it is not obvious to me why we're working with an index of `newReduxPath.length - 2` ... could/should that be explained in the comment?

src/components/editor/property/PropertyTemplateOutline.jsx
```diff
@@ -82,49 +56,39 @@ export class PropertyTemplateOutline extends Component {
...
+      </PropertyTypeRow>,
+    )
 
     this.setState({ propertyTypeRow: newOutput, rowAdded: true })
   }
 
   render() {
     return (
-      <div className="rtOutline" key={shortid.generate()}>
+      <div className="rtOutline">
```
ndushay: So the key attribute that was here is now moved to PropertyComponent?

__tests__/GraphBuilder.test.js
```diff
@@ -71,12 +83,22 @@ describe('GraphBuilder', () => {
...
+      const result2 = graph.filter(quad => quad.object.equals(rdf.literal('Sparkly'))).toArray()
+
+      expect(result2.length).toEqual(1)
+
+      // Multiple resources (notes)
+      const result3 = graph.filter(quad => quad.object.equals(rdf.literal('Shiney'))).toArray()
+
+      expect(result3.length).toEqual(1)
+      // Each note has separate blank nodes.
+      expect(result1[0].subject).toEqual(result2[0].subject)
```
ndushay: showing that these 2 things are equal ... isn't showing that they are separate blank nodes ... what is this line actually testing?

src/components/editor/property/PropertyTemplateOutline.jsx
```diff
@@ -10,25 +10,6 @@ import { isResourceWithValueTemplateRef, resourceToName, templateBoolean } from
 import PropertyComponent from './PropertyComponent'
 import ResourceProperty from './ResourceProperty'
 
-export const addResourceTemplate = (resourceTemplate, reduxPath) => {
```
ndushay: yay!?  getting rid of code.

__tests__/integration/repeatableResources.test.js
justinlittman: Agree and open to suggestions. I have 0 expertise in this area.

src/components/editor/property/PropertyResourceTemplate.jsx
justinlittman: Adding an example redux path so that it is clearer why `-2`.

src/components/editor/property/PropertyTemplateOutline.jsx
justinlittman: Not that I fully understand this aspect of react, but this was forcing new components every time in rendered rather than re-using existing.

__tests__/GraphBuilder.test.js
justinlittman: Clarified.

src/components/editor/property/PropertyTemplateOutline.jsx
justinlittman: I'd feel better if I understood why it was there to begin with ...

__tests__/integration/repeatableResources.test.js
ndushay: You probably already keep this page handy while writing tests:  https://www.w3schools.com/cssref/css_selectors.asp.

justinlittman: I do now.

PR #700, opened 2019-06-13
Moves display validations to state from prop.

### LINKED ISSUE #698: Move displayValidations to state
Currently it is a property that originates in Editor and is passed as a property to just about every component.

justinlittman:
closes #698

### DIFF
--- a/src/components/editor/ResourceTemplate.jsx
+++ b/src/components/editor/ResourceTemplate.jsx
@@ -46,7 +46,6 @@ class ResourceTemplate extends Component {
             resourceTemplate = {this.state.rtData}
             parentResourceTemplate = {this.props.resourceTemplateId}
             rtId = {this.state.rtData.id}
-            displayValidations = {this.props.displayValidations}
         />
       </div>
     </div>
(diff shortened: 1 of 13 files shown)


src/components/editor/ResourceTemplate.jsx
```diff
@@ -46,7 +46,6 @@ class ResourceTemplate extends Component {
             resourceTemplate = {this.state.rtData}
             parentResourceTemplate = {this.props.resourceTemplateId}
             rtId = {this.state.rtData.id}
-            displayValidations = {this.props.displayValidations}
```
ndushay: so doing this via redux store at the Editor component level means we don't need it for these other places?  It seems like it moved from ResourceTemplate and ResourceTemplateForm and PropertyComponent to Editor and to property InputList, InputLiteral and InputLookupQA? 

Would it be possible for someone to clarify a bit more about how this will work?  (I just want to keep up with the group's thinking; not pushing back as much as wanting to be brought along)

justinlittman: Sure. I think we're trying to move from passing around properties to components getting what they need from state.

In the case of displayValidations, it was being passed as a property through the entire hierarchy of components from Editor on down. Now, it is updated in state by the Editor and only the components that need it receive the update.

ndushay: @justinlittman thanks for the explanation!

PR #702, opened 2019-06-13
For InputQA, switches from properties to state for propertyTemplate a…

### LINKED ISSUE #692: InputLookupQA rendering driven by Resource and Entity State
The InputLookupQA component currently uses a combination of PropertyTemplate passed in as prop and it's React state for handling properties like repeatable and mandatory and for setting defaults. All props and defaults should be set in the Redux reducer call in the `mapStateToProps` based on the InputLookupQAs' reduxPath with the reduxPath using the stored Resource Templates in the entities object in the Redux state.

This should also fix #670 

Blocks #673

justinlittman:
…nd lookupConfig.

closes #692

### DIFF
--- a/src/Utilities.js
+++ b/src/Utilities.js
@@ -1,5 +1,7 @@
 // Copyright 2018, 2019 Stanford University see LICENSE for license
 
+import lookupConfig from '../static/spoofedFilesFromServer/fromSinopiaServer/lookupConfig.json'
+
 const _ = require('lodash')
 
 export const isResourceWithValueTemplateRef = property => property?.type === 'resource'
@@ -73,3 +75,13 @@ export const defaultLangTemplate = () => ({
     },
   ],
 })
+
+export const getLookupConfigItems = (propertyTemplate) => {
+  const vocabUriList = propertyTemplate?.valueConstraint?.useValuesFrom
+
+  if (vocabUriList === undefined || vocabUriList.length === 0) return []
+
+  const templateConfigItems = lookupConfig.filter(configItem => vocabUriList.includes(configItem.uri))
+
+  return templateConfigItems
+}
--- a/src/components/editor/property/InputLookupQA.jsx
+++ b/src/components/editor/property/InputLookupQA.jsx
@@ -7,9 +7,9 @@ import PropTypes from 'prop-types'
 import Swagger from 'swagger-client'
 import swaggerSpec from '../../../lib/apidoc.json'
 import { connect } from 'react-redux'
-import { getProperty, getDisplayValidations } from '../../../reducers/index'
+import { getProperty, getDisplayValidations, getPropertyTemplate } from '../../../reducers/index'
 import { changeSelections } from '../../../actions/index'
-import { booleanPropertyFromTemplate, defaultValuesFromPropertyTemplate } from '../../../Utilities'
+import { booleanPropertyFromTemplate, defaultValuesFromPropertyTemplate, getLookupConfigItems } from '../../../Utilities'
 import Config from '../../../Config'
 
 const AsyncTypeahead = asyncContainer(Typeahead)
@@ -114,6 +114,11 @@ class InputLookupQA extends Component {
   }
 
   render() {
+    // Don't render if don't have property templates yet.
+    if (!this.props.propertyTemplate) {
+      return null
+    }
+
     let authority
     let language
     let subauthority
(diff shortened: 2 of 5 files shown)


src/Utilities.js
```diff
@@ -1,5 +1,7 @@
 // Copyright 2018, 2019 Stanford University see LICENSE for license
 
+import lookupConfig from '../static/spoofedFilesFromServer/fromSinopiaServer/lookupConfig.json'
```
ndushay: i'm uncomfy having code in production that is from a dir called "spoofedFilesFromServer" ... this feels like the wrong direction.  Move it to a non-spoof related directory and have the spoof read it from there?

src/components/editor/property/InputLookupQA.jsx
```diff
@@ -114,6 +114,11 @@ class InputLookupQA extends Component {
   }
 
   render() {
+    // Don't render if don't have property templates yet.
```
ndushay: naive question:  if we don't have them, do we know we will get them imminently?  Seems like that would be desirable ... (or surfacing to user what is missing)

src/components/editor/property/PropertyComponent.jsx
```diff
@@ -6,26 +6,16 @@ import PropTypes from 'prop-types'
 import InputLiteral from './InputLiteral'
 import InputListLOC from './InputListLOC'
 import InputLookupQA from './InputLookupQA'
-import lookupConfig from '../../../../static/spoofedFilesFromServer/fromSinopiaServer/lookupConfig.json'
+import { getLookupConfigItems } from '../../../Utilities'
```
ndushay: ok, I see we're making a move in the right direction here ...

src/Utilities.js
jermnelson: 👍

jermnelson:
I agree with @ndushay that having Spoofed calls in the code base isn't desirable; however it is working and moving it to Utilities is good enough for now. I created ticket #703

jcoyne:
@ndushay I think it's fine to say, "get rid of the spoofed stuff", but that's not the intention of this PR and so I think it's unreasonable to demand that here.

src/components/editor/property/InputLookupQA.jsx
justinlittman: If the application is going to work, then we will need them imminently. However, as far as this component, it has no idea ...

justinlittman: To be clearer: In normal operation, this component may render before a resource template has been loaded. However, since loading that template will change the state, this component will be re-rendered.

src/components/editor/property/InputLookupQA.jsx
```diff
@@ -227,16 +232,26 @@ InputLookupQA.propTypes = {
...
   reduxPath: PropTypes.oneOfType([PropTypes.string, PropTypes.array]),
   displayValidations: PropTypes.bool,
 }
 
 const mapStateToProps = (state, props) => {
-  const result = getProperty(state, props)
+  const reduxPath = props.reduxPath
+  const resourceTemplateId = reduxPath[reduxPath.length - 2]
+  const propertyURI = reduxPath[reduxPath.length - 1]
   const displayValidations = getDisplayValidations(state)
```
jermnelson: I think for now calling `getDisplayValidations`, `getPropertyTemplate`, and `getLookupConfigItems` directly from reducers/index.js module is fine. A future rewrite would have a call functions in the `actions/index.js` instead.

src/components/editor/property/InputLookupQA.jsx
ndushay: thank you for the explanation!  Mebbe chg comment to 

```
// Don't render if we don't have property templates yet -- we will re-render when we have it (immenently)

ndushay:
I'm not suggesting we get rid of spoofing now.  I'm suggesting that whatever we need from the spoofing directory, for production, be moved into a more appropriate directory/class and that the spoofing code can use it from the new location, and the prod code can also use it from the new location.  Not expecting a huge re-write ... just the separation of "spoof" from "config stuff we actually are using in prod".

jermnelson:
> I'm not suggesting we get rid of spoofing now. I'm suggesting that whatever we need from the spoofing directory, for production, be moved into a more appropriate directory/class and that the spoofing code can use it from the new location, and the prod code can also use it from the new location. Not expecting a huge re-write ... just the separation of "spoof" from "config stuff we actually are using in prod".

@ndushay can this be moved to a separate issue? We can tag it M4 so this done in this work-cycle.

ndushay:
Ya, if we do it this work cycle, I'm good with it being separate work.

PR #708, opened 2019-06-13
Change propLabel to data-label

### LINKED ISSUE #707: propLabel is not a valid property
I'm seeing this warning:
```
Warning: React does not recognize the `propLabel` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `proplabel` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
```

jcoyne:
propLabel was not recognized by react and raised a warning.  data-label can fulfill the same purpose and it doesn't raise a warning

Fixes #707


ndushay:
thanks for fixing this - sorry i missed the React warning when I put in my PR.

PR #710, opened 2019-06-13
Ensure tests can be run indenpendently

jcoyne:
Rename mocks so they aren't confusing.

### DIFF
--- a/__tests__/components/editor/property/InputLiteral.test.js
+++ b/__tests__/components/editor/property/InputLiteral.test.js
@@ -76,15 +76,19 @@ describe('When the user enters input into field', () => {
   const removeMockDataFn = jest.fn()
 
   shortid.generate = jest.fn().mockReturnValue(0)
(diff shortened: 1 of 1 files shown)


__tests__/components/editor/property/InputLiteral.test.js
```diff
@@ -71,20 +71,25 @@ describe('checkMandatoryRepeatable', () => {
...
-                                            reduxPath={[
-                                              'resourceTemplate:bf2:Monograph:Instance',
-                                              'http://id.loc.gov/ontologies/bibframe/instanceOf',
-                                            ]}
-                                            handleMyItemsChange={mockFormDataFn}
-                                            handleRemoveItem={removeMockDataFn}
-                                            handleMyItemsLangChange={jest.fn()} />)
+  let mockWrapper
+
+  beforeEach(() => {
```
jermnelson: Like this refactoring; setting `mockItemsChange` to the jest function mock in the beforeEach makes sense.

Issue #693, opened 2019-06-12
Preload all necessary ResourceTemplates


jcoyne:
Currently child templates are being loaded in `PropertyTemplateOutline.handleClick` and in `ResourceTemplateForm.componentDidMount`

ndushay:
Per discussion in meeting 2019-06-13:

Goal of this ticket:

When seeking a resource template (i.e. in `ResourceTemplateForm.componentDidMount` and `PropertyTemplateOutline.handleClick`), do the following:
   1. get it from redux store, if it's there
   2. if not in redux store, fetch it from sinopia server
   3. (not in scope right now):  if not in sinopia server, look elsewhere

Implementation Notes:
- *not* planning to do async calls in reducers at the present time
- will implement resourceTemplate retrieval per above either as new function in new class or as part of sinopia server wrapper, or something.  `ResourceTemplateForm.componentDidMount` and `PropertyTemplateOutline.handleClick` will use the new function whenever a resourceTemplate is needed.

Gains:
- do not do external calls for info we already have
- external call code in single place
- resource templates stored in single place (redux store) and always retrieved from single place (redux store) (?)

ndushay:
For reference:

https://redux.js.org/basics/reducers#handling-actions
> We'll explore how to perform side effects in the advanced walkthrough. For now, just remember that the reducer must be pure. *Given the same arguments, it should calculate the next state and return it. No surprises. No side effects. No API calls. No mutations. Just a calculation.* [emphasis in original]
(edited)

(JMartin: they dive into more async stuff here:  https://redux.js.org/advanced/async-actions)
> In the basics guide, we built a simple todo application. It was fully synchronous. Every time an action was dispatched, the state was updated immediately.
>
> In this guide, we will build a different, asynchronous application. It will use the Reddit API to show the current headlines for a selected subreddit. How does asynchronicity fit into Redux flow?

but my vote would be to do the async thing in the component as a first pass (as is happening now), and then to use that info to put all the resource templates into the redux state (the "fetching all the templates" part seems to be the functionality that is not yet there). (edited) 

---
Naomi sez:  I think the async thing should be in a sep module so it can be used by both react components that use it.

jmartin-sul:
> @ndushay sez: I think the async thing should be in a sep module so it can be used by both react components that use it.

yup, i'd agree w/ that

ndushay:
per PR #716 -- 

line 47 of ResourceTemplateForm:

[...]

PR #716, opened 2019-06-13
Get resource template from redux store

jmartin-sul:
rebased on master.  paired w/ @ndushay and @aaron-collier (left the original get-rt-from-store branch alone, just so you guys can make sure the rebase and linting didn't mess anything up).

lays a little ground work for making #693 easier, but doesn't actually resolve it.

### DIFF
--- a/src/components/editor/ResourceTemplateForm.jsx
+++ b/src/components/editor/ResourceTemplateForm.jsx
@@ -39,14 +39,14 @@ export class ResourceTemplateForm extends Component {
    */
   fulfillRTPromises = async (promiseAll) => {
     await promiseAll.then(async (rts) => {
-      rts.map((fulfilledResourceTemplateRequest) => {
-        const joinedRts = [...this.state.nestedResourceTemplates]
+      const joinedRts = [...this.state.nestedResourceTemplates]
 
+      rts.map((fulfilledResourceTemplateRequest) => {
         joinedRts.push(fulfilledResourceTemplateRequest.response.body)
         // Add the resource template into the store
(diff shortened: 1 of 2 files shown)


src/components/editor/ResourceTemplateForm.jsx
```diff
@@ -32,21 +32,53 @@ export class ResourceTemplateForm extends Component {
...
   fulfillRTPromises = async (promiseAll) => {
     await promiseAll.then(async (rts) => {
-      rts.map((fulfilledResourceTemplateRequest) => {
-        const joinedRts = [...this.state.nestedResourceTemplates]
+      const joinedRts = [...this.state.nestedResourceTemplates]
 
+      rts.map((fulfilledResourceTemplateRequest) => {
         joinedRts.push(fulfilledResourceTemplateRequest.response.body)
         // Add the resource template into the store
         store.dispatch(resourceTemplateLoaded(fulfilledResourceTemplateRequest.response.body))
```
jermnelson: Not blocking but any particular reason this isn't part of the `mapDispatchToProps`?

jmartin-sul: not sure.  i also found that slightly odd, but i didn't bother bringing it up during this PR, since that wasn't really the main point of it.  but i'd be happy to see this switch to using `mapDispatchToProps`, and we have some follow-on work to do to actually get #693 done, so we could maybe iron that out in the follow-on work, if you don't want to block this PR over it?

ndushay: added a comment to #693 so we remember to address this in future PR, if it isn't done here.

PR #722, opened 2019-06-14
Move local editor state for showing group choice modal to the Redux s…

jcoyne:
…tore

### DIFF
--- a/__tests__/components/editor/Editor.test.js
+++ b/__tests__/components/editor/Editor.test.js
@@ -2,9 +2,9 @@
 
 import React from 'react'
 import { shallow } from 'enzyme'
-import Editor from '../../../src/components/editor/Editor'
-import ResourceTemplate from '../../../src/components/editor/ResourceTemplate'
-import Header from '../../../src/components/editor/Header'
+import Editor from 'components/editor/Editor'
+import ResourceTemplate from 'components/editor/ResourceTemplate'
+import Header from 'components/editor/Header'
 
 const props = {
   location: { state: { resourceTemplateId: 'resourceTemplate:bf:Note' } },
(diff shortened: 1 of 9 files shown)


__tests__/components/editor/Editor.test.js
```diff
@@ -38,36 +38,32 @@ describe('<Editor />', () => {
...
-  describe('save button', () => {
-    const wrapper = shallow(<Editor.WrappedComponent {...props} />)
+  describe('Save & Publish button', () => {
+    const mockOpenHandler = jest.fn()
+    const wrapper = shallow(<Editor.WrappedComponent {...props} openGroupChooser={mockOpenHandler}/>)
 
     it('attempts to save the RDF content when save is clicked', () => {
-      wrapper.find('button').at(1).simulate('click')
-      expect(wrapper.state('showGroupChooser')).toBeTruthy()
+      wrapper.findWhere(n => n.type() === 'button' && n.contains('Save & Publish')).simulate('click')
```
ndushay: OMG I wish I had figured this out a long time ago.

__tests__/components/editor/Editor.test.js
```diff
@@ -2,9 +2,9 @@
 
 import React from 'react'
 import { shallow } from 'enzyme'
-import Editor from '../../../src/components/editor/Editor'
-import ResourceTemplate from '../../../src/components/editor/ResourceTemplate'
-import Header from '../../../src/components/editor/Header'
+import Editor from 'components/editor/Editor'
+import ResourceTemplate from 'components/editor/ResourceTemplate'
+import Header from 'components/editor/Header'
```
ndushay: love these cleaner imports

src/components/editor/Editor.jsx
```diff
@@ -3,7 +3,9 @@
 import React, { Component } from 'react'
 import { connect } from 'react-redux'
 import PropTypes from 'prop-types'
-import { removeAllItems, assignBaseURL, runValidation } from '../../actions/index'
+import {
+  removeAllItems, assignBaseURL, runValidation, showGroupChooser,
+} from '../../actions/index'
```
ndushay: if it's getting too big for a single line ... should we bother to import the pieces, or just import the whole thing?

src/reducers/index.js
```diff
@@ -1,10 +1,12 @@
 // Copyright 2018, 2019 Stanford University see LICENSE for license
+/* eslint complexity: ["warn", 13] */
 
 import { combineReducers } from 'redux'
 import shortid from 'shortid'
 import authenticate from './authenticate'
 import {
-  removeAllContent, removeMyItem, setMyItems, setMySelections, setBaseURL, setMyItemsLang, validate,
+  removeAllContent, removeMyItem, setMyItems, setMySelections, setBaseURL, setMyItemsLang,
+  validate, showGroupChooser,
```
ndushay: same question as above -- if we're importing this many pieces, would it be better to import the whole thing?

src/reducers/inputs.js
```diff
@@ -4,6 +4,16 @@ import Validator from '../Validator'
 
 export const validate = (state, action = { payload: {} }) => new Validator(state).validate(action.payload.show)
 
+/**
+ *  @param {Object} action the payload of the action is a boolean that says to show or not to show the chooser
+ */
+export const showGroupChooser = (state, action) => {
```
ndushay: need test for this, no?

jcoyne: We haven't added any functionality.  If this function was adequately tested before, why do we need more tests now?

src/reducers/index.js
jcoyne: I like being explicit here. I think it makes it easier to tell what we're actually using.

src/reducers/inputs.js
ndushay: It seems adding a function implies there should be tests for the function, in most cases.  If it wasn't adequately tested before ... isn't now the time to fix that?

src/components/editor/Editor.jsx
jmartin-sul: i've always had a slight personal preference for explicit and limited imports, but it's not all that strong

src/reducers/index.js
jmartin-sul: i vote for leaving like this, but am happy to go with consensus

src/reducers/inputs.js
jmartin-sul: that seems like a good idea (and a thing we've been doing elsewhere, right?).  no pref from me as to whether that happens in this PR or a follow on.

__tests__/components/editor/Editor.test.js
aaron-collier: Ditto!

src/reducers/inputs.js
jmartin-sul: if the current test is at the component level, but the component functionality is now wired differently (managing its state via redux) than before... and if there's no higher level integration test, is there anything actually testing the redux wiring?  that is to say, the previously adequate testing no longer seems as thorough with the new implementation, right?  i would expect this to be the reason for the small test coverage drop?  or am i missing how the new reducer gets exercised?

__tests__/components/editor/Editor.test.js
jmartin-sul: same!

src/reducers/inputs.js
jcoyne: Okay, test added.  @jmartin-sul yes, the integration tests cover this, but we're just not recording the data from that. We're only doing coverage of unit tests.   I think unit tests for simple operations like this are not very valuable.

jmartin-sul: ah, yeah, ok.  sorry, didn't realize there was an integration test for this.

PR #726, opened 2019-06-14
Remove unused variable tempRtState

jcoyne:
This was first introduced in https://github.com/LD4P/sinopia_editor/commit/30f49b0d1940a15e9c103aacdbe0307c54a5174c

### DIFF
--- a/src/components/editor/Editor.jsx
+++ b/src/components/editor/Editor.jsx
@@ -22,7 +22,6 @@ class Editor extends Component {
   constructor(props) {
     super(props)
     this.state = {
-      tempRtState: true,
       resourceTemplateId: '',
       showRdf: false,
       showGroupChooser: false,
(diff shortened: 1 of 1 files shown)


src/components/editor/Editor.jsx
```diff
@@ -22,23 +22,19 @@ class Editor extends Component {
...
   componentDidMount() {
-    if (this.state.tempRtState) {
-      if (this.props.location.state !== undefined) {
-        this.setState({
-          resourceTemplateId: this.props.location.state.resourceTemplateId,
-        })
-      } else {
-        this.props.history.push('/templates')
-      }
-      this.setState({ tempRtState: false })
```
ndushay: so presumably it was trying to avoid doing stuff more than once in componentDidMount ...  but spelunking around isn't giving me any more illumination.  I say if there's a problem, it will surface.

jcoyne: When this was originally written it was in componentWillUpdate, now that it's in componentDidMount, it doesn't make any sense as that is only called once.

jmartin-sul: ah, ok, that makes sense

PR #727, opened 2019-06-14
Do not add extra property rows when clicking the plus/add button.

### LINKED ISSUE #697: Clicking expand in OutlineHeader, collapse, and expand again duplicates PropertyTypeRow
Using the Spoofed `resourceTemplate:bf2:WorkTitle` as an example, clicking the Title note's **+** Note Text expands and displays an InputLiteral. Entering a value and hitting enter saves the value as seen here:

Screen Shot 2019-06-12 at 2 39 46 PM

Collapsing the PropertyTypeRow with the **-** button and then expanding again results in the InputLiteral with the same Redux path being duplicated:

Screen Shot 2019-06-12 at 2 41 32 PM

## NOTE:
This may be fixed with the refactoring to use Redux state for both Resources and Entities.

aaron-collier:
Fixes #697 

I am working on a more complete test for this. I'm happy to continue that test here or in a follow up PR if this fix needs to get in ASAP.

### DIFF
--- a/src/components/editor/property/PropertyTemplateOutline.jsx
+++ b/src/components/editor/property/PropertyTemplateOutline.jsx
@@ -51,13 +51,17 @@ export class PropertyTemplateOutline extends Component {
   // When the plus button is clicked, load reference templates (property.valueConstraint.valueTemplateRefs)
   handleClick = property => (event) => {
     event.preventDefault()
+
+    if (this.state.collapsed && !this.state.added) {
+      const templateRefList = isResourceWithValueTemplateRef(property) ? property.valueConstraint.valueTemplateRefs : []
+
+      this.fulfillRTPromises(this.resourceTemplatePromises(templateRefList)).then(() => {
+        this.addPropertyTypeRows(this.props.propertyTemplate)
+      })
+      this.setState({ added: true })
+    }
+
     this.setState({ collapsed: !this.state.collapsed })
-
-    const templateRefList = isResourceWithValueTemplateRef(property) ? property.valueConstraint.valueTemplateRefs : []
-
-    this.fulfillRTPromises(this.resourceTemplatePromises(templateRefList)).then(() => {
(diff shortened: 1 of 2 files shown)


ndushay:
Glad for bug fix;  not a fan of merging stuff without its tests, so I'd prefer to wait but can be overruled on that.

src/components/editor/property/PropertyTemplateOutline.jsx
```diff
@@ -51,13 +51,17 @@ export class PropertyTemplateOutline extends Component {
...
+    if (this.state.collapsed && !this.state.added) {
+      const templateRefList = isResourceWithValueTemplateRef(property) ? property.valueConstraint.valueTemplateRefs : []
 
-    this.fulfillRTPromises(this.resourceTemplatePromises(templateRefList)).then(() => {
-      this.addPropertyTypeRows(this.props.propertyTemplate)
-    })
+      this.fulfillRTPromises(this.resourceTemplatePromises(templateRefList)).then(() => {
+        this.addPropertyTypeRows(this.props.propertyTemplate)
+      })
+      this.setState({ added: true })
```
jcoyne: Is it possible to look at the redux store to tell if this is added without having to keep a separate piece of state around?

aaron-collier: I'll take a look. My first attempt at that led to a different bug, I'll time box working through that and report back soon.

aaron-collier: @jcoyne using the existing `rowAdded` state worked instead of creating a new state value. How do you feel about that?

jermnelson: @aaron-collier and @jcoyne is this now resolved or still open?

jcoyne: Yes, no problem with this.

PR #728, opened 2019-06-14
Switch GroupChoiceModal to a functional style (without state)

jcoyne:


jcoyne:
This is the "new way" to write React: https://reactjs.org/docs/hooks-intro.html `setState` is going to be deprecated in future versions of React.

jmartin-sul:
thanks for link to info on hooks, @jcoyne, wasn't aware of that

PR #734, opened 2019-06-14
Fix Blank Screen when performing a Lookup

### LINKED ISSUE #712: Linked Data Editor - Closes when trying to select a look-up entity
In stage - select sinopia:resourceTemplate:bf2:Monograph:Work Under "Primary Contribution" select "Primary contributor" > "Person" > "Search LCNAF"  > type name in look-up "Twain, Mark" and try to select an entity.  Editor page unexpectedly closes and the entity is not selected.  Similar behavior with other QA look-ups in different profiles and with different source look-ups.

### LINKED ISSUE #718: screen goes blank when you select a result of a lookup
reporting on behalf of Alexis McClimans at UW (@alexiskeely) 

take this template in stage
rt:rda:expression:monograph

go to the field "has creator agent of expression (RDA 20.2)(Lookup)"

search for a name like `smith`

choose one of the resulting names

Sinopia goes blank

jgreben:
Fixes #712:
Similar to https://github.com/LD4P/sinopia_editor/pull/684, if the lookup component has a `defaultURI` defined as an empty string, the `TypeaheadComponent` will expect a label to go along with that default and without one will crash. This PR redefines when the defaults should return an empty array to that component.

Fixes #718:
If the initial state of a redux resource object is undefined, and the defined property component assigns a `shortid` in the redux path, the component crashes when trying to filter on an undefined object in that path, so check for that in the reducer when reassigning state.

I ran across the bug surfaced in #718 while testing #712, so that's why this PR fixes both issues. I can split the PRs up if desired, but the changes here are not extensive.

### DIFF
--- a/src/reducers/inputs.js
+++ b/src/reducers/inputs.js
@@ -129,6 +129,10 @@ export const setMySelections = (state, action) => {
       obj[key].items = action.payload.items
     }
 
+    if (!Object.keys(obj).includes(key)) {
+      obj[key] = {}
+    }
+
     return obj[key]
   }, newState)
 
--- a/src/reducers/inputs.js
+++ b/src/reducers/inputs.js
@@ -65,7 +65,7 @@ export const removeAllContent = (state, action) => {
   return validate(newState)
 }
 
-export const setMyItems = (state, action) => {
+export const setItemsOrSelections = (state, action) => {
   const newState = { ...state }
   const reduxPath = action.payload.reduxPath
   let level = 0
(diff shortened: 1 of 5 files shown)


src/reducers/inputs.js
```diff
@@ -85,6 +85,10 @@ export const setMySelections = (state, action) => {
       obj[key].items = action.payload.items
     }
 
+    if (!Object.keys(obj).includes(key)) {
```
jermnelson: I think we are duplicating this code in the `setMyItems` in this reducer https://github.com/LD4P/sinopia_editor/blob/890b116ebb3f195d77969bd22e04abe2e488fd2a/src/reducers/inputs.js#L29-L44, 
Maybe collapse these functions or create a new function that both `setMySelections` and `setMyItems` call?

jermnelson: @jgreben, should I can make this a separate issue for follow-up work?

jgreben: > @jgreben, should I can make this a separate issue for follow-up work?

No need, I will take a look and make a follow-up commit.

jgreben: @jermnelson I combined the `setMyItems` and `changeMySelections` reducers into a single `setItemsOrSelections` reducer. LMK if this is what you had in mind.

jermnelson: LGTM; Thanks for doing this!

PR #735, opened 2019-06-14
ResourceProperty: separate state changes from rendering

### LINKED ISSUE #649: State change error when first opening "Title" on Work.
If you create a "BIBFRAME Instance" and in the "Instance of" (Work) panel, you click the button to expand the "Title Information"
Screen Shot 2019-06-06 at 9 37 20 AM

 this error occurs:

Screen Shot 2019-06-06 at 9 39 51 AM

This may be a contributor to #648

ndushay:
Fixes #649 - warning about state changes during rendering of `ResourceProperty` component.

Assumptions / Questions:
- [ ] It appears that we do not need to concern ourselves with `componentDidUpdate()` state changes, as this component is seemingly never updated after the first render (?)  -- all the pertinent state information is passed to child components (or they read it from redux)
- [ ] I believe that the same error from #649 would surface if I used component specific state (rather than redux) and updated component state during rendering -- hence a "class variable" approach.
- [ ] I'm unclear on the best way to have a "class variable" so that the `renderResourcePropertyJsx()` method can write to it, and it will never have cruft.
- [ ] I set the class variable to empty array in the constructor "just in case" - not sure it's needed.
- [ ] Because this is about a console warning that only appears when running in React development mode, (https://developmentarc.gitbooks.io/react-indepth/content/life_cycle/birth/component_render.html), I assumed allowing the existing tests to pass would be sufficient testing (aside from what I did manually).

### DIFF
--- a/src/components/editor/property/ResourceProperty.jsx
+++ b/src/components/editor/property/ResourceProperty.jsx
@@ -12,8 +12,17 @@ import { booleanPropertyFromTemplate } from 'Utilities'
 const _ = require('lodash')
 
 export class ResourceProperty extends Component {
+  dispatchPayloads = [] // to separate needed state changes from rendering
+
   constructor(props) {
     super(props)
+    this.dispatchPayloads = []
+  }
+
+  // NOTE:  if this component is ever re-rendered, we will need to ensure we get *changed* payloads to redux
+  //   as of 2019-06-18, the code base never re-renders this component, AFAICT.  We load the resource template up once.
+  componentDidMount() {
+    this.dispatchPayloads.forEach((payload) => { this.props.initNewResourceTemplate(payload) })
   }
 
   renderResourcePropertyJsx = () => {
@@ -56,8 +65,8 @@ export class ResourceProperty extends Component {
 
         propertyReduxPath.push(rtProperty.propertyURI)
         const payload = { reduxPath: propertyReduxPath, property: rtProperty }
+        this.dispatchPayloads.push(payload)
 
-        this.props.initNewResourceTemplate(payload)
         const isAddDisabled = !booleanPropertyFromTemplate(rtProperty, 'repeatable', false)
 
         jsx.push(
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -134,7 +134,8 @@ module.exports = {
     "import/no-namespace": "off",
     "no-nested-ternary": "off",
     'lines-around-comment': ['error', { 'allowBlockStart': true }],
-    'capitalized-comments': "off"
+    'capitalized-comments': "off",
+    'multiline-comment-style': "off"
   },
   overrides: [
     {


src/components/editor/property/ResourceProperty.jsx
```diff
@@ -56,8 +63,8 @@ export class ResourceProperty extends Component {
 
         propertyReduxPath.push(rtProperty.propertyURI)
         const payload = { reduxPath: propertyReduxPath, property: rtProperty }
+        this.dispatchPayloads.push(payload)
```
jcoyne: This pushes onto the array, but "componentDidMount" has already been called, so this won't do anything, right?

ndushay: - the tests failed before I added the calls to `initNewResourceTemplate` in componentDidMount();  there are tests expecting it to be called ...  but maybe they aren't actually testing the redux state, just that the method is called.

Does this imply I need to add a test to show the redux store is correct after render is called?

(I'm increasingly concerned our test coverage data is misleading - that we have code lines being executed by our tests, but the tests aren't sufficient to show the app is working properly ...)

ndushay: @jcoyne "The componentDidMount() method runs after the component output has been rendered to the DOM."  https://reactjs.org/docs/state-and-lifecycle.html

jcoyne: Thanks for the clarification. 👍

ndushay: I think this PR is incorrect *IFF* we ever re-render a ResourceProperty, but I don't believe we do???

jcoyne: If this re-pushes every time we re-render, does `dispatchPayload` grow without bounds?

ndushay: I don't think we ever do re-render.

jmartin-sul: more from the docs https://reactjs.org/docs/react-component.html#componentdidmount

> `componentDidMount()` is invoked immediately after a component is mounted (inserted into the tree). Initialization that requires DOM nodes should go here. If you need to load data from a remote endpoint, this is a good place to instantiate the network request.

based on that, and naomi's documentation find above, that implies to me that `componentDidMount()` runs after the output has been rendered to the DOM, but only for the render that happens after the first time the component mounts.  it doesn't _seem_ to me, based on what i've seen, like that function runs after every render.

i'm, uh, not even entirely sure this comment is helpful... but left it just in case. 🙂

ndushay: agree: componentDidMount() is only run for initial render.  

have no conflicting evidence that we would ever re-render this component.  adding a comment to that effect.

ndushay:
okay - added a comment re: code not updating redux on a re-render.  I hope that's enough to get this merged.

PR #737, opened 2019-06-14
Use Utility function in InputLiteral component to determine the boolean values from JSON strings

### LINKED ISSUE #736: InputLiteral TypeError: _this.props.formData.items is undefined
When loading the `Sinopia:RT:Demo:SampleResourceFuller` or by expanding the Barcode in the spoofed `resourceTemplate:bf2:Monograph:Instance` resource template, a **TypeError: _this.props.formData.items is undefined** is being raised with a blank Editor view. This may be due to a possible issue with the defaults not being set currently in the Redux state for an `InputLiteral`

jgreben:
PR #738 Fixed #736. This PR is now only about using the Utility function to determine the boolean values from JSON strings.

### DIFF
--- a/src/components/editor/property/InputLiteral.jsx
+++ b/src/components/editor/property/InputLiteral.jsx
@@ -9,7 +9,7 @@ import shortid from 'shortid'
 import { removeItem, setItems, setLang } from 'actions/index'
 import { findNode, getDisplayValidations, getPropertyTemplate } from 'reducers/index'
 import InputLang from './InputLang'
-import { defaultLangTemplate } from 'Utilities'
+import { booleanPropertyFromTemplate, defaultLangTemplate } from 'Utilities'
 
 
 // Redux recommends exporting the unconnected component for unit tests.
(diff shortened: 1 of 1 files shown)


jgreben:
Waiting for #738 to be merged

jcoyne:
@jgreben In which situations would the props have null formData?

jgreben:
> @jgreben In which situations would the props have null formData?

@jcoyne OK, I see that `findNode` will set `formData` to `{}` by default. Thanks.

src/components/editor/property/InputLiteral.jsx
```diff
@@ -118,13 +119,13 @@ export class InputLiteral extends Component {
     this.inputLiteralRef.current.focus()
   }
 
+  /**
+   * @return {bool} true if the field should be marked as required (e.g. not all obligations met)
```
ndushay: say default is false?  or is that obvious?

jcoyne: I think it's pretty clear from how it's documented that if it doesn't meet that condition it's false.

PR #738, opened 2019-06-14
Items must be properties or removing them doesn't cause re-render

### LINKED ISSUE #739: Pressing the remove button on a literal input does not rerender.

jcoyne:
Fixes #739

### DIFF
--- a/src/components/editor/property/InputLiteral.jsx
+++ b/src/components/editor/property/InputLiteral.jsx
@@ -25,22 +25,13 @@ export class InputLiteral extends Component {
     this.state = {
       show: {},
       content_add: '',
-      disabled: false,
       lang_payload: null,
     }
     this.inputLiteralRef = React.createRef()
   }
 
-  componentDidMount = () => {
-    // Note that it is possible that the resource template hasn't been loaded yet, hence there will not be a property template.
-    if (this.props.propertyTemplate
-        && this.props.propertyTemplate.repeatable === 'false'
-        && this.props.formData !== undefined
-        && this.props.formData.items.length > 0) {
-      this.setState({ disabled: true })
-    }
-  }
-
+  disabled = () => this.props.propertyTemplate.repeatable === 'false'
+      && this.props.items?.length > 0
 
   handleShow = (id) => {
     const showState = {}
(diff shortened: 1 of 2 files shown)


src/components/editor/property/InputLiteral.jsx
```diff
@@ -25,23 +25,16 @@ export class InputLiteral extends Component {
...
-  componentDidMount = () => {
-    // Note that it is possible that the resource template hasn't been loaded yet, hence there will not be a property template.
-    if (this.props.propertyTemplate
-        && this.props.propertyTemplate.repeatable === 'false'
-        && this.props.formData !== undefined
-        && this.props.formData.items.length > 0) {
-      this.setState({ disabled: true })
-    }
+  disabled = () => {
+    return this.props.propertyTemplate.repeatable === 'false' &&
```
jgreben: @jcoyne I think we both worked on refactoring the same thing, but I see here that we do not need the componentDidMount method anymore... How about using this instead (import from Utilities) which uses safe navigation for `repeatable`?

`!booleanPropertyFromTemplate(this.props.propertyTemplate, 'repeatable', true)`

jcoyne: This was already here, it's not something I changed in this PR. I try not to change more than is absolutely necessary. If you want to change it in another PR, please go ahead.

jgreben:
> Fixes #739

This sould also fix #736

src/components/editor/property/InputLiteral.jsx
jgreben: Ok when this is out of draft I will merge it and then rebase.

src/components/editor/property/InputLiteral.jsx
```diff
@@ -64,15 +55,7 @@ export class InputLiteral extends Component {
...
-  notRepeatableAfterUserInput = (userInputArray, currentcontent) => {
-    let newId = null
-
-    if (this.props.formData === undefined || this.props.formData.items < 1) {
-      newId = this.addUserInput(userInputArray, currentcontent)
-      this.setState({ disabled: true })
-    }
-    return newId
-  }
+  notRepeatableAfterUserInput = (userInputArray, currentcontent) => this.addUserInput(userInputArray, currentcontent)
```
jgreben: Is this necessary anymore? If it just calls `addUserInput()` then can line 78 just call `addUserInput()` regardless of whether the field is repeatable?

jcoyne: Thanks. That's a good suggestion. I've made that change.

PR #740, opened 2019-06-15
get the correct list of groups to display (and sort them in alphabetical order) when choosing where to save a resource

### LINKED ISSUE #732: add rest of group names to Save & Publish dialog
Container names. Label in quotes. slug following (I'm not sure what a slug is so if it's something different you want lmk)

"University of Alberta" | alberta
"University of Colorado, Boulder" | boulder
"University of Chicago" | chicago
"Cornell University" | cornell
"Library of Congress" | dlc
"Duke University" | duke
"Frick Art Reference Library" | frick
"Harvard University" | harvard
"University of Texas, Austin, Harry Ransom Center" | hrc
"University of Michigan" | michigan
"University of Minnesota" | minnesota
"National Library of Medicine" | nlm
"Northwestern University" | northwestern
"PCC" | pcc
"University of Pennsylvania" | penn
"Princeton University" | princeton
"Stanford" | stanford
"Texas A&M University" | tamu
"University of California, Davis" | ucdavis
"University of California, San Diego" | ucsd
"University of Washington" | washington
"Yale University" | yale

_Originally posted by @michelleif in https://github.com/LD4P/sinopia/issues/196#issuecomment-482147892_

jmartin-sul:
a bit of light refactoring and tweaking to make it happen, plus a new integration test to make sure it worked.

the sort wasn't asked for, but it seemed easy enough to do, so i went ahead and did that.

if this gets merged as-is, we should also merge the comment in LD4P/sinopia_acl#75

closes #732

### DIFF
--- a/__tests__/components/editor/GroupChoiceModal.test.js
+++ b/__tests__/components/editor/GroupChoiceModal.test.js
@@ -51,23 +51,15 @@ describe('<GroupChoiceModal />', () => {
     })
 
     it('has the first select option as "Cornell University"', () => {
-      expect(wrapper.find(Modal.Body).find('form').find('select').find('option')
-        .first()
-        .childAt(0)
-        .text()).toEqual('Cornell University')
-      expect(wrapper.find(Modal.Body).find('form').find('select').find('option')
-        .first()
-        .prop('value')).toEqual('cornell')
+      const firstOpt = wrapper.find(Modal.Body).find('form select option').first()
+      expect(firstOpt.text()).toEqual('Cornell University')
+      expect(firstOpt.prop('value')).toEqual('cornell')
     })
 
     it('has the last select option as "Yale University"', () => {
-      expect(wrapper.find(Modal.Body).find('form').find('select').find('option')
-        .last()
-        .childAt(0)
-        .text()).toEqual('Yale University')
-      expect(wrapper.find(Modal.Body).find('form').find('select').find('option')
-        .last()
-        .prop('value')).toEqual('yale')
+      const lastOpt = wrapper.find(Modal.Body).find('form select option').last()
+      expect(lastOpt.text()).toEqual('Yale University')
+      expect(lastOpt.prop('value')).toEqual('yale')
     })
 
     it('displays the groups from the config in alphabetical order, with ld4p omitted', () => {
--- a/__tests__/components/editor/GroupChoiceModal.test.js
+++ b/__tests__/components/editor/GroupChoiceModal.test.js
@@ -4,6 +4,7 @@ import React from 'react'
 import Modal from 'react-bootstrap/lib/Modal'
 import Button from 'react-bootstrap/lib/Button'
 import { shallow } from 'enzyme'
+import Config from 'Config'
 import GroupChoiceModal from 'components/editor/GroupChoiceModal'
 /* eslint import/namespace: 'off' */
 import * as server from 'sinopiaServer'
@@ -19,13 +20,10 @@ describe('<GroupChoiceModal />', () => {
   const rdfFunc = jest.fn()
   const closeFunc = jest.fn()
   const closeRdfPreview = jest.fn()
-  const groups = [['pcc', 'PCC'],
-    ['wau', 'University of Washington']]
 
   const wrapper = shallow(<GroupChoiceModal.WrappedComponent show={true}
                                                              rdf={rdfFunc}
                                                              close={closeFunc}
-                                                             groups={groups}
                                                              closeRdfPreview={closeRdfPreview} />)
 
   it('renders the <GroupChoiceModal /> component as a Modal', () => {
@@ -52,24 +50,36 @@ describe('<GroupChoiceModal />', () => {
       expect(wrapper.find(Modal.Body).find('form').find('select').length).toEqual(1)
     })
 
-    it('has the first select option as "PCC"', () => {
+    it('has the first select option as "Cornell University"', () => {
       expect(wrapper.find(Modal.Body).find('form').find('select').find('option')
         .first()
         .childAt(0)
-        .text()).toEqual('PCC')
+        .text()).toEqual('Cornell University')
       expect(wrapper.find(Modal.Body).find('form').find('select').find('option')
         .first()
-        .prop('value')).toEqual('pcc')
+        .prop('value')).toEqual('cornell')
     })
 
-    it('has the last select option as "University of Washington"', () => {
+    it('has the last select option as "Yale University"', () => {
       expect(wrapper.find(Modal.Body).find('form').find('select').find('option')
         .last()
         .childAt(0)
-        .text()).toEqual('University of Washington')
+        .text()).toEqual('Yale University')
       expect(wrapper.find(Modal.Body).find('form').find('select').find('option')
         .last()
-        .prop('value')).toEqual('wau')
+        .prop('value')).toEqual('yale')
+    })
+
+    it('displays the groups from the config in alphabetical order, with ld4p omitted', () => {
+      const expectedGroups = Object.entries(Config.groupsInSinopia)
+        .filter(([groupSlug]) => groupSlug !== 'ld4p')
+        .sort(([, groupLabelA], [, groupLabelB]) => groupLabelA.localeCompare(groupLabelB))
+
+      const actualGroups = wrapper.find(Modal.Body)
+        .find('form.group-select-options select option')
+        .map(node => [node.prop('value'), node.text()])
+
+      expect(actualGroups).toEqual(expectedGroups)
     })
 
     it('has a Cancel link', () => {
--- a/src/Config.js
+++ b/src/Config.js
@@ -72,20 +72,33 @@ class Config {
     return process.env.MAX_RECORDS_FOR_QA_LOOKUPS || 8
   }
 
+  // WARNING: the groups section in config/default.js in the sinopia_acl codebase *must* be kept in sync with this section
   static get groupsInSinopia() {
-    return [
-      ['ld4p', 'LD4P'],
-      ['pcc', 'PCC'],
-      ['cub', 'University of Colorado Boulder'],
-      ['cornell', 'Cornell University'],
-      ['harvard', 'Harvard University'],
-      ['nlm', 'National Library of Medicine'],
-      ['stanford', 'Stanford University'],
-      ['ucsd', 'University of California, San Diego'],
-      ['penn', 'University of Pennsylvania'],
-      ['hrc', 'Harry Ransom Center, University of Texas at Austin'],
-      ['wau', 'University of Washington'],
-    ]
+    return {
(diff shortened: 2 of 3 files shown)


src/Config.js
```diff
@@ -73,19 +73,31 @@ class Config {
...
+      pcc: 'PCC',
+      penn: 'University of Pennsylvania',
+      princeton: 'Princeton University',
+      stanford: 'Stanford University',
+      tamu: 'Texas A&M University',
+      ucdavis: 'University of California, Davis',
+      ucsd: 'University of California, San Diego',
+      washington: 'University of Washington',
+      yale: 'Yale University',
+    }
```
jmartin-sul: i just copied the structure seen here:
https://github.com/LD4P/sinopia_acl/blob/21284eb72cbe13012d700b225fec05b1b2a59bec/config/default.js#L24-L48

always unfortunate to have duplicated config, but this is at least a fix for the immediate issue that doesn't make things worse (now that the structure lines up, it's copying without reformatting, at least).

i'd considered getting the info from the server, since the `/repository` container lists all the groups (since the groups are all direct children).  unfortunately, that wouldn't contain the label info for each group, just the slug, and another request to trellis for _each_ group would be needed to get the labels that way (i think, unless there's a more clever way to query it that i'm not aware of).

two ideas i can think of off-hand to get rid of the config duplication:
* in sinopia_acl, put this config in a separate file, and pull a copy of that into this repo when building the editor image
* index group info to elastic search and ask elastic search for it

i think i prefer the latter of those two.

or, if we had a graph store (e.g. neo4j or neptune), we could do graphier queries.

but hopefully this'll do for now.

src/components/editor/Editor.jsx
```diff
@@ -61,7 +61,9 @@ class Editor extends Component {
   }
 
   // The ld4p group is only for templates
-  groupsToSaveInto = () => Config.groupsInSinopia.filter(group => group[0] !== 'ld4p')
+  groupsToSaveInto = () => Object.entries(Config.groupsInSinopia)
```
jmartin-sul: i ripped off the `Object.entries` approach from the sinopia_acl codebase, since that's where i got the new config structure.

src/components/editor/Editor.jsx
```diff
@@ -61,7 +61,9 @@ class Editor extends Component {
...
+    .filter(([groupSlug]) => groupSlug !== 'ld4p')
```
jmartin-sul: i thought the object destructuring on the array var made things a bit more readable, since it gave a var name for the thing being compared, instead of just an array subscript

src/components/editor/GroupChoiceModal.jsx
```diff
@@ -33,7 +33,7 @@ const GroupChoiceModal = (props) => {
           <div>
             <form className="group-select-options" >
               <select defaultValue={ selectedValue } onBlur={ event => updateSelectedValue(event)} >
-                { props.groups.map((group, index) => <option key={index} value={ group[0] }>{ group[1] }</option>) }
+                { props.groups.map(([groupSlug, groupLabel], index) => <option key={index} value={ groupSlug }>{ groupLabel }</option>) }
```
jmartin-sul: ditto here with the object destructuring instead of referencing by index

__tests__/integration/previewSaveRDF.test.js
```diff
@@ -27,6 +28,13 @@ describe('Previewing the RDF', () => {
     await pupExpect(page).toMatch('Which group do you want to save to?')
     await pupExpect(page).toMatch('Which group do you want to associate this record to?')
 
+    // test to make sure select list has sorted list of groups from config?
+    const groupsForSaving = await page.$$eval('form.group-select-options select option', nodes => nodes.map(n => [n.value, n.innerText]))
+    const expectedGroups = Object.entries(Config.groupsInSinopia)
```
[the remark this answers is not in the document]
jmartin-sul: @jcoyne, yeah, fair point.

i'm not really sure, though.  the thing i wanted to test is that all of the groups we care about get displayed in alphabetical order.  since the filtering and alphabetizing happens in `Editor`, but the display happens in `GroupChoiceModal`, i wasn't sure how to test that.

i guess maybe i could try to write a unit test for `Editor` that makes sure the right list gets passed to `GroupChoiceModal`?  i do sorta like having an integration test that gets at what's being done here, but i don't feel that strongly about it, and displaying the right list in the right order is what i was really trying to get at here, one way or the other.

how strong a pref do you have about this?  like what would be your pref as far as blocking merge of this PR to turn this test into a unit test?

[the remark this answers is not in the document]
jcoyne: Well, you can see how big the test already grew (as you need to rebase it).  It's already super complicated.  I'd be pretty satisfied by just a unit test for Editor.

src/Config.js
ndushay: I'm fine with the duplication for now; might I suggest adding a comment about the duplication in both places?

__tests__/integration/previewSaveRDF.test.js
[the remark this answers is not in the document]
ndushay: I'd suggest possibly creating a separate integration test file with just this one test??  Also, maybe some of the pupExpect statements could change to waitForSelector so they are clearly more about setup than test expectations?

src/Config.js
jmartin-sul: @ndushay already have a PR up for the comment in sinopia_acl (LD4P/sinopia_acl#75), and added a comment to that effect in this PR too

__tests__/integration/previewSaveRDF.test.js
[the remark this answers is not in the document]
jmartin-sul: @jcoyne and @ndushay, i just took the idea of the integration test and moved it over to `GroupChoiceModal.test.js`, which, as luck would have it, was much easier in light of @jcoyne's #748.  that felt slightly more satisfying to me than testing what `Editor` passed along.

ended up just backing out of the integration test.

jmartin-sul:
@ndushay, are you ok with the lack of integration test now that the encapsulation of group list display makes the desired behavior more unit testable within one component?  wanting to be able to test what was displayed based on the config was the only reason i went for the integration test in the first place, since the list retrieval and display previously spanned two components.

__tests__/components/editor/GroupChoiceModal.test.js
```diff
@@ -19,13 +20,10 @@ describe('<GroupChoiceModal />', () => {
   const rdfFunc = jest.fn()
   const closeFunc = jest.fn()
   const closeRdfPreview = jest.fn()
-  const groups = [['pcc', 'PCC'],
-    ['wau', 'University of Washington']]
 
   const wrapper = shallow(<GroupChoiceModal.WrappedComponent show={true}
                                                              rdf={rdfFunc}
                                                              close={closeFunc}
-                                                             groups={groups}
```
jmartin-sul: this was a bit of cruft from #748

__tests__/components/editor/GroupChoiceModal.test.js
```diff
@@ -52,24 +50,28 @@ describe('<GroupChoiceModal />', () => {
...
-        .childAt(0)
-        .text()).toEqual('University of Washington')
-      expect(wrapper.find(Modal.Body).find('form').find('select').find('option')
-        .last()
-        .prop('value')).toEqual('wau')
+    it('has the last select option as "Yale University"', () => {
+      const lastOpt = wrapper.find(Modal.Body).find('form select option').last()
+      expect(lastOpt.text()).toEqual('Yale University')
+      expect(lastOpt.prop('value')).toEqual('yale')
+    })
```
jmartin-sul: my first instinct was to get rid of this test and the one above, in light of the one below, which implicitly tests what these two test.  but... i couldn't quite get myself to do it, because i like the simplicity/directness of these two.  hopefully i at least made them a little more readable?

happy to ditch if people would prefer.

__tests__/components/editor/GroupChoiceModal.test.js
jcoyne: Thanks for cleaning this up.

__tests__/components/editor/GroupChoiceModal.test.js
```diff
@@ -52,24 +50,28 @@ describe('<GroupChoiceModal />', () => {
...
+
+    it('displays the groups from the config in alphabetical order, with ld4p omitted', () => {
+      const expectedGroups = Object.entries(Config.groupsInSinopia)
```
jcoyne: I'm not a big fan of this test because basically it's just duplicating the code that is being added to `GroupChoiceModal.jsx`.  It's saying "this code" behaves the same as "this code".  Not really worth holding the PR over though.

jmartin-sul: yeah, so i had a little consternation over that.  i think the thing of value that it adds is that it's testing that the component renders the select box as desired.  so it's not _solely_ "does this code implement this particular filter and sort?"  but it's close, and i wasn't wild about that.

i'd thought about instead sorting and filtering through some more naive/different approach.  but each one that i started writing was way more unwieldy than the current approach, which seems like the obvious way to do those two things to a list in JS.  doing otherwise felt a little hokey, and i wasn't sure how people would feel about that.

putting the whole sorted list in explicitly felt too fragile, but maybe that's what i should've done, and just called it a day?

anyway, yeah, i'm inclined to merge this as-is if @ndushay doesn't object, and maybe touch this up as follow on work if you think it's worth it?

__tests__/components/editor/GroupChoiceModal.test.js
```diff
@@ -52,24 +50,28 @@ describe('<GroupChoiceModal />', () => {
...
 
-    it('has the last select option as "University of Washington"', () => {
-      expect(wrapper.find(Modal.Body).find('form').find('select').find('option')
-        .last()
-        .childAt(0)
-        .text()).toEqual('University of Washington')
-      expect(wrapper.find(Modal.Body).find('form').find('select').find('option')
-        .last()
-        .prop('value')).toEqual('wau')
+    it('has the last select option as "Yale University"', () => {
```
ndushay: quick question I wouldn't hold this up for -- would we want to have a test for the number of expected groups??

__tests__/components/editor/GroupChoiceModal.test.js
ndushay: I'm fine with it as a follow up ...

__tests__/components/editor/GroupChoiceModal.test.js
jmartin-sul: if the test that @jcoyne doesn't like sticks around, i think that's unnecessary, because it is implicitly tested there (since it tests for equality against the entire expected list).

but since no one really likes that test a lot as is, if it gets reworked or discarded, maybe? 🙂

jmartin-sul:
merging since this has been approved...

PR #741, opened 2019-06-17
[Refactor] Organize the components into directories by routes

jcoyne:


__tests__/components/Header.test.js
```diff
@@ -1,45 +1,28 @@
-// Copyright 2018 Stanford University see LICENSE for license
```
jcoyne: This file moved to `__tests__/components/home/Header.test.js` and the file that was in `__tests__/components/editor/Header.test.js` has replaced it.

src/components/Header.jsx
```diff
@@ -1,38 +1,43 @@
-// Copyright 2018 Stanford University see LICENSE for license
```
jcoyne: This file moved to `src/components/home/Header.jsx` and the file that was in `src/components/editor/Header.jsx` has replaced it.

jermnelson:
> These changes look logical to me. I cannot currently judge how React-idiomatic this practice is, so I'm deferring to @jermnelson and others on the team.

Organizing by routes looks like a common [React file structure pattern](https://reactjs.org/docs/faq-structure.html). Following their advice, we are probably at the limit of recommended nesting in the directory structure with this PR.

PR #749, opened 2019-06-17
Move logic for saving to GroupChoiceModal from Editor

jcoyne:
This reduces coupling between the components and Editor doesn't have to know how to save


jgreben:
@jcoyne this is good to go IMO, just needs conflicts fixed.

PR #753, opened 2019-06-18
move lookupConfig out of spoofed directory

### LINKED ISSUE #719: prod code shouldn't do imports from static/spoofedFilesFromServer directory
per discussion in PR #702:

Whatever we need from the spoofing directory, for production, should be moved into a more appropriate directory/class.

The spoofing code can use it from the new location, and the prod code can also use it from the new location. 

This will be the separation of "spoof" from "config stuff we actually are using in prod".

See:

- `src/Utilities` (if #702 is merged)  -  used to be in `PropertyComponent`

If it is simple to do, consider #457, too.  (Or just move "lookupConfig" file to new dir)

ndushay:
Fixes #719 

@jcoyne @justinlittman @jermnelson this is all I meant.  Simple.


jcoyne:
Works for me, but I'm not certain about the history here.  @jermnelson ?

jermnelson:
The spoofed location decision was just done for expediency, I'm fine with the file move but still seems kinda of pedantic.

jmartin-sul:
assuming my understanding is correct, and this isn't a file that would come from trellis, i appreciate the move.  i think the tendency is to assume that inherited code is correct until proven otherwise, and so it could be a minor point of confusion to see a file that's not from the server in a dir called `fromSinopiaServer` (as a new dev, or an old one who hasn't seen this code in a while).

like not-quite-right var names, i tend to find that those things don't seem worth fussing over when you write them... but then when you're re-reading the code months later, you wish that you'd fussed over them a bit more 😢

PR #754, opened 2019-06-18
Convert the Editor component to a Function

jcoyne:


src/components/editor/Editor.jsx
```diff
@@ -14,32 +14,32 @@ import AuthenticationMessage from './AuthenticationMessage'
 /**
  * This is the root component of the resource edit page
  */
-class Editor extends Component {
-  componentDidMount() {
-    if (!this.props.location.state) {
-      this.props.history.push('/templates')
+const Editor = (props) => {
+  useEffect(() => {
```
jermnelson: Like the use of the `useEffect` hook; (always good to learn something new!)

src/components/editor/ResourceTemplate.jsx
```diff
@@ -20,9 +20,11 @@ class ResourceTemplate extends Component {
     }
   }
 
-  // Called immediately after the component is mounted
+  // Called immediately after the component is rendered for the first time
```
ndushay: someone wise has conditioned me away from comments and tests that are about "how it works" - and this comment is precisely how "componentDidMount()" works.

PR #756, opened 2019-06-18
Allow resources to be updated 

### LINKED ISSUE #664: When a resource is saved a second time it should persist to the same URI
Related to #669

### LINKED ISSUE #669: After first save & publish, subsequent calls should use updateResource
After saving the RDF the first time and the URI is persistented to the Redux state,  any changes in the UI are reflected in the Redux and show up in the **Preview RDF** modal. Clicking the **Save &amp; Publish** button should bypass the select group modal and also do a [`updateResource`](https://github.com/LD4P/sinopia_server/blob/master/sinopia_client/docs/LDPApi.md#updateResource) `sinopia_client` call.

jcoyne:
Fixes #669
Fixes #664

### DIFF
--- a/__tests__/actionCreators.test.js
+++ b/__tests__/actionCreators.test.js
@@ -1,6 +1,10 @@
 // Copyright 2019 Stanford University see LICENSE for license
 
-import { authenticationFailed, authenticationSucceeded, signedOut } from '../src/actionCreators'
+import {
+  authenticationFailed, authenticationSucceeded, signedOut, update,
+} from 'actionCreators'
+/* eslint import/namespace: 'off' */
+import * as server from 'sinopiaServer'
 
 describe('authenticationFailed', () => {
   const currentUser = { hello: 'world' }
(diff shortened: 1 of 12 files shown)


__tests__/actionCreators.test.js
```diff
@@ -35,3 +37,9 @@ describe('signedOut', () => {
     expect(signedOut().type).toEqual('SIGN_OUT_SUCCESS')
   })
 })
+
+describe('update', () => {
+  it('returns a function', () => {
+    expect(update()).toBeInstanceOf(Function)
```
ndushay: would we not want a test that is a bit more useful than this?  Like, showing what the function does?

src/sinopiaServer.js
```diff
@@ -154,3 +154,11 @@ export const publishRDFResource = async (currentUser, rdf, group) => {
   await authenticate(currentUser)
   return await instance.createResourceWithHttpInfo(group, rdf, { contentType: 'application/n-triples' })
 }
+
+export const updateRDFResource = async (currentUser, uri, rdf) => {
```
ndushay: would this new function warrant a test?

__tests__/actionCreators.test.js
jcoyne: I've added a test, but it's sort of a mess of async mocks.  I'm not sure if it's more helpful.

src/sinopiaServer.js
jcoyne: This is only making two calls: authenticate and then the server.  I don't know that a test is going to be useful here other than to say: "Here are some mocks that show it's doing exactly what it's doing"

jcoyne: Also, this is covered by the integration test.

__tests__/actionCreators.test.js
ndushay: I see what you mean.

PR #759, opened 2019-06-18
Restore the ability for defaults to be used in a lookup

### LINKED ISSUE #755: Lookup with default is not using the default
I'm seeing that "Example 9: lookup, repeatable, required, has default" on the Fuller Demo seems to not have a default.
Screen Shot 2019-06-18 at 8 52 16 AM

jcoyne:
This was causing problems because it was expecting props to be set in the constructor rather than the render function

Fixes #755

### DIFF
--- a/__tests__/components/editor/property/InputLookupQA.test.js
+++ b/__tests__/components/editor/property/InputLookupQA.test.js
@@ -1,4 +1,4 @@
-// Copyright 2018 Stanford University see LICENSE for license
+// Copyright 2019 Stanford University see LICENSE for license
 
 import 'jsdom-global/register'
 import React from 'react'
@@ -125,16 +125,6 @@ describe('<InputLookupQA />', () => {
       jest.restoreAllMocks()
     })
 
-    it('sets the default values according to the property template if they exist', () => {
-      const defaults = [{
-        id: 'http://id.loc.gov/vocabulary/carriers/nc',
-        uri: 'http://id.loc.gov/vocabulary/carriers/nc',
-        label: 'volume',
-      }]
-
-      expect(wrapper.state('defaults')).toEqual(defaults)
-    })
-
     it('logs an error when no defaults are set', () => {
       const plProps = {
         id: 'lookupComponent',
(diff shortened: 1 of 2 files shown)


src/components/editor/property/InputLookupQA.jsx
```diff
@@ -122,6 +114,13 @@ class InputLookupQA extends Component {
     let authority
     let language
     let subauthority
+    const defaults = defaultValuesFromPropertyTemplate(this.props.propertyTemplate)
+
+    if (defaults.length === 0) {
+      // Property templates do not require defaults but we like to know when this happens
+      console.info(`no defaults defined in property template: ${JSON.stringify(this.props.propertyTemplate)}`)
```
ndushay: i think this could get really noisy.  Maybe make it a debug?  (yes, i know it's what was there, but I'm looking at it now!  ;-)

__tests__/components/editor/property/InputLookupQA.test.js
```diff
@@ -120,16 +120,6 @@ describe('<InputLookupQA />', () => {
       jest.restoreAllMocks()
     })
 
-    it('sets the default values according to the property template if they exist', () => {
```
ndushay: is there a test showing this from the new approach?  If it's not approp for a unit test, make a ticket to make it an integration test?

src/components/editor/property/InputLookupQA.jsx
jcoyne: If we want to make a ticket to remove this logging, I'd certainly be supportive.  I don't think that change fits in well with the intent of this PR.

__tests__/components/editor/property/InputLookupQA.test.js
jcoyne: This is already covered by https://github.com/LD4P/sinopia_editor/blob/010f086f6901b3e218d78c97e0f176e76ae75db5/__tests__/components/editor/property/InputLookupQA.test.js#L152-L160

src/components/editor/property/InputLookupQA.jsx
ndushay: I guess I'm more of the school "if I'm looking at it now and thinking about it, might as well change it."  But you're not of that school, and that's ok.  This isn't anything I'd go to the mat for.

jmartin-sul: here's a tiny little PR that makes this change:  https://github.com/LD4P/sinopia_editor/pull/767

jmartin-sul: eh, eslint is complaining about `console.debug`.  i don't particularly care one way or the other about this issue.  happy to address per my comment on the PR, or people can close the PR, whatever.

ndushay: meh, i'm starting to believe we should stop logging this at all.

PR #767, opened 2019-06-18
when there are no defaults defined for a property template, log a debug level message to console

jmartin-sul:
instead of an info level message


jmartin-sul:
turns out eslint doesn't like the `console.debug`.  if people think this is a good fix, i can add `debug` to the allowed `console` log functions.  if people don't care or there's disagreement, feel free to close.  this was super quick anyway.

jmartin-sul:
closing in favor of #774

PR #768, opened 2019-06-18
Fixes bug where subsequently selected resource templates are appended to selectorReducer.resource

jgreben:
Selections of new resource templates from list should replace the state of `selectorReducer.resource`

### DIFF
--- a/__tests__/reducers/index.test.js
+++ b/__tests__/reducers/index.test.js
@@ -1,4 +1,5 @@
 // Copyright 2019 Stanford University see LICENSE for license
+
 import shortid from 'shortid'
 import selectorReducer, {
   populatePropertyDefaults,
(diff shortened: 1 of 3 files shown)


__tests__/reducers/index.test.js
```diff
@@ -314,26 +315,62 @@ describe('refreshResourceTemplate', () => {
     })
   })
 
-  it('missing reduxPath in payload should return the state', () => {
-    const missingPayload = refreshResourceTemplate({}, {
```
jgreben: AFAIKT there is not a way for the `refreshResourceTemplate` function to get a missing payload, and it would otherwise have to add extra lines of code to check for the existence of `Object.keys`

src/reducers/index.js
```diff
@@ -67,17 +68,19 @@ export const populatePropertyDefaults = (propertyTemplate) => {
 }
 
 export const refreshResourceTemplate = (state, action) => {
-  const newState = { ...state }
+  const rtId = Object.keys(state.resource).pop()
```
jgreben: This variable expects that there is an existing resource state, so I removed the test that (AFAICT) artificially creates a missing payload.

src/components/editor/property/ResourceProperty.jsx
```diff
@@ -65,7 +65,6 @@ export class ResourceProperty extends Component {
                                    propertyTemplate={rtProperty}
                                    reduxPath={propertyReduxPath}
                                    addButtonDisabled={isAddDisabled}
-                                   initNewResourceTemplate={this.props.initNewResourceTemplate}
```
jgreben: AFAICT this is no longer used in the `PropertyTemplateOutline` component.

__tests__/reducers/index.test.js
```diff
@@ -314,26 +315,62 @@ describe('refreshResourceTemplate', () => {
...
+    selectorReducer(initialState, {
+      type: 'SET_RESOURCE_TEMPLATE',
+      payload: {
+        id: 'resourceTemplate:bf2:Note',
+        resourceURI: 'http://id.loc.gov/ontologies/bibframe/Note',
+        propertyTemplates: samplePropertyTemplate,
+      },
+    })
+
+    const overwiteStateResult = refreshResourceTemplate(initialState.selectorReducer, {
```
jcoyne: Typo in "overwrite".

src/reducers/index.js
jcoyne: rtId is not a very expressive name.  How about `resourceTemplateId` instead?

src/reducers/index.js
```diff
@@ -67,17 +68,19 @@ export const populatePropertyDefaults = (propertyTemplate) => {
 }
 
 export const refreshResourceTemplate = (state, action) => {
```
jcoyne: Would you please add documentation about what this method is doing.

src/reducers/index.js
```diff
@@ -67,17 +68,19 @@ export const populatePropertyDefaults = (propertyTemplate) => {
...
-  const newState = { ...state }
+  const rtId = Object.keys(state.resource).pop()
+  const newResource = { resource: { [rtId]: state.resource[rtId] } }
```
jcoyne: I'm not familiar with this syntax (`[rtId]`).  does that evaluate the value of `rtId`?

jgreben: Yes, I just found out about it myself. It's the "Object Literal Syntax". See https://www.javascripttutorial.net/es6/object-literal-extensions/

jgreben:
Thanks for the 👀 's @jcoyne. I pushed a couple of commits to address the changes that you requested.

PR #769, opened 2019-06-18
eslint: alphabetize rules

ndushay:
it was getting ridiculous, and my OCD couldn't take it.


jmartin-sul:
my OCD made me copy the lists out and run them through a sort to compare them.  they're (of course) the same and this is green, so merging...

Issue #392, opened 2019-03-06
remove tooltip "please fill out this field" for required literal fields


michelleif:
in the Linked Data Editor, the "please fill out this field" tooltip that is being shown for required literal fields is leading to an inconsistent user experience.
1) if the profile creator has put their own tooltip in, the "please fill out this field" then doesn't show up, even if the field is required
2) we aren't displaying the "please fill out this field" tooltip for lookup-type fields that are required
3) we are using the red asterisk anyway to indicate required fields

PR #770, opened 2019-06-18
Fix empty tooltip and missing required tooltip

### LINKED ISSUE #757: Tooltip is empty
I'm seeing that "Example 17" on the Fuller Demo shows a tooltip that doesn't have any content.

Screen Shot 2019-06-18 at 12 16 55 PM

jgreben:
Fixes #757 and #392

### DIFF
--- a/src/components/editor/property/PropertyLabel.jsx
+++ b/src/components/editor/property/PropertyLabel.jsx
@@ -5,6 +5,7 @@ import React, { Component } from 'react'
 import { OverlayTrigger, Popover } from 'react-bootstrap'
 import PropTypes from 'prop-types'
 import RequiredSuperscript from './RequiredSuperscript'
+
 const _ = require('lodash')
 
 export class PropertyLabel extends Component {
--- a/src/components/editor/property/PropertyLabel.jsx
+++ b/src/components/editor/property/PropertyLabel.jsx
@@ -5,6 +5,7 @@ import React, { Component } from 'react'
 import { OverlayTrigger, Popover } from 'react-bootstrap'
 import PropTypes from 'prop-types'
 import RequiredSuperscript from './RequiredSuperscript'
+const _ = require('lodash')
 
 export class PropertyLabel extends Component {
   constructor(props) {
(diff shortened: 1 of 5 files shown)


src/components/editor/property/PropertyLabel.jsx
```diff
@@ -47,7 +47,7 @@ export class PropertyLabel extends Component {
       <span key={key}>{property.propertyLabel}</span>
     )
 
-    url !== undefined ? title.push(urlLabel) : property.remark !== undefined ? title.push(toolTipLabel) : title.push(plainLabel)
+    url !== undefined ? title.push(urlLabel) : property.remark !== undefined && property.remark !== '' ? title.push(toolTipLabel) : title.push(plainLabel)
```
jcoyne: can you use ` _.isEmpty(property.remark)`?

jgreben:
> Screen Shot 2019-06-19 at 8 48 04 AM
> 
> This doesn't seem easy to read. I don't think we want both a title and a popover menu. Maybe @astridu can help?

I made this change pending @astridu's feedback:
Screen Shot 2019-06-19 at 11 51 29 AM

astridu:
Yes, the second example looks much better (just the 1 tooltip). Thank you!

jcoyne:
I'm seeing this now.  Since we have the tooltip, should we get rid of the `<abbr/>`?

Screen Shot 2019-06-20 at 9 24 45 AM

jcoyne:
@michelleif @jermnelson what sort of commitments to we have to accessibility?  I'm pretty sure the `<abbr>` with a title are useful for screen readers.  I'm not sure the popover is useful for anyone not using a mouse.

ndushay:
would it be stupid to add a css styling that doesn't do the popover (because we display it another way?)

jermnelson:
> @michelleif @jermnelson what sort of commitments to we have to accessibility? I'm pretty sure the `<abbr>` with a title are useful for screen readers. I'm not sure the popover is useful for anyone not using a mouse.

I would like us to be more accessible. We're running this elint plugin: https://www.npmjs.com/package/eslint-plugin-jsx-a11y with some rule exceptions.

jcoyne:
So do we know if a screenreader can focus on a `<sup>` element that has a single child marked `aria-hidden="true"`?

jgreben:
I added a tabIndex to the `sup` to make it keyboard accessible, and role-tooltip for ARIA. Do you  think this will address the accessibility issues?

jcoyne:
Yes, that probably ought to do it.

PR #773, opened 2019-06-19
Export language tags to RDF

### LINKED ISSUE #666: Language tag should be transformed to RDF
Currently, when "preview" is clicked, the language tag isn't displayed.

jcoyne:
This involved switching to using iso639-1 language codes because these are required for language tags per RFC 5646.

Fixes #666

### DIFF
--- a/src/GraphBuilder.js
+++ b/src/GraphBuilder.js
@@ -89,11 +89,8 @@ export default class GraphBuilder {
 
       if (value.items) {
         for (const item of value.items) {
-          if (item.uri) {
-            this.dataset.add(rdf.quad(baseURI, rdf.namedNode(predicate), rdf.namedNode(item.uri)))
-          } else {
-            this.dataset.add(rdf.quad(baseURI, rdf.namedNode(predicate), rdf.literal(item.content)))
-          }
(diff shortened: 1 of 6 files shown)


src/GraphBuilder.js
```diff
@@ -107,6 +104,18 @@ export default class GraphBuilder {
     }
   }
 
+  /**
+   * Returns a literal with an optional language tag
+   * @param {Object} item from the redux store
+   * @return {rdf.LiteralExt} the literal with a language value
+   */
+  createLiteral(item) {
+    if (item.lang) {
```
mjgiarlo: If `item.lang` is truthy, is it a safe bet that it has an `items` array of length > 0 that contains things with an `id`? If so, this is fine. If we're less sure, we can do something like this and ditch the condition:

```javascript
return rdf.literal(item.content, item?.lang?.items?.[0]?.id)
```

jcoyne: Yes, we're sure of the shape.   But I did revise this method to use the safe navigation.

ndushay:
note that ^^ is exactly the sort of integration test I would want to see accompany this PR, once we have that stuff in place.

PR #774, opened 2019-06-19
don't bother logging to console when there are no defaults defined for a property template

jmartin-sul:
kinda noisy, easy to add back if we want it again

supersedes #767


jmartin-sul:
merging since @mjgiarlo approved...

ndushay:
🎉 🎊 🎈

PR #775, opened 2019-06-19
Separate selectors from reducers

jcoyne:


jcoyne:
@ndushay that's it.

src/selectors/resourceSelectors.js
```diff
@@ -1,3 +1,40 @@
 export const rootResource = state => Object.values(state.selectorReducer.resource)[0]
```
jermnelson: This module needs a copyright statement.

jmartin-sul: created #775.

is it ok to merge this one in light of that?

jermnelson: Sure, resolving now.

PR #776, opened 2019-06-19
Document the setItemsOrSelections reducer function

jgreben:
@jermnelson please have a look and make sure that I described this all correctly.

### DIFF
--- a/src/reducers/inputs.js
+++ b/src/reducers/inputs.js
@@ -65,6 +65,13 @@ export const removeAllContent = (state, action) => {
   return validate(newState)
 }
 
+/**
+ * Takes the reduxPath (an array of keys that correspond to the redux state tree for a 'resource') and performs a reduce
+ * function on each of the keys, searching for the last key in the path and then appending an object with an `items` array.
+ * Also checks for needed blank nodes along the reduxPath in the state tree and appends intermediate objects.
+ *
+ * @returns a {function} with {{}}
+ */
 export const setItemsOrSelections = (state, action) => {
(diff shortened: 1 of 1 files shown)


jgreben:
This used to be two separate functions: `setMyItems` and `changeMySelections` that did the same things within the reducer method. These were combined into one function with the following forking:

```
      if (action.type === 'SET_ITEMS') {
        action.payload.items.map((row) => {
          obj[key].items.push(row)
        })
      }
      else if (action.type === 'CHANGE_SELECTIONS') {
        obj[key].items = action.payload.items
      }
```

If it makes more sense to people to have two separate functions despite mostly duplicative code, we can change it back. See https://github.com/LD4P/sinopia_editor/pull/734/files

src/reducers/inputs.js
```diff
@@ -65,33 +65,53 @@ export const removeAllContent = (state, action) => {
   return validate(newState)
 }
 
+/**
+ * Takes the reduxPath (an array of keys that correspond to the redux state tree for a 'resource') and performs a reduce
+ * function on each of the keys, searching for the last key in the path and then appending an object with an `items` array.
+ * Also checks for needed blank nodes along the reduxPath in the state tree and appends intermediate objects.
+ *
+ * @returns a {function} with {{}}
```
mjgiarlo: Should this be:

```javascript
 * @returns {function} a function with? that... ?
```

jgreben: Thanks. Not really sure how to write these things...LMK if it is better...

Issue #603, opened 2019-05-31
Support xsd:anyURI constraints in InputLiteral


jermnelson:
We're currently not supporting the `propertyTemplate.valueConstraint.valueDataType` property in Resource Templates that allows the user to define the literal's dataType like `xsd:dateTime`, `xsd:string`, `xsd:decimal` (see W3C [link](https://www.w3.org/TR/2004/REC-xmlschema-2-20041028/#built-in-primitive-datatypes) for more information) including `xsd:anyURI`. For the MVP we want to at least be able to differentiate between an URI verses an arbitrary string value that is entered by the user as the RDF node will be different. 

This could be as simple as parsing the value as a Javascript [URI](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL) and catching an exception to generate a string literal node. 

```javascript

try {
  objectNode = URI(value)
catch (e) {
 if (e instanceof TypeError) {
  objectNode = value
 }
}
```

justinlittman:
For going beyond just `xsd:anyURI` as described above, are you aware of any JS library that handles type conversion / validation? Having to do that ourselves seem like a bit of work.

jermnelson:
Yes, I agree that doing this all ourselves would be a pain. I am not aware of any specific library that does this type validation or conversion but since we're already using the [ajv](https://github.com/epoberezkin/ajv) module for Profile and Resource Template validation, it looks like we may be able to use ajv to both validate the type and to [coerce](https://ajv.js.org/#coercing-data-types) it into the type we want.

jgreben:
@jermnelson I think you are saying that the profile editor currently only allows the user to choose or enter a URL in the Value Data Type field and not a string. So are we still expecting templates where the user has manually edited the JSON profile to include a string as the `propertyTemplate.valueConstraint.valueDataType` value?

jcoyne:
@jermnelson Do we have to trigger URI types by `propertyTemplate.valueConstraint.valueDataType === 'xsd:anyURI'`  could we have `propertyTemplate.type === 'resource'` instead?

jgreben:
@jermnelson I am agreeing with @jcoyne because it seems counterintuitive if there is an "InputLiteral" field that is not actually type literal, but instead a resource, and has a URI and not a literal as the input.

PR #784, opened 2019-06-19
Display a message if a property component is misconfigured as a resource

jgreben:
I came across this while testing various sever templates for the demo:

When `type: 'resource'` without `valueTemplateRefs`

Before:
Screen Shot 2019-06-19 at 3 59 19 PM

After:
Screen Shot 2019-06-19 at 3 59 26 PM

### DIFF
--- a/__tests__/components/editor/property/PropertyComponent.test.js
+++ b/__tests__/components/editor/property/PropertyComponent.test.js
@@ -100,6 +100,22 @@ describe('<PropertyComponent />', () => {
       expect(wrapper.find('Connect(InputLookupQA)').length).toEqual(0)
       expect(wrapper.find('Connect(InputLiteral)').length).toEqual(0)
     })
+
+    it('returns a warning message if the property type is not literal (i.e. resource)', () => {
+      const template = {
+        propertyURI: 'http://id.loc.gov/ontologies/bibframe/note',
+        type: 'resource',
+        valueConstraint: {
+          valueTemplateRefs: [],
+          useValuesFrom: [],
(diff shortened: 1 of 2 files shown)


astridu:
I agree. Here are 2 options. The first one is just text in the warning color of dark orange. The second one uses fontawesome if it's readily available to you. Either one will work. thank you! 
image

jgreben:
> Is that symbol the same as the required field symbol? If so, I suspect we will serve users better if we change it .. is there a good "warning" symbol?
> 
> @michelleif @astridu

The asterisk is there because the template is marked as `mandatory: true` I added the font-awesome icon to the alert.

ndushay:
haha - silly to have a required screwed up template.

jcoyne:
I've proposed that this configuration means, "enter a URL into a text field" https://github.com/LD4P/sinopia_editor/issues/603#issuecomment-503686604
So, this would not be a bad configuration. I'm working on this in https://github.com/LD4P/sinopia_editor/pull/789

michelleif:
The Library of Congress uses this configuration (type = resource, no resource template specified) but we don't understand what they are trying to achieve by it.

ndushay:
@michelleif I suspect they are achieving "whoops - we shouldn't allow that to happen"

jgreben:
I will close this PR for the time being. Hopefully we can get #789 vetted and merged before Sunday...

ndushay:
IMO, this PR is an improvement;  I would be ok merging this and creating a ticket for remaining open questions.

PR #792, opened 2019-06-20
Simplify removing an item.

jcoyne:
We only need to pass reduxPath and item ID

### DIFF
--- a/src/components/editor/property/InputLiteral.jsx
+++ b/src/components/editor/property/InputLiteral.jsx
@@ -91,17 +91,7 @@ export class InputLiteral extends Component {
   }
 
   handleDeleteClick = (event) => {
-    const labelToRemove = event.target.dataset.content
-    const idToRemove = event.target.dataset.item
-
-    this.props.handleRemoveItem(
-      {
-        id: idToRemove,
-        label: labelToRemove,
-        reduxPath: this.props.reduxPath,
-        uri: this.props.propertyTemplate.propertyURI,
-      },
-    )
+    this.props.handleRemoveItem(this.props.reduxPath, event.target.dataset.item)
   }
 
   handleEditClick = (event) => {
@@ -285,8 +275,8 @@ const mapDispatchToProps = dispatch => ({
   handleMyItemsChange(userInput) {
     dispatch(setItems(userInput))
   },
-  handleRemoveItem(id) {
-    dispatch(removeItem(id))
+  handleRemoveItem(reduxPath, itemId) {
+    dispatch(removeItem(reduxPath, itemId))
   },
   handleMyItemsLangChange(payload) {
     dispatch(setLang(payload))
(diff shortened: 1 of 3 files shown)


src/components/editor/property/InputLiteral.jsx
```diff
@@ -91,17 +91,7 @@ export class InputLiteral extends Component {
...
-    const labelToRemove = event.target.dataset.content
-    const idToRemove = event.target.dataset.item
-
-    this.props.handleRemoveItem(
-      {
-        id: idToRemove,
-        label: labelToRemove,
-        reduxPath: this.props.reduxPath,
-        uri: this.props.propertyTemplate.propertyURI,
-      },
```
ndushay: why did this object have all these vars before?   label and uri?

src/components/editor/property/InputLiteral.jsx
```diff
@@ -91,17 +91,7 @@ export class InputLiteral extends Component {
...
   handleDeleteClick = (event) => {
-    const labelToRemove = event.target.dataset.content
-    const idToRemove = event.target.dataset.item
-
-    this.props.handleRemoveItem(
-      {
-        id: idToRemove,
-        label: labelToRemove,
-        reduxPath: this.props.reduxPath,
-        uri: this.props.propertyTemplate.propertyURI,
```
ndushay: why were label and uri here before?  were they never used? 

Putting it another way -- we had no test anywhere showing why/how these were used?  If so, please write a ticket to write such a test (showing functionality of the delete, prolly as an integration test)

src/components/editor/property/InputLiteral.jsx
```diff
@@ -285,8 +275,8 @@ const mapDispatchToProps = dispatch => ({
   handleMyItemsChange(userInput) {
     dispatch(setItems(userInput))
   },
-  handleRemoveItem(id) {
-    dispatch(removeItem(id))
+  handleRemoveItem(reduxPath, itemId) {
+    dispatch(removeItem(reduxPath, itemId))
```
ndushay: my spotting redux and sinopia-redux knowledge says. so we need the redux path so we can scope the item id and/or ensure we are hooking up to the correct place in the nested resource template madness.  true?

src/components/editor/property/InputLiteral.jsx
jcoyne: They were never used. That's why I got rid of them, because it's confusing that we're suppling them.

src/components/editor/property/InputLiteral.jsx
jcoyne: yes, the reduxPath and id in combination is the address of the data element to remove

src/components/editor/property/InputLiteral.jsx
jermnelson: I think initially we searched based on label of the item. This is no longer needed in PR.

ndushay:
created #794 about integration test needed.  Merging.

Issue #630, opened 2019-06-05
Create Search Express route in Editor for ElasticSearch queries


jermnelson:
Instead of the Editor making direct calls to an open-to-the-world Elasticsearch instance, create an Express server-side route in the [server.js](https://github.com/LD4P/sinopia_editor/blob/master/server.js) that interacts with our AWS Elasticsearch instance on the server-side instead of in the React app on the client-side. We could look at using the official ES node.js [client](https://github.com/elastic/elasticsearch-js).

jcoyne:
Does this server need a swagger API description?

jermnelson:
> Does this server need a swagger API description?

The Swagger API json we're using for QA is here: https://github.com/LD4P/sinopia_editor/blob/master/src/lib/apidoc.json

jermnelson:
> > Does this server need a swagger API description?
> 
> The Swagger API json we're using for QA is here: https://github.com/LD4P/sinopia_editor/blob/master/src/lib/apidoc.json

On some further reflections, I think for the MVP all we need is support are two searchrd:
1. For the type-ahead in Elasticsearch  (called a Suggester Completion in https://www.elastic.co/guide/en/elasticsearch/reference/current/search-suggesters-completion.html as a  keyword search. I have used variations for a typeahead search using Python elasticsearch clients in past projects. 
1. General search https://www.elastic.co/guide/en/elasticsearch/reference/current/search-uri-request.html For the basic search on the Search tab.

mjgiarlo:
Blocked by #800?

Issue #688, opened 2019-06-12
Dereferenced URI for RDF Resources


jermnelson:
In the current master branch, saving a RDF graph to Trellis generates a URI depending on what environment:

- (locally) http://platform:8080/repository/pcc/ed46b285-1979-440d-83cb-7ca46318c89f
- (dev) https://trellis.development.sinopia.io/repository/stanford/ef23778a-781d-49fa-8320-64fb882098f3
- (prod) https://trellis.sinopia.io/repository/stanford/ef23778a-781d-49fa-8320-64fb882098f3

We want to publish these resources on Development, Stage, and Production by replacing the Trellis URI and `/repository` path with just URI of the environment and the group name like these examples:

`https://trellis.development.sinopia.io/repository/stanford/ef23778a-781d-49fa-8320-64fb882098f3` is published as
 `https://development.sinopia.io/stanford/ef23778a-781d-49fa-8320-64fb882098f3`
or on Production like
`https://trellis.sinopia.io/repository/stanford/ef23778a-781d-49fa-8320-64fb882098f3` becomes `https://sinopia.io/stanford/ef23778a-781d-49fa-8320-64fb882098f3`

I think there a number of ways we can accomplish this including:

- Express route 
- React BrowserRouter
- Trellis slug
- Sinopia Server client change

For now, I think we can return whatever Trellis is returns when the Trellis URI is accessed directly. For further discussion with the Devs, we may want to use this de-referenced URI to display in the Editor tab loading the data through best approximate Resource Templates.

ndushay:
Based on discussions today b/t John, Aaron, Jeremy and Naomi:



Assumptions:
- express is the web server, both for local dev and deployed in AWS
- we can white list expected group names
- we can postpone displaying retrieved resource in context of the editor when a cognito JWT is passed in ... 

Work Pieces:
1.  resource url functions (in Utilities?):
- canonicalResourceUrl(url)  returns the shorter, canonical form of url passed in.  If canonical form passed in, return it.
- expandedResourceUrl(url)  returns the longer form of url, dns preceded with "trellis" or whatever and "repository" first part of path

2. get a list of known group names from Sinopia Server

3. given list of known group names from Sinopia Server, have express app's `server.js` trap for canonical routes and translate them to the appropriate expanded resourceURL ... then do fetch from sinopia server and return whatever the Accept header indicated (default to json-ld).  Presumably trellis will return json-ld or ttl or rdfxml if we ask for it properly.

[...]

PR #793, opened 2019-06-20
Use webpack to build both dev and prod servers

mjgiarlo:
And use webpack-dev-middleware instead of webpack-dev-server, so that the dev server can make use of the same Express server used in prod. This brings dev and prod configurations closer together.

Refs #630

This is the first step in tackling #630, exposing a ElasticSearch proxy via the application express server. This work aligns our production server infrastructure with dev server infrastructure, making sure that when the proxy is built, it is exposed identically in both dev and prod.


ndushay:
FYI:  This is currently tabled, but our implementation thoughts expected express server:  https://github.com/LD4P/sinopia_editor/issues/688.  Possibly worth a discussion?

mjgiarlo:
@ndushay 	See revised description above!

PR #795, opened 2019-06-20
Clear validation errors when a resource template is loaded

### LINKED ISSUE #745: Errors not being cleared when new Resource Template is loaded
When a resource template is loaded and validation errors are in Redux following the user clicking on the **Save &amp; Publish**, if the user loads a second resource template, the errors are still present in the new form.

Screen Shot 2019-06-17 at 12 17 12 PM

Now, loading a second resource template, the errors from the first resource template are still present:

Screen Shot 2019-06-17 at 12 18 19 PM

This is likely related to https://github.com/LD4P/sinopia_editor/issues/651

aaron-collier:
Fixes #745 

Clears the errors and sets displayValidations to false when a resource template is loaded. Adds unit testing for the values and an integration test for displaying and clearing the validation errors on the page.

Before: After validation errors and clicking on a new template - Validation errors are still displayed. 

Screen Shot 2019-06-20 at 11 14 57 AM

After:  After validation errors and clicking on a new template - Validation errors are cleared.

Screen Shot 2019-06-20 at 11 15 28 AM

### DIFF
--- a/__tests__/integration/previewRDFHelper.js
+++ b/__tests__/integration/previewRDFHelper.js
@@ -31,14 +31,4 @@ export async function fillInRequredFieldsForBibframeInstance() {
 export async function incompleteFieldsForBibframeInstance() {
   // This assertion adds 1 to each it blocks assertion count
   await pupExpect(page).toClick('a', { text: 'BIBFRAME Instance' })
-
-  // Click on one of the property type rows to expand a nested resource
-  await page.waitForSelector('a[data-id=\'title\']')
-  await page.click('a[data-id=\'title\']')
-  await page.waitForSelector('a[data-id=\'mainTitle\']')
-
-  // Fill in required element
-  await page.click('a[data-id=\'mainTitle\']')
-  await page.type('[placeholder=\'Preferred Title for Work (RDA 6.2.2, RDA 6.14.2) (BIBFRAME: Main title)\']', 'Hello')
-  await page.keyboard.press('Enter')
 }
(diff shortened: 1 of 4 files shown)


__tests__/integration/previewRDFHelper.js
```diff
@@ -27,3 +27,18 @@ export async function fillInRequredFieldsForBibframeInstance() {
   await page.waitForSelector('#rbt-menu-item-0')
   await page.click('#rbt-menu-item-0')
 }
+
+export async function incompleteFieldsForBibframeInstance() {
+  // This assertion adds 1 to each it blocks assertion count
+  await pupExpect(page).toClick('a', { text: 'BIBFRAME Instance' })
+
+  // Click on one of the property type rows to expand a nested resource
```
jcoyne: do we need the rest of this function?

aaron-collier: Oh yeah, probably not. That's just how I was manually testing it. Will update.

aaron-collier: Done. Good catch.

__tests__/integration/previewRDFHelper.js
```diff
@@ -27,3 +27,8 @@ export async function fillInRequredFieldsForBibframeInstance() {
   await page.waitForSelector('#rbt-menu-item-0')
   await page.click('#rbt-menu-item-0')
 }
+
+export async function incompleteFieldsForBibframeInstance() {
+  // This assertion adds 1 to each it blocks assertion count
```
ndushay: PR looks fine but I don't understand this comment.

aaron-collier: I'm not great and clarifying this, but this is the assertion that means any block "it" block that includes this call needs to check for it's assertions+1.

ndushay: ah!  how about

```
// this assertion adds 1 to each "it" block's expect.assertion(n) count
```

PR #799, opened 2019-06-20
Add an aria-lable so a screen reader will understand the sup tag

jcoyne:


mjgiarlo:
Can you `s/lable/label/` in the commit message as well?

jmartin-sul:
d'oh!  sorry, missed @mjgiarlo's comments.
> Is the `tabIndex="0"` not needed?

i had the same question originally, but then i figured if the component was repeated, having a hardcoded tab index wouldn't be useful?

apologies if i jumped the gun on this merge!

jcoyne:
I talked to @metageeky on slack and she pointed me in this direction.

jmartin-sul:
> I talked to @metageeky on slack and she pointed me in this direction.

ah, ok.  if the community's accessibility expert approves, that's good enough for me! 😁

PR #803, opened 2019-06-21
Revert "Use webpack to build both dev and prod servers"

ndushay:
Reverts LD4P/sinopia_editor#793

I think this is breaking builds locally for @jmartin-sul @mjgiarlo and myself.  I also think this PR could/should be split out into at least 3 separate PRs:

- one for webpack work
- one for moving images files out of styles folder
- one for adding stuff to docker-compose


ndushay:
Note:  I am STILL unable to get THIS branch to run tests cleanly on my machine.  Is there a way some local stuff needs to be reset for tests to pass locally?

ndushay:
the error is in finding the sinopia server locally for tests.  Not sure why circleci is happy but our laptops aren't.

PR #809, opened 2019-06-21
simplify fixture loading

### LINKED ISSUE #785: load all fixtures in fixtures directory except bad ones
following #763 - it would be great to automagically add fixture data for tests and development mode by just adding them to the fixtures directory.  

- due to javascript not being allow to access the filesystem when running in a browser, the files MUST be read via "require"
- we may be able to do one of these things:
    - use the require-all npm package
    - add an index.js file to the fixtures folder to do the requires (?)
- there are some files that should NOT be loaded.  They could be renamed to follow a particular pattern (like "noload_xxx")  -- these fixtures are used for schema validation testing

```
const doNotAutoloadFixtures = [
  'ddc_bad_json.json',
  'edition_bad_schema',
  'lcc_no_schema_specified',
  'lcc_v0.0.9.json',
  'lcc_v0.2.0.bad_id.json',
  'lcc_v0.2.0_invalid.json',
  'lcc_v0.2.0.json',
  'Note(Profile).json', // detritus from "spoof" that could be removed?
  'place_profile_no_schema_specified',
  'place_profile_v0.0.9.json',
  'place_profile_v0.2.0.json',
]
```

- ~~the unit tests in `rtFixtureLoaderHelper.test.js` no longer execute due to exclusion pattern for test files in package.json.~~  I'm not convinced we need or want these unit tests.
    - removing these tests would let us reduce the number of exports in `rtFixtureLoaderHelper` to 3 -- the ones used by sinopiaServer

ndushay:
Closes #785 

Simplifies code for loading our resourceTemplate fixtures, and what is needed to add a new fixture.

Note that this represents my best pragmatic solution to be able to load specific files from the file system in a browser context.


.eslintrc.js
```diff
@@ -48,6 +48,7 @@ module.exports = {
     "brace-style": "off",
     'capitalized-comments': "off",
     "class-methods-use-this": "off",
+    "comma-dangle": "off",
```
jcoyne: I don't think this change ought to be incorporated in the PR, unless you want to bog this down with a style discussion.

package.json
```diff
@@ -80,8 +80,8 @@
     "dev-build": "webpack --progress --mode development",
     "dev-build-test": "npm run dev-build && npm run test",
     "dev-start": "webpack-dev-server --config ./webpack.config.js --mode development",
-    "eslint": "eslint --max-warnings 10 --color -c .eslintrc.js --ext js,jsx .",
-    "eslint-fix": "eslint --max-warnings 10 --color -c .eslintrc.js --ext js,jsx --fix .",
+    "eslint": "eslint --max-warnings 11 --color -c .eslintrc.js --ext js,jsx .",
```
jcoyne: Why are you bumping this threshold?

.eslintrc.js
jmartin-sul: i'm personally neutral on this rule.  not wild about the way it reads when i'm looking at code, but it does lead to slightly nicer diffs because it means just a little bit less churn, since the last list entry is no longer special in terms of not having a comma after it.

would agree w/ @jcoyne's point that it warrants discussion, since i suspect some might want this rule?  when i ran into it in the past, i just played along with the restriction.

package.json
jmartin-sul: was wondering that myself

.eslintrc.js
ndushay: I can up the warnings and do it separately.  It actually creates invalid json for our inline json and for json objects.

package.json
ndushay: b/c I have another non-literal in a require statement.

ndushay:
@jcoyne @jmartin-sul I took out the eslint rule change;  I had to up the warnings to 11 because the new fixtureLoader logic does a non-literal-require rule violation (we already have a couple of those for versoSpoof so it's nothing new).  Should be ready to merge.

package.json
jcoyne: Can you add the eslint ignore pragma to that file instead?  I think the warnings we are listing out here are getting too noisy, so I don't want to make more of them.

__tests__/fixtureLoaderHelper.js
```diff
@@ -1,57 +1,63 @@
...
+  'adminMetadataStatus.json',
+  'rdaItemMonograph.json',
 ]
 
+export const resourceTemplateId2Json = loadFixtureResourceTemplates()
+
+function loadFixtureResourceTemplates() {
+  const result = []
+  rtFileNames.forEach((filename) => {
+    /* eslint security/detect-non-literal-require: 'off' */
```
ndushay: now with pragma!

package.json
```diff
@@ -80,8 +80,8 @@
     "dev-build": "webpack --progress --mode development",
     "dev-build-test": "npm run dev-build && npm run test",
     "dev-start": "webpack-dev-server --config ./webpack.config.js --mode development",
-    "eslint": "eslint --max-warnings 10 --color -c .eslintrc.js --ext js,jsx .",
-    "eslint-fix": "eslint --max-warnings 10 --color -c .eslintrc.js --ext js,jsx --fix .",
+    "eslint": "eslint --max-warnings 7 --color -c .eslintrc.js --ext js,jsx .",
+    "eslint-fix": "eslint --max-warnings 7 --color -c .eslintrc.js --ext js,jsx --fix .",
```
ndushay: now reduced to 7!

src/versoSpoof.js
```diff
@@ -13,7 +13,9 @@ function loadProfiles() {
   if (profiles.length === 0) {
     const profilesDirPath = path.join(__dirname, '..', 'static', 'spoofedFilesFromServer', 'from_verso', 'data', 'profiles')
 
+    /* eslint security/detect-non-literal-fs-filename: 'off' */
     fs.readdirSync(profilesDirPath).forEach((file) => {
+      /* eslint security/detect-non-literal-require: 'off' */
```
ndushay: these pragmas remove 3 eslint warnings

PR #810, opened 2019-06-21
fix circleci config typos

ndushay:
oops!!!!!


ndushay:
merging my own here:
1. it's only a change to circleci register-image step
2. it's my typo.

PR #815, opened 2019-06-21
Use thunks for loading the root resource template

jcoyne:
When we load the resource from Trellis, we need a way to load resourceTemplates
So first we'll convert the existing system to use thunks

Thunks take the loading logic out of the components

### DIFF
--- a/src/actionCreators.js
+++ b/src/actionCreators.js
@@ -1,3 +1,5 @@
+// Copyright 2019 Stanford University see LICENSE for license
+
 import {
   authenticationFailure, authenticationSuccess, signOutSuccess,
   updateStarted, updateFinished,
--- a/src/actionCreators.js
+++ b/src/actionCreators.js
@@ -1,7 +1,10 @@
 import {
-  authenticationFailure, authenticationSuccess, signOutSuccess, updateStarted, updateFinished,
+  authenticationFailure, authenticationSuccess, signOutSuccess,
+  updateStarted, updateFinished,
+  retrieveResourceTemplateStarted, rootResourceTemplateLoaded,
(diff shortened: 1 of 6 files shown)


src/actionCreators.js
```diff
@@ -1,7 +1,10 @@
 import {
```
jermnelson: Needs Copyright statement.

src/actionCreators.js
```diff
@@ -21,3 +26,13 @@ export const update = currentUser => (dispatch, getState) => {
   return updateRDFResource(currentUser, uri, rdf)
     .then(response => dispatch(updateFinished(response)))
 }
+// A thunk that retrieve the root resource template
```
ndushay: typo "retrieves"

src/components/editor/ResourceTemplate.jsx
```diff
@@ -1,50 +1,31 @@
...
-    if (_.isEmpty(this.props.resourceTemplate)) {
-      let errorMessage = <span/>
+    let errorMessage = <span/>
 
-      if (this.state.error) {
-        errorMessage = <div className="alert alert-warning">Sinopia server is offline or has no resource templates to display</div>
-      }
-      return errorMessage
+    if (this.props.error) {
+      errorMessage = <div className="alert alert-warning">Sinopia server is offline or has no resource templates to display.</div>
```
ndushay: is this error message still correct?  Is this component loading the root template or any resource template?  (and if it's template singular, is should this error message be singular too?  Is it more like "Sinopia server is offline or unable to retrieve resource template (id of resource template)" ?

Issue #423, opened 2019-03-18
RDF expectations from a given RT


michelleif:
what RDF is expected from a given RT (and its Property Templates)?

michelleif:
assuming the Resource Templates **Sinopia:RT:Fixture:SampleResource** and **Sinopia:RT:Fixture:Suppl** found in the attached Profile
**Sinopia:Profile:Fixture**, and the following data entered/selected by the cataloger in the following Linked Data Editor fields:

**Example 2**: cataloger enters 2010
**Example 10**: cataloger searches `coffee` and selects "Coffee in Literature" from the results
**Example 17**: cataloger opens the referenced template (**Sinopia:RT:Fixture:Suppl**) and fills in its two fields:
	**Supplementary Content**: `this is supplemental`
	**URI for Supplementary Content**: `http://www.supplement.com`

Then the RDF output should be the following (where xyz is the suffix of the minted Sinopia URI).
```
<http://sinopia.io/group/xyz> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://id.loc.gov/ontologies/bibframe/Work> .

<http://sinopia.io/group/xyz> <http://id.loc.gov/ontologies/bibframe/copyrightDate> "2010" .

<http://sinopia.io/group/xyz> <http://id.loc.gov/ontologies/bibframe/subject> <http://id.loc.gov/authorities/subjects/sh85027716> .

<http://sinopia.io/group/xyz> <http://id.loc.gov/ontologies/bibframe/supplementaryContent> _:b1 .

_:b1 <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://id.loc.gov/ontologies/bibframe/SupplementaryContent> .

_:b1 <http://www.w3.org/2000/01/rdf-schema#label> "this is supplemental" .

_:b1 <http://id.loc.gov/ontologies/bflc/target> "http://www.supplement.com" .
```
if we wanted to mint a URI for the Supplementary Content node... then replace `_:b1` with that minted URI

NOTES:
-haven't included RDF expectations for defaults yet. the logic is: if user doesn't do anything to the default value, output the default value. If user changes default value, only output the changed value.
-also have not included RDF expectations for repeatable fields.
-Also did not include any admin metadata that we plan to output automatically, like date the info was saved

michelleif:
@jermnelson can you check this to see if this meets the team's need for specification of RDF output expectations from Editor?

michelleif:
[...]

PR #816, opened 2019-06-21
correct rdf "acceptance" tests for literals

ndushay:
connects to #423

This is the first of the "basic building blocks" integration tests to ensure we've got the RDF from simple propertyTemplates correct.

Note that this is NOT an end-to-end test as it does not write to trellis and ensure that trellis has the 3 expected triples for the resource created.

I am deliberately using very small, targeted resourceTemplates for these.  I believe this approach will make it easier to hone in on problems for maintenance.

As is obvious from the "todo" tests, the intent is to write the rest of these for diff wrinkles with "literal" propertyTemplates.   After that, "lookup" propertyTemplates.


__tests__/integration/correctRdfInPreview/literalPropertyTemplates.test.js
```diff
@@ -0,0 +1,39 @@
...
+    expect.assertions(7)
+    await pupExpect(page).toClick('a[href="/editor"]', { text: 'test literal, not repeatable, required, no default' })
+    await pupExpect(page).toFill('input[placeholder=\'literal, not repeatable, required, no default\']', 'splendid')
+    await page.keyboard.press('Enter')
+    await expect(page).toClick('button', { text: 'Preview RDF' })
+    const previewRdf = await page.$eval('pre', e => e.textContent)
+    expect(previewRdf).toMatch('<> <http://examples.org/bogusOntologies/literal1> "splendid"@en .')
+    expect(previewRdf).toMatch('<> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://examples.org/bogusOntologies/Resource> .')
+    expect(previewRdf).toMatch('<> <http://www.w3.org/ns/prov#wasGeneratedBy> "Sinopia:RT:Fixture:LiteralNoRepeatNoDefault" .')
+    // ensure we have exactly 3 triples
```
jcoyne: Why do we need to see if we have exactly 3 triples?  Is there a requirement about this?

ndushay: because we are generating incorrect RDF if there are additional triples.

jcoyne: I'm not sure that adding a rdfs:label or some other type assertion would make this incorrect.  Furthermore, I think a test with these kind of attention to detail belongs at the component level.  Could it go in the GraphBuilder.test.js for example?  We don't want to have 10-12 integration tests that are all going to count the correct number of triples.  If we did, and we made a change (added a type triple for example), then we'd have to change a whole bunch of different tests.

ndushay: I'm not sure I agree with you.  At some point, we need to have integration tests that say that "given this resource template, we expect exactly these triples" and the easiest way to do that is to have specific tests for each wrinkle of possible resource templates.  So even if this was moved to graph builder, I'd still want to write the same number of tests and the same "I expect this number of triples exactly" assertion for each of them.  Moreover, this test is not just about graph builder, but about "given a particular resource template and particular user input, I get the RDF I expect" ... testing that at the component level seems wrong to me because it is not showing that the app is working properly -- that the input into a text box in the form gets into the RDF in exactly the right way.

Basically, this is the first of maybe 10-12 building block integration tests that will help pinpoint changes, in addition to the broader tests Michelle outlined in #423, which I intend to pursue after the building block tests are done.

I think making a change to the rdf generated SHOULD affect a bunch of different tests.  It's the absolutely core concept of this whole app:  what triples are generated from what user inputs associated with which resource templates.  Getting that exactly correct per our product owner feels right on the money, to me.

ndushay: it may be that having component level tests that check that graphbuilder is building the right triples from the redux store are *also* a good idea.

jcoyne: > specific tests for each wrinkle
This to me is the role of unit tests.   Integration tests should just ensure the parts are fitting together in the right way.

ndushay: These are more like "acceptance tests" in a sense - an automated way to do a bunch of tests from user input into whatever our UI looks like + resource template to exact RDF produced.  This PR is sort of prep for more complex examples from Michelle in #423 and maybe even #421.  I'm essentially working with stupidly simple resource templates to do the baby steps from her doc here:  https://docs.google.com/spreadsheets/d/1V0yNUMBV5W4Hv1KU455tq4-APW190hUWuqBbyrZYVls/edit#gid=0.  And then my intent is to do the more complex stuff in #423.

Having tests to show we have "correct" redux state and that we're building the graph "correctly" from the redux state are both worth doing, and I suspect we are shy on those tests as well.  In fact, perhaps a lot of the current unit tests could/should morph into just this, AND we should be testing more wrinkles.

mjgiarlo: @ndushay @jcoyne My $0.02: insufficient integration testing has really bitten us on this project. I'm OK with these changes going forward for now, and if we find these integration tests are not meeting our needs (bc overly specific or bc not covering the right behavior), we can always make changes later.

aaron-collier: Agree with @mjgiarlo.

ndushay:
Made ticket #835 re: choosing a testing philosophy

jmartin-sul:
per discussion in standup the last couple of days, the consensus seems to be that we will merge this PR, but will also have some follow on discussion next week about our testing philosophy for balancing integration tests vs unit tests, and about how to write tests that let us feel safe making changes and refactoring (without said tests being overly brittle).  that discussion will likely happen in person and on #835.

PR #818, opened 2019-06-21
[Refactor] Rename setResourceTemplate to rootResourceTemplateLoaded

jcoyne:
This makes it clearer this is only for the root template.  It also is more consistent with other actions that use "loaded" as a suffix


ndushay:
two thumbs way up!  I will merge when it is green.

ndushay:
makes me wonder if we have a component we should rename to RootResourceTemplate

PR #820, opened 2019-06-22
fix repeated placeholder text for input literals in RT where propertyURI is repeated but propertyLabel differs

### LINKED ISSUE #765: when a resource template has a property template with the same property URI as the preceding property template, the placeholder text for the second field incorrectly uses the property label for the first field
example template on stage showing this:
test:repeated:tooltips
there's one Resource Template with 2 Property Templates;
both use propertyURI http://id.loc.gov/ontologies/bibframe/geographicCoverage
(yes, there are use cases for this)
the first field has propertyLabel "Geographic Coverage 1"
the second field has propertyLabel "Geographic Coverage 2"
yet the light-colored placeholder text (the text in the entry box) says "Geographic Coverage 1" for both fields
Screen Shot 2019-06-18 at 2 15 07 PM
these fields are both literals. I also tested with the first field being a lookup and the second a literal, both with same propertyURI, and had the same problem

NOTE: an easy way to "fix" this might be to remove placeholder text altogether. There is no requirement for it that I know of.

jmartin-sul:
since fix is to pass the property template down from `PropertyComponent` to `InputLiteral`*, write a simple integration test that checks to see that a fixture RT renders as expected.

*parent already has the right instance, no need to look it up from the redux store, which just returns the property that first matches on propertyURI

closes #765

### DIFF
--- a/__tests__/fixtureLoaderHelper.js
+++ b/__tests__/fixtureLoaderHelper.js
@@ -58,6 +58,7 @@ const rtFileNames = [
   'literalRepeatDefaultLiteralOnly.json',
   'literalRepeatDefaultUriOnly.json',
   'literalRepeatNoDefault.json',
+  'propertyURIRepeated.json',
   'rdaItemMonograph.json',
 ]
 
--- a/__tests__/integration/repeatedPropertyURI.js
+++ b/__tests__/integration/repeatedPropertyURI.js
@@ -1,0 +1,17 @@
+// Copyright 2019 Stanford University see LICENSE for license
+
+import pupExpect from 'expect-puppeteer'
+import { testUserLogin } from './loginHelper'
+
+describe('A repeated propertyURI', () => {
+  beforeAll(async () => {
+    await testUserLogin()
+  })
+
+  it('renders the form with the right propertyLabels as the placeholders', async () => {
+    expect.assertions(3)
+    await pupExpect(page).toClick('a', { text: 'repeated propertyURI with differing propertyLabel' })
+    await pupExpect(page).toMatchElement('div[data-label="Geographic Coverage 1"] input[placeholder="Geographic Coverage 1"]')
+    await pupExpect(page).toMatchElement('div[data-label="Geographic Coverage 2"] input[placeholder="Geographic Coverage 2"]')
+  })
+})
--- a/src/components/editor/property/PropertyComponent.jsx
+++ b/src/components/editor/property/PropertyComponent.jsx
@@ -46,6 +46,7 @@ export class PropertyComponent extends Component {
         switch (property.type) {
           case 'literal':
             return (<InputLiteral key={keyId} id={keyId}
+                                  propertyTemplate = {property}
                                   reduxPath={reduxPath} />)
           case 'resource':
             return (<InputURI key={keyId} id={keyId}
--- a/src/components/editor/property/PropertyComponent.jsx
+++ b/src/components/editor/property/PropertyComponent.jsx
@@ -18,7 +18,7 @@ export class PropertyComponent extends Component {
     }
   }
 
-  inputComponentType = (property) => {
+  inputComponentType = (propertyTemplate) => {
     let config
 
     // We do not support mixed list and lookups, so we will just go with the value of the first config item found
@@ -30,7 +30,7 @@ export class PropertyComponent extends Component {
 
     const reduxPath = Object.assign([], this.props.reduxPath)
 
-    reduxPath.push(property.propertyURI)
+    reduxPath.push(propertyTemplate.propertyURI)
     const keyId = shortid.generate()
 
     switch (config) {
@@ -40,19 +40,19 @@ export class PropertyComponent extends Component {
       case 'list':
         return (<InputListLOC key = {this.props.index}
                               reduxPath={reduxPath}
-                              propertyTemplate = {property}
+                              propertyTemplate = {propertyTemplate}
                               lookupConfig = {this.state.configuration[0]} />)
       default:
-        switch (property.type) {
+        switch (propertyTemplate.type) {
           case 'literal':
             return (<InputLiteral key={keyId} id={keyId}
-                                  propertyTemplate = {property}
+                                  propertyTemplate = {propertyTemplate}
                                   reduxPath={reduxPath} />)
           case 'resource':
             return (<InputURI key={keyId} id={keyId}
                               reduxPath={reduxPath} />)
           default:
-            console.error(`Unknown property type (component=${config}, type=${property.type})`)
+            console.error(`Unknown propertyTemplate type (component=${config}, type=${propertyTemplate.type})`)
         }
     }
 
--- a/__tests__/__fixtures__/propertyURIRepeated.json
+++ b/__tests__/__fixtures__/propertyURIRepeated.json
@@ -1,0 +1,39 @@
+{
+  "propertyTemplates": [
+    {
+      "mandatory": "false",
+      "repeatable": "true",
+      "type": "literal",
+      "resourceTemplates": [],
+      "valueConstraint": {
+        "valueTemplateRefs": [],
+        "useValuesFrom": [],
+        "defaults": [],
+        "valueDataType": {}
+      },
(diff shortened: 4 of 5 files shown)


src/components/editor/property/PropertyComponent.jsx
```diff
@@ -46,6 +46,7 @@ export class PropertyComponent extends Component {
         switch (property.type) {
           case 'literal':
             return (<InputLiteral key={keyId} id={keyId}
+                                  propertyTemplate = {property}
```
jmartin-sul: i was tempted to rename the `property` parameter (in `inputComponentType`) to `propertyTemplate`, since that seemed like it'd be more accurate afaict.  but i wasn't sure about that.  if that does sound agreeable, i'm happy to add a touchup commit to this PR, or put up a tiny follow-on PR to that effect.

__tests__/integration/repeatedPropertyURI.js
```diff
@@ -0,0 +1,17 @@
+// Copyright 2019 Stanford University see LICENSE for license
+
+import pupExpect from 'expect-puppeteer'
+import { testUserLogin } from './loginHelper'
+
+describe('A repeated propertyURI', () => {
+  beforeAll(async () => {
+    await testUserLogin()
```
jmartin-sul: fwiw, i found that tests consistently passed for me without a `return` here.  which is what i'd sorta expected, since it doesn't seem like anything cares about the value to which the anonymous `beforeAll` promise resolves to (said promise being returned by the anonymous `async` function passed into it on line 7).  the `await` on the `testUserLogin` promise seemed like the crux of it, since that's what gates the resolving of the promise passed to the `beforeAll` (that promise resolves when the anon function finishes executing).

if people really want, i can add the `return` in, but i just wanted to note that the deviation from elsewhere is intentional.

if the above logic is sound, i'd also gently urge us to try getting rid of similar unnecessary `return`s, since they always seemed slightly misleading to me (but it's for sure not something i'd prioritize, since it's functionally harmless and it works).

ndushay: as I understand it, the "return" ensures the promise is resolved before returning, as opposed to just firing off an async thing and NOT paying attention.

https://jestjs.io/docs/en/setup-teardown

```
beforeEach and afterEach can handle asynchronous code in the same ways that tests can handle asynchronous code - they can either take a done parameter or return a promise. For example, if initializeCityDatabase() returned a promise that resolved when the database was initialized, we would want to return that promise:

beforeEach(() => {
  return initializeCityDatabase();
});
```

So I hear you about "it's an async anyway" but I don't know the jest code well enough to know if it properly handles the async, as beforeEach() is not inherently async.  And if we *need* the return, then let's use it consistently.  I'd wonder if multiple await statements in a beforeEach would *all* be executed without the return.  I'd hope so, in which case I'd happily get rid of the "return" in the beforeXXX.

__tests__/fixtureLoaderHelper.js
```diff
@@ -44,6 +44,7 @@ const rtFileNames = [
   'adminMetadata.json',
   'adminMetadataStatus.json',
   'rdaItemMonograph.json',
+  'propertyURIRepeated.json',
```
ndushay: alpha order, please

src/components/editor/property/PropertyComponent.jsx
jermnelson: 👍 I think `propertyTemplate` reduces the ambiguity that the generic `property` parameter can introduce..

jermnelson:
LGTM; I like change suggested in the parameter name.

__tests__/integration/repeatedPropertyURI.js
jmartin-sul: > as I understand it, the "return" ensures the promise is resolved before returning, as opposed to just firing off an async thing and NOT paying attention.

my understanding of this differs.  afaict, the only thing the `return` does is specify the value to which the implicit wrapping `Promise` resolves if the `async` function completes successfully.  the `return` doesn't gate any action, or force any waiting for anything, unless that waiting is conditional on the value to which the `Promise` resolves.  but i _think_ all we care about is that the `Promise` given to `beforeEach`, `beforeAll`, `it`, etc resolves if and only if the setup or test code has completed.  by definition, the `async` function `Promise` will only resolve when the body of the function completes:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function#Return_value
> ### Return Value
> A `Promise` which will be resolved with the value returned by the async function, or rejected with an uncaught exception thrown from within the async function.

> don't know the jest code well enough to know if it properly handles the async

i think from the perspective of the Jest functions, it's just getting a `Promise`.  it doesn't care whether that's because it was given an async function that always implicitly returns a `Promise`, or if it's because a "regular" function explicitly returned a `Promise` object (of its creation, from a method call, etc).

>  And if we need the return, then let's use it consistently

i would agree, but i guess i have the opposite conclusion, that it's not necessary, and that we should not use it "just in case", because that's subtly misleading.  i would be in favor of removing it where it's not needed, but i also wouldn't consider that a high priority, because other than being superfluous, it's harmless.  i don't think anything cares one way or the other about the values to which these promises resolve.

>  I'd wonder if multiple await statements in a beforeEach would all be executed without the return.

i believe they will, and i wrote a bit of test code last night to illustrate this:  https://gist.github.com/jmartin-sul/6f23e9d61b6a6f986efa99306b73ef14

happy to pair on running it (if only to be shown that i'm wrong if i am).  but ultimately, i _think_ the `waitWaitWait` part of that gist shows that there is no need for a `return` value to have `await` block in an `async` function, and i also think it shows that having an explicit `return` does not block on the completion of the `async` function unless that function call is explicitly `await`ed (e.g. `returnsAwaitResult()` and `noReturnValue()` both immediately return a `Promise` upon execution, even though neither function body will finish for a number of seconds after the function execution starts.  you can see from playing with that code that the returned `Promise` will be resolved as soon as the function body completes, but not before.

see also:  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await

> The `await` expression causes `async` function execution to pause until a `Promise` is settled, that is fulfilled or rejected, and to resume execution of the `async` function after fulfillment. When resumed, the value of the `await` expression is that of the fulfilled `Promise`.

which is to say that returning the value of the `await` does not return a `Promise` (because the result of an `await` is not a `Promise`, but the value to which the `await`ed `Promise` resolves).  an `async` method can't ever return anything other than a `Promise`

you can see some illustration of the return types of `await` statements and `async` functions in this other way-too-verbose comment:  https://github.com/LD4P/sinopia_profile_editor/pull/213#issuecomment-493646501

but i think the gist in this comment illustrates the `await` blocking better and more concisely.

as a side-note, i don't think it's particularly useful to think of a function as `async` if it is not declared as such.  `async` is just syntactic sugar for wrapping a function in a `Promise` and letting you wait for other `Promise`s that are kicked off in said function body, without so much callback boilerplate.  a function is `async` if and only if it's declared as such, regardless of whether it deals with `Promise`s internally or does asynchronous stuff inside of itself.

jmartin-sul:
@ndushay and @jermnelson, i've implemented your suggested feedback.

@ndushay, i think i may have responded to your `async` comment, but happy to discuss more.

ready for another review!

__tests__/integration/repeatedPropertyURI.js
ndushay: Thank you for the detailed explanation; I believe I am finally starting to understand, rather than trying to parrot the exact language in jest docs without fully understanding.   I *would* like to pair or o.w. go through the examples empirically w you as I think it will help solidify my understanding of how this promise stuff and async stuff works.

Feel free to make a ticket about converting existing async tests to do the right thing in their setup/teardown.

jmartin-sul: will do, thanks for reading and sanity checking!

jmartin-sul: filed #840

jmartin-sul:
merging now that it's been approved and rebased and is green again.

thanks for the feedback, all!

PR #822, opened 2019-06-22
swap the group col for author and remark ("guiding statement") cols in the template list page (/templates)

### LINKED ISSUE #636: add RT author column to list of available templates
in list of available resource templates in Linked Data Editor, add a column for the "author" field of the Resource Template, so users can see who created the Resource Template

### LINKED ISSUE #802: expose Resource Template remark in list of available resource templates
in the Linked Data Editor, Resource Templates tab, show the user the remark field (in Profile Editor this is what the profile creator puts into the "Guiding Statement for the Use of this Resource" field)

this will help users decide which RT to select

could show this as another column in the table, or a mouseover? @astridu please advise

jmartin-sul:
~~starting this out in draft because i'd be interested in getting advice or pairing on writing a simple test for this... maybe an integration test?  maybe that's overkill?  but i'll note that the code change didn't fail any existing tests (and i only made the change i did to the component's unit tests so that the example there wouldn't be confusingly out of sync with the app code, even though lack of change would've left things passing).~~

Screen Shot 2019-06-21 at 8 01 38 PM

closes #802 (remark)
closes #636 (author)

### DIFF
--- a/__tests__/components/templates/SinopiaResourceTemplates.test.js
+++ b/__tests__/components/templates/SinopiaResourceTemplates.test.js
@@ -185,6 +185,22 @@ describe('<SinopiaResourceTemplates />', () => {
     })
   })
 
+  describe('display', () => {
+    const renderRoutes = () => mount(<SinopiaResourceTemplates messages={[]}/>)
+
+    it('renders the table of resource templates with name, id, author, and guiding statement columns', () => {
+      expect.assertions(1)
+
+      const component = renderRoutes()
+      const tableHeaderCellText = component.find('table#resource-template-list th').map(thWrapper => thWrapper.text())
+      expect(tableHeaderCellText).toEqual(['Template name', 'ID', 'Author', 'Guiding statement'])
+    })
+
+    afterAll(() => {
+      renderRoutes.unmount()
+    })
+  })
+
   describe('setStateFromServerResponse()', () => {
     it('calls getResourceTemplate() once when passed a string', async () => {
       expect.assertions(2)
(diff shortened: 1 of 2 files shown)


jmartin-sul:
per some advice from @jcoyne, i started by adding a simple test to make sure that the table displays with the expected columns, since there are integration tests which cover whether the table rows display.

that said, i wanted a unit test in this component that would demonstrate the table rows displaying as expected, with the content provided by trellis, and an error message in the case that there's an issue connecting.  but i had some trouble mocking the `sinopiaServer` methods so that they'd return our fixtures (i couldn't seem to actually override the implementations of `getResourceTemplate` or `listResourcesInGroupContainer`).  i paired for a bit on that problem with @mjgiarlo, but he was also unable to spot any obvious errors in the approach i was using.  so we agreed that the thing to do is to table it for now, take this PR out of draft, and file an issue for the possible follow on work of better testing.

__tests__/components/templates/SinopiaResourceTemplates.test.js
```diff
@@ -6,7 +6,7 @@ import { mount, shallow } from 'enzyme'
 import SinopiaResourceTemplates from 'components/templates/SinopiaResourceTemplates'
 import { getEntityTagFromGroupContainer, getResourceTemplate, listResourcesInGroupContainer } from 'sinopiaServer'
 
-jest.mock('../../../src/sinopiaServer')
+jest.mock('sinopiaServer')
```
jcoyne: I like to do `import * as server from 'sinopiaServer'`.  Then you can do `server.getResourceTemplate = jest.fn()`

ndushay: i like justn's suggestion - it reads better.

jmartin-sul: noted this in a comment on #838.  will merge this since @jcoyne approved, and will try that suggestion in follow on work.

PR #823, opened 2019-06-24
note version of node needed and reference package manager that c…

peetucket:
we think there was some discussion of node v12 not working?  if so, we should reference this in the readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,9 @@ Technical documentation specific to the Sinopia Linked Data Editor may also be f
 4.  Get latest npm: `npm install -g npm@latest`
 5.  Run `npm install`. This installs everything needed for the build to run successfully.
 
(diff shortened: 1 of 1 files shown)


ndushay:
i don't know if node 12 works or not, but we're configured in circleci and images to use node 10.11, i think, so I'm good with adding this to docs.  Can look at `circleci/config.yml` for more specific info.

Also ok with upgrading to newer node;  maybe make a ticket for that, too?

peetucket:
created https://github.com/LD4P/sinopia_editor/issues/826

jcoyne:
There's a typo in your commit message "pacakage"

peetucket:
> There's a typo in your commit message "pacakage"

fixed and force pushed

PR #828, opened 2019-06-24
Don't mutate the state of objects in the store

jcoyne:
Mutation isn't picked up as a change by redux, so the component is not re-rendered.  This allows us to stop using the state change hack


ndushay:
approve, but build is failing on eslint.

jmartin-sul:
the build issue now appears to be with actual test failures, though it isn't clear to me at a glance whether it's just an issue of brittle/overly-implementation-dependent tests, or if there's actual functional breakage.

ndushay:
this PR breaks stuff.  Running from this branch, I cannot step through the nestedResource integration test steps to get to here (pic is from master branch):

image

PR #829, opened 2019-06-25
Make PropertyResourceTemplate a connected component

jcoyne:


__tests__/__fixtures__/MonographInstance.json
```diff
@@ -133,12 +133,7 @@
       "resourceTemplates": [],
       "type": "resource",
       "valueConstraint": {
-        "defaults": [
-          {
-            "defaultLiteral": "DEFAULT",
-            "defaultURI": "http://default"
-          }
-        ],
+        "defaults": [ ],
```
jcoyne: It doesn't make sense to have defaults and valueTemplateRefs, so I chose to eliminate the default.

ndushay: I'm not seeing `valueTemplateRefs` here, so I'm confused by your comment.  And it wouldn't surprise me if some tests depend on the defaults here.

src/components/editor/property/PropertyResourceTemplate.jsx
```diff
@@ -38,69 +39,72 @@ import PropertyTemplateOutline from './PropertyTemplateOutline'
...
+        <section className="col-md-12">
+          <h4>{this.props.resourceTemplate.resourceLabel}</h4>
+        </section>
       </div>
-    </div>
     )
   }
+
+  render() {
+    // repeatable defaults to false, so isAddDisabled defaults to true
```
ndushay: this comment should be moved before line 59

PR #830, opened 2019-06-25
Refactor so that PropertyResourceTemplate makes the child rows

jcoyne:
Since PropertyResourceTemplate has the add button it makes sense for it to be
responsible for generating reduxPath for the children.  This will enable the editor
to load up existing resources from Trellis.

### DIFF
--- a/__tests__/components/editor/property/PropertyResourceTemplate.test.js
+++ b/__tests__/components/editor/property/PropertyResourceTemplate.test.js
@@ -6,10 +6,14 @@ import { mount, shallow } from 'enzyme'
 import { PropertyActionButtons } from 'components/editor/property/PropertyActionButtons'
 import PropertyResourceTemplate from 'components/editor/property/PropertyResourceTemplate'
 import PropertyTemplateOutline from 'components/editor/property/PropertyTemplateOutline'
+import shortid from 'shortid'
 
 describe('<PropertyResourceTemplate />', () => {
+  shortid.generate = jest.fn().mockReturnValue('AE6Be-DJGj')
+
   const propertyRtProps = {
     resourceTemplate: {
+      id: 'resourceTemplate:bf2:WorkTitle',
       resourceLabel: 'Test Schema Thing Template',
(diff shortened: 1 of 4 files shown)


__tests__/components/editor/property/PropertyResourceTemplate.test.js
```diff
@@ -18,7 +22,7 @@ describe('<PropertyResourceTemplate />', () => {
         },
       ],
     },
-    reduxPath: ['resourceTemplate:test'],
+    reduxPath: ['http://id.loc.gov/ontologies/bibframe/title'],
```
ndushay: for those of us lagging behind on the redux store (me!) - what is this URL representing?  The propertyURI?  It might make it more obvious to use a fake one like 'http://example.org/propertyUriGoesHere'.

src/components/editor/property/PropertyResourceTemplate.jsx
```diff
@@ -6,12 +6,43 @@ import PropTypes from 'prop-types'
...
+ *         'http://id.loc.gov/ontologies/bibframe/mainTitle: {
+ *
+ *         }
+ *       }
+ *     }
+ *   }
+ *
+ *  In the props we are passed the resourceTemplate, so we can get the rows from
+ *  the redux store by filtering path for items that have a key that matches
+ *  resourceTemplate.id
```
ndushay: thanks for this documentation - super helpful.

__tests__/components/editor/property/PropertyResourceTemplate.test.js
jcoyne: I think it's important to use a real one so we can show that title can have more than one possible sub-resources. In this case VariantTitle, TransliteratedTitle or WorkTitle.

src/components/editor/property/PropertyResourceTemplate.jsx
jmartin-sul: +1

PR #833, opened 2019-06-25
Show the URI on the template page after saving a resource

### LINKED ISSUE #766: when saving RDF from preview, user should see verification that it works and should get the URI
right now, the only way to see is by clicking "preview RDF" again and noting that there is now a URI as the subject where previously it was empty.

aaron-collier:
Fixes #766 

This adds a small template that is turned on when a resource is saved successfully and the baseURL is set.

BEFORE:

Screen Shot 2019-06-25 at 10 09 40 AM

AFTER:

Screen Shot 2019-06-25 at 11 20 36 AM

A follow up PR is necessary to write a proper integration/end-to-end test that saves to Trellis.


package.json
```diff
@@ -80,8 +80,8 @@
     "dev-build": "webpack --progress --mode development",
     "dev-build-test": "npm run dev-build && npm run test",
     "dev-start": "webpack-dev-server --config ./webpack.config.js --mode development",
-    "eslint": "eslint --max-warnings 7 --color -c .eslintrc.js --ext js,jsx .",
-    "eslint-fix": "eslint --max-warnings 7 --color -c .eslintrc.js --ext js,jsx --fix .",
+    "eslint": "eslint --max-warnings 8 --color -c .eslintrc.js --ext js,jsx .",
+    "eslint-fix": "eslint --max-warnings 8 --color -c .eslintrc.js --ext js,jsx --fix .",
```
aaron-collier: Note: I had to make this change as the "complexity" configuration was apparently being ignored, and producing a warning on the `reducers/index.js` switch statement.

src/components/editor/ResourceTemplate.jsx
```diff
@@ -50,6 +51,7 @@ class ResourceTemplate extends Component {
         <div id="resourceTemplate" style={{ marginTop: '-30px' }}>
           <section className="col-md-9">
             <h1><em>{this.props.title}</em></h1>
+            <BaseURL />
```
ndushay: if there is a problem saving ... we don't want this to show, correct?  We would want the error message modal and to NOT show new component?

src/components/editor/baseURL.jsx
```diff
@@ -0,0 +1,30 @@
+// Copyright 2018 Stanford University see LICENSE for license
+
+import React from 'react'
+import { connect } from 'react-redux'
+import PropTypes from 'prop-types'
+
+// Renders the base URL after Save & Publish
+const BaseURL = (props) => {
```
ndushay: i'm not in love with this name -- to me, this is like the first part of an API url, before the API specific args are added.  Also, this component is showing a div with a message containing the URI (not URL) from trellis.

How about:

ResourceUriMessage

.. and changing the action and reducer method names as well?

aaron-collier: Yeah, I went back and forth on that, the return value from save is "saveBaseURL", so I went with this for consistency, but started out with "ResourceURI". Happy to switch back for clarity.

src/components/editor/ResourceTemplate.jsx
aaron-collier: It shouldn't show up if there is a problem I believe, but will try to verify.

src/components/editor/baseURL.jsx
ndushay: if you want to keep "baseURL" maybe "BaseUriMessage" ?  But I think all of the usage of baseURL was arbitrarily chosen and ripe for change to "resourceUri".  Just my opinion, tho.

src/components/editor/ResourceTemplate.jsx
ndushay: wonder if it would be worth writing a test (with a mock, not with network call to server)

src/components/editor/baseURL.jsx
aaron-collier: Renamed to ResourceURIMessage :-)

src/components/editor/ResourceTemplate.jsx
aaron-collier: I'll try to get a test going, confirmed that it does not show up if there is an error.

src/components/editor/resourceURIMessage.jsx
```diff
@@ -0,0 +1,30 @@
...
+
+// Renders the resource URI message after Save & Publish
+const ResourceURIMessage = (props) => {
+  if (!props.show) {
+    return null
+  }
+
+  return (
+    <div>
+      <h4>URI for this resource: &lt;{ props.uri }&gt;</h4>
```
ndushay: should there be an if in here to be sure props.uri has a value?

should there be simple unit tests that check that the uri is output when present, and that it outputs something suitable when it's not present?  (first part only, if having a check in here isn't warranted.)

lastly, should there be an "integration" test showing this message displays after Save & Publish

__tests__/reducers/inputs.test.js
```diff
@@ -344,6 +348,23 @@ describe('setBaseURL', () => {
...
+
+    const result = showResourceURIMessage(initialState, {
+      type: 'SHOW_RESOURCE_URI_MESSAGE',
+      payload: 'http://example.com/foo/123',
+    })
+
+    expect(result.editor.resourceURIMessage.show).toBe(true)
+    expect(result.editor.resourceURIMessage.uri).toEqual('http://example.com/foo/123')
+  })
+})
```
ndushay: 👍 .   question - how does editor.resourceURIMessage.show become false again?

jmartin-sul: my read of the code is that that changing back isn't expected or implemented behavior for now.  i don't see anything to transform the state back in that direction in this PR.

src/components/editor/resourceURIMessage.jsx
jmartin-sul: i think a simple `ResourceURIMessage.WrappedComponent` test for the basic cases (no show, show but no URI, show is true and there's a URI) would be easy and a nice start (shouldn't require any mocking).

agreed an integration test that shows things are wired up correctly via `mapStateToProps` would be nice too.

PR #839, opened 2019-06-26
Change empty list label to something more user-friendly

jgreben:


ndushay:
I agree with Mike - I assume such a test would be time consuming to write;  ow LGTM

jgreben:
A test for this would essentially be testing the 3rd-party component functionality: https://github.com/ericgio/react-bootstrap-typeahead/blob/6b45fbb7fc09a5694851fab616a21a023af8f399/test/utils/getAccessibilityStatusSpec.js#L16

PR #842, opened 2019-06-26
Amcollie 766 jmartin followon

jmartin-sul:
a couple small touchups that were as easy to make as they were to comment on, so PR to the PR, for your convenience if you like the suggestions

spawned by #833

### DIFF
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -44,9 +44,9 @@ export const assignBaseURL = item => ({
   payload: item,
 })
 
-export const showResourceURIMessage = show => ({
+export const showResourceURIMessage = resourceUri => ({
   type: 'SHOW_RESOURCE_URI_MESSAGE',
-  payload: show,
+  payload: resourceUri,
 })
 
 export const updateStarted = () => ({
--- a/src/components/editor/ResourceTemplate.jsx
+++ b/src/components/editor/ResourceTemplate.jsx
@@ -7,7 +7,7 @@ import ResourceTemplateForm from './ResourceTemplateForm'
 import { rootResourceTemplateLoaded } from 'actions/index'
 import { getResourceTemplate } from 'sinopiaServer'
 import { rootResource } from 'selectors/resourceSelectors'
-import ResourceURIMessage from './resourceURIMessage'
+import ResourceURIMessage from './ResourceURIMessage'
 
 const _ = require('lodash')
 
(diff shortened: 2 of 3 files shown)


src/components/editor/ResourceURIMessage.jsx
```diff
@@ -1,24 +1,20 @@
...
+// Copyright 2019 Stanford University see LICENSE for license
 
 import React from 'react'
 import { connect } from 'react-redux'
 import PropTypes from 'prop-types'
 
 // Renders the resource URI message after Save & Publish
 const ResourceURIMessage = (props) => {
-  if (!props.show) {
+  if (!props.show || !props.uri) {
```
jmartin-sul: ~~stockholm~~ rubocop syndrome has me looking for places to use guard clauses

src/components/editor/ResourceTemplate.jsx
```diff
@@ -7,7 +7,7 @@ import ResourceTemplateForm from './ResourceTemplateForm'
 import { rootResourceTemplateLoaded } from 'actions/index'
 import { getResourceTemplate } from 'sinopiaServer'
 import { rootResource } from 'selectors/resourceSelectors'
-import ResourceURIMessage from './resourceURIMessage'
+import ResourceURIMessage from './ResourceURIMessage'
```
ndushay: wow - this worked with the wrong case?  my world is broken.

src/actions/index.js
```diff
@@ -44,9 +44,9 @@ export const assignBaseURL = item => ({
   payload: item,
 })
 
-export const showResourceURIMessage = show => ({
+export const showResourceURIMessage = resourceUri => ({
```
ndushay: I like this change.

src/components/editor/ResourceURIMessage.jsx
ndushay: I like this change.  And the humor.

src/components/editor/ResourceTemplate.jsx
jmartin-sul: ah, no, i renamed the file too (that was my first change).

actually, i think it did work without the case change, but i did it the other way (renamed the file, ran the tests successfully iirc, got called out by eslint on the import when circle ran the build).

src/actions/index.js
jmartin-sul: thx!

src/components/editor/ResourceURIMessage.jsx
aaron-collier: I like this change. I started here, as my mind is likewise in ruby guard clauses, but didn't see it often in examples...

PR #843, opened 2019-06-27
tests: if setup or tear down is async, no need for explicit return

### LINKED ISSUE #840: jest test refinement:  probably don't need to return await result in async setup functions
per [this comment](https://github.com/LD4P/sinopia_editor/pull/820#discussion_r297858708), it seems unnecessary to explicitly `return` the last `await`ed value in an `async` setup function (passed e.g. to `beforeEach` or `beforeAll`).  though if my understanding is correct, it is also harmless (since it seems unlikely that anything pays attention to the values to which those `Promise`s resolve, just that they resolve at all).

anyway, since the `return` is unnecessary, it seems slightly misleading, so i'm filing this backlog ticket to clean things up if and when we have time.  @ndushay thinks there are not that many instances of this sort of thing anyway, so probably not a huge undertaking.

**NOTE**:  this is _not_ suggesting that all `return await` constructions should be removed.  there are absolutely places where we want to return an `await`ed value so that the `Promise` resolves to that value (e.g. in `src/sinopiaServer.js`).  this ticket is only about unnecessary returns in test setup.

ndushay:
fixes #840

This PR adapts our tests to have better syntax for async setup and teardown methods. 

from https://jestjs.io/docs/en/setup-teardown

> beforeEach and afterEach can handle asynchronous code in the same ways that tests can  handle asynchronous code - they can either take a done parameter or return a promise. For example, if [ promiseFromFunction() ] returned a promise that resolved when the database was initialized, we would want to return that promise:

```js
beforeEach(() => {
  return promiseFromFunction()
})
```

As @jmartin-sul points out in #840 and also in [this PR comment](https://github.com/LD4P/sinopia_editor/pull/820#discussion_r297858708), an `async` method, by definition, returns a promise.  So this is also an appropriate way for a setup/teardown method to handle asynchronous code:

```js
beforeEach(async () => {
  await promiseFromFunction()
})
```

The benefit of the second approach is it also allows for multiple await statements, and makes the asynchronous code more obviously async:

```js
beforeEach(async () => {
  await promiseToReviewThisPR()
  await promiseToMergeThisPR()
  await promiseToFinishSinopiaWorkCycle()
})
```

### DIFF
--- a/__tests__/integration/correctRdfInPreview/literalPropertyTemplates.test.js
+++ b/__tests__/integration/correctRdfInPreview/literalPropertyTemplates.test.js
@@ -9,12 +9,12 @@ import { testUserLogin } from '../loginHelper'
  */
 describe('RDF from literal property templates', () => {
   beforeAll(async () => {
-    return await testUserLogin()
+    await testUserLogin()
   })
 
   beforeEach(async () => {
     await page.goto('http://127.0.0.1:8888/templates')
-    return await page.waitForSelector('.react-bootstrap-table')
+    await page.waitForSelector('.react-bootstrap-table')
   })
 
   it('non-repeatable, no default value, default language', async () => {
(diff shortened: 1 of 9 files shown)


ndushay:
and don't worry - it took me longer to write up the above than to make the changes and create the PR.

PR #848, opened 2019-06-27
Get resource template title for display in editor

### LINKED ISSUE #844: Title of resource template is not being displayed in editor
Screen Shot 2019-06-27 at 8 18 45 AM

aaron-collier:
Fixes #844 

This sets the title of the resource Template into `resourceTemplateLabel` in order to be displayed on the editor page.

Note: This bug was missed by the simple unit test due to the injection of the title guaranteeing it's availability in the display. Basic integration tests like the one below are critical in surfacing these kinds of bugs by using fixtures that are compared to the output.

### DIFF
--- a/__tests__/components/editor/ResourceTemplate.test.js
+++ b/__tests__/components/editor/ResourceTemplate.test.js
@@ -13,7 +13,7 @@ describe('<ResourceTemplate />', () => {
   const wrapper = shallow(<ResourceTemplate.WrappedComponent resourceTemplateId="resourceTemplate:bf2:Note"
                                                              resourceTemplate={reduxState}
                                                              retrieveResourceTemplate={jest.fn()}
-                                                             rtLabel="BF2 Work"/>)
+                                                             resourceTemplateLabel="BF2 Work"/>)
 
   it('has div with class "ResourceTemplate"', () => {
     expect(wrapper.find('div.ResourceTemplate').length).toEqual(1)
--- a/__tests__/integration/verifyTemplateTitle.test.js
+++ b/__tests__/integration/verifyTemplateTitle.test.js
@@ -2,7 +2,6 @@
 
 import pupExpect from 'expect-puppeteer'
 import { testUserLogin } from './loginHelper'
-import { fillInRequredFieldsForBibframeInstance } from './previewRDFHelper'
 
 describe('Verify the resource template title is displayed in the editor', () => {
   beforeAll(async () => {
@@ -14,6 +13,5 @@ describe('Verify the resource template title is displayed in the editor', () => 
 
     await pupExpect(page).toClick('a', { text: 'Note' })
     await pupExpect(page).toMatchElement('h1', { text: 'Note' })
-
   })
 })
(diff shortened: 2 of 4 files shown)


src/components/editor/ResourceTemplate.jsx
```diff
@@ -51,6 +52,7 @@ ResourceTemplate.propTypes = {
 
 const mapStateToProps = state => ({
   resourceTemplate: rootResource(state),
+  title: state.selectorReducer.editor.rootResourceTitle,
```
ndushay: so it didn't work as

`state.selectorReducer.entities.resourceTemplates[rtId].label ?`

cause that would avoid the addition of something to editor in redux state

aaron-collier: Let me try that. I think I missed `rtId` when I tried this the first time.

aaron-collier: Nope, getting related syntax errors `rtId is unknown` when I try that. Happy to pair on this if preferred, but I'm clear on how to make that work in this case.

aaron-collier: DONE :-)

__tests__/integration/verifyTemplateTitle.test.js
```diff
@@ -0,0 +1,17 @@
+// Copyright 2019 Stanford University see LICENSE for license
+
+import pupExpect from 'expect-puppeteer'
+import { testUserLogin } from './loginHelper'
+
```
ndushay: For the record, Aaron write this, I did not.  Note that this is called an "integration" test, but it uses fixture data and is focused on a simple thing.

What was broken had to do with a mismatch between was going in redux vs. what was expected from redux.  I would be keen to learn techniques to make our unit tests less vulnerable to this sort of problem.

__tests__/components/editor/ResourceTemplate.test.js
```diff
@@ -13,14 +13,14 @@ describe('<ResourceTemplate />', () => {
...
-                                                             title="BF2 Work"/>)
+                                                             resourceTemplateLabel="BF2 Work"/>)
 
   it('has div with class "ResourceTemplate"', () => {
     expect(wrapper.find('div.ResourceTemplate').length).toEqual(1)
   })
 
   it('displays the resource label of the resource template', () => {
-    expect(wrapper.find('h1').text()).toEqual('BF2 Work')
+    expect(wrapper.find('#resourceTemplate h1').text()).toEqual('BF2 Work')
```
ndushay: this selector was made more specific and improves comprehension when reading the code.

src/components/editor/ResourceTemplate.jsx
```diff
@@ -27,11 +27,12 @@ class ResourceTemplate extends Component {
     if (!this.props.resourceTemplate) {
       return errorMessage
     }
+
     return (
       <div className="ResourceTemplate">
         <div id="resourceTemplate" style={{ marginTop: '-30px' }}>
           <section className="col-md-9">
-            <h1><em>{this.props.title}</em></h1>
+            <h1><em>{this.props.resourceTemplateLabel}</em></h1>
```
ndushay: we changed this from "title" in this context, because inside the resourceTemplate, it is "resourceLabel".  It is a "title" of sorts from the perspective of the html we generate.

jgreben: Also, the props are now mapped from the redux state, and not passed down as `title` from the parent component.

jermnelson: 👍

Issue #694, opened 2019-06-12
InputListLOC rendering driven by Resource and Entity State


jermnelson:
The InputListLOC component currently uses a combination of PropertyTemplate passed in as prop and it's React state for handling properties like repeatable and mandatory and for setting defaults. All props and defaults should be set in the Redux reducer call in the `mapStateToProps` based on the InputListLOC's reduxPath with the reduxPath using the stored Resource Templates in the entities object in the Redux state.

Blocks #673

justinlittman:
completing this ticket depends on #698.

jcoyne:
See https://github.com/LD4P/sinopia_editor/issues/772#issuecomment-503630523

ndushay:
I mistyped the issue number in PR #856;  this should have been closed when #856 was merged.

PR #852, opened 2019-06-28
npm update eslint-plugin-react

### LINKED ISSUE #468: update eslint-plugin-react npm package
updating from 7.11.1 to 7.12.4 causes eslint errors

jmartin-sul:
this puts us on the latest version of `eslint-plugin-react`.

upgrade eslint-plugin-react, clean up newly found lint.
* add propTypes declarations
  * fix a test that wasn't actually testing what the code was trying to do, and which led to an error under stricter type checking

closes #468

**note**:  i always get changes of the sort seen in the second commit when i `npm install`.  is there any harm in resolving those packages to specific `https://registry.npmjs.org` URLs?  the specificity maybe even seems desirable to me from a stability/reproducibility/security standpoint?  but maybe there are downsides i'm unaware of?

i can't really find any advice on the subject.  the most relevant thing i came across was:
https://stackoverflow.com/a/53827243
https://github.com/npm/npm/issues/16849#issuecomment-312442508
@zkat said:
> The purpose of `resolved` in `package-lock.json` is to bypass the dependency resolution step (fetching metadata) when you are missing packages. `integrity` is to verify that you're getting the same thing. Without the `resolved` field, uncached installations can break due to metadata changes, and they'll also be significantly slower because we have to do a full metadata fetch before we can actually download anything.
> 
> `package-lock.json` is intended to make sure that you get the exact same thing, as much as possible, down to the byte.
> 
> Note that `package-lock.json` does not allow different packages to be fetched from different registries. Even if you have a package lock with different packages using different `resolved` fields, all of the packages will _always_ be fetched from whatever your current `registry=` setting is, in `npmrc`. `resolved` fields that do not match the configured registry will go through the (slower) metadata fetching I mentioned above, but will still be fetched only from the current registry.

which implies to me that it *may* be an efficiency improvement in some situations, and otherwise, it doesn't do anything?

the npmjs.org doc is light on context/advice:  https://docs.npmjs.com/files/package-lock.json
>  * For bundled dependencies this is not included, regardless of source.
> * For registry sources this is path of the tarball relative to the registry URL. If the tarball URL isn’t on the same server as the registry URL then this is a complete URL.

anyway, happy to lop off that commit if people want.

### DIFF
--- a/src/components/editor/property/InputListLOC.jsx
+++ b/src/components/editor/property/InputListLOC.jsx
@@ -132,6 +132,10 @@ class InputListLOC extends Component {
 }
 
 InputListLOC.propTypes = {
+  defaults: PropTypes.arrayOf(PropTypes.object),
+  displayValidations: PropTypes.bool,
+  handleSelectedChange: PropTypes.func,
+  lookupConfig: PropTypes.arrayOf(PropTypes.object),
   propertyTemplate: PropTypes.shape({
     propertyLabel: PropTypes.string,
     propertyURI: PropTypes.string,
@@ -141,7 +145,7 @@ InputListLOC.propTypes = {
       useValuesFrom: PropTypes.oneOfType([PropTypes.string, PropTypes.array]),
     }),
   }),
-  displayValidations: PropTypes.bool,
+  reduxPath: PropTypes.array,
 }
 
 const mapStateToProps = (state, ownProps) => {
--- a/package-lock.json
+++ b/package-lock.json
@@ -5314,16 +5314,49 @@
       }
     },
     "eslint-plugin-react": {
-      "version": "7.11.1",
-      "resolved": "https://registry.npmjs.org/eslint-plugin-react/-/eslint-plugin-react-7.11.1.tgz",
-      "integrity": "sha512-cVVyMadRyW7qsIUh3FHp3u6QHNhOgVrLQYdQEB1bPWBsgbNCHdFAeNMquBMCcZJu59eNthX053L70l7gRt4SCw==",
+      "version": "7.14.2",
+      "resolved": "https://registry.npmjs.org/eslint-plugin-react/-/eslint-plugin-react-7.14.2.tgz",
+      "integrity": "sha512-jZdnKe3ip7FQOdjxks9XPN0pjUKZYq48OggNMd16Sk+8VXx6JOvXmlElxROCgp7tiUsTsze3jd78s/9AFJP2mA==",
       "dev": true,
       "requires": {
         "array-includes": "^3.0.3",
         "doctrine": "^2.1.0",
         "has": "^1.0.3",
-        "jsx-ast-utils": "^2.0.1",
-        "prop-types": "^15.6.2"
+        "jsx-ast-utils": "^2.1.0",
+        "object.entries": "^1.1.0",
+        "object.fromentries": "^2.0.0",
+        "object.values": "^1.1.0",
+        "prop-types": "^15.7.2",
+        "resolve": "^1.10.1"
+      },
+      "dependencies": {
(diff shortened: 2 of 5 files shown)


src/components/editor/property/InputListLOC.jsx
```diff
@@ -138,7 +141,7 @@ InputListLOC.propTypes = {
       useValuesFrom: PropTypes.oneOfType([PropTypes.string, PropTypes.array]),
     }),
   }).isRequired,
-  displayValidations: PropTypes.bool,
+  reduxPath: PropTypes.array,
```
jmartin-sul: at first i made `reduxPath` a `PropTypes.array.required`, but tests failed because `InputListLOC.test.js` wasn't passing it in.  but it seems like IRL, the parent component always passes it in, and it's required in the parent component, so maybe `.required` would be more appropriate, and the tests should then be changed to reflect that?

if that's the case, i'm inclined  to ticket that as follow on work.  this still seems like an improvement, since there's now at least type checking where there previously wasn't any.

__tests__/components/editor/property/InputListLOC.test.js
```diff
@@ -65,8 +65,8 @@ describe('<InputListLOC /> configuration', () => {
     expect(global.alert.mock.calls.length).toEqual(1)
   })
 
-  it('displays a browser alert if the lookupConfig is an array of objects and not a single object', () => {
-    shallow(<InputListLOC.WrappedComponent {...plProps} handleSelectedChange={mockFormDataFn} lookupConfig={mockLookupConfig} />)
+  it('displays a browser alert if the lookupConfig is missing the uri field', () => {
+    shallow(<InputListLOC.WrappedComponent {...plProps} handleSelectedChange={mockFormDataFn} lookupConfig={{ label: 'no URI' }} />)
```
ndushay: nice catch.  👍

src/components/editor/property/InputListLOC.jsx
ndushay: While working on #694, I bumped into similar stuff.  Did you mean `.isRequired` as is on the propertyTemplate on line 143?  

Also https://github.com/LD4P/sinopia_editor/commit/5a0090af5656b986fb6c29375f23dfbf067ce0f1#diff-93b04186261f1a6fa9e8f243d588ef94 removed `.isRequired` from propertyTemplate and I found I had to do that while working on #694

mjgiarlo:
@jmartin-sul This looks ready to go once conflicts are resolved. Moved from in progress to needs review.

src/components/editor/property/InputListLOC.jsx
jmartin-sul: tried `isRequired` too, but same result.  might investigate as follow-on work.

jmartin-sul: filed #901

jmartin-sul:
@ndushay said:
> Two follow on ideas, neither urgent.
>
> 1. what you mention below about required stuff
> 1. use "ncu" to see if there is a major version bump available. https://www.npmjs.com/package/npm-check-updates.
>

filed an issue for the first thing (#901).

to the second point, 7 is still the latest version of eslint-plugin-react, so this PR gets us to the latest by way of regular `npm update`.  captured the larger issue in this infra maintenance card:  https://github.com/orgs/sul-dlss/projects/1#card-21576766

> I update packages for any of these reasons:
>
> 1.  we are getting security alerts on npm install (or circleCI is, or github is showing them)
> 1. we need features from newer version
> 1. trying to update all packages at once

same!  no security alerts on `npm install` at the moment (or any others i'm aware of).  second and third points seem out of scope for this issue.

PR #856, opened 2019-06-28
InputListLOC uses redux state for propertyTemplate and lookupConfig 

### LINKED ISSUE #604: Add resourceURI for Resource Templates in  Redux State
In the current Redux state for the application, the Resource Template ID is used as a key for all of the properties associated with it. We want to be able to associated the `resourceURI` for that Resource template assist in the generation of the `rdf:type` for that entity.

For example:

Current Redux shape for a resource template:
```javascript
 selectorReducer: {
    'resourceTemplate:bf2:Monograph:Instance': {
      'http://id.loc.gov/ontologies/bibframe/instanceOf': {}
     ...
  }
```

With the resourceURI stored as `rdfClass` (could be another name):

```javascript
 selectorReducer: {
    'resourceTemplate:bf2:Monograph:Instance': {
     'rdfClass': '"http://id.loc.gov/ontologies/bibframe/Instance",
      'http://id.loc.gov/ontologies/bibframe/instanceOf': {},
    ...
}
```
 We will likely need this to be place when we mint the initial URI for the entity.

ndushay:
Closes #604.  Note that #604 was at the top of the ready column when I started the work.

In addition to getting propertyTemplate and lookupConfig from redux, we also retrieve the propertyTemplate defaults from the selectorReducer.

Lastly, saw an opportunity to tweak some comments in 2 other spots, so I took it.  Happy to break those into separate PRs if they are controversial.

### DIFF
--- a/__tests__/components/editor/property/InputListLOC.test.js
+++ b/__tests__/components/editor/property/InputListLOC.test.js
@@ -5,7 +5,7 @@ import { shallow } from 'enzyme'
 import { Typeahead } from 'react-bootstrap-typeahead'
 import InputListLOC from 'components/editor/property/InputListLOC'
 
-const plProps = {
+const propsOk = {
   propertyTemplate:
     {
       propertyURI: 'http://id.loc.gov/ontologies/bflc/target',
@@ -19,7 +19,6 @@ const plProps = {
           defaultURI: 'http://id.loc.gov/vocabulary/carriers/nc',
           defaultLiteral: 'volume',
         }],
-        valueTemplateRefs: [],
         useValuesFrom: [
           'vocabulary:bf2:frequencies',
         ],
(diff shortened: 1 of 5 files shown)


__tests__/components/editor/property/InputListLOC.test.js
```diff
@@ -108,52 +174,6 @@ describe('<Typeahead /> component', () => {
     expect(wrapper.find('#targetComponent').props().placeholder).toMatch('Frequency (RDA 2.14)')
   })
 
-  describe('default values', () => {
```
ndushay: since we are using selectorReducer function to get the defaults from the propertyTemplate, there is no need to test this here.

src/components/editor/property/InputListLOC.jsx
```diff
@@ -6,27 +6,25 @@ import PropTypes from 'prop-types'
...
-      error: null,
     }
   }
 
   componentDidMount() {
     this._isMounted = true
-    if (this.state.defaults.length > 0) {
-      this.selectionChanged(this.state.defaults)
-    } else {
-      // Property templates do not require defaults but we like to know when this happens
```
ndushay: We decided to ditch this kind of message in another spot, so I ditched it here also.  I suspect this is from earlier efforts when we were less confident of our handling of default values.

src/components/editor/property/InputListLOC.jsx
```diff
@@ -62,11 +60,20 @@ class InputListLOC extends Component {
   }
 
   render() {
-    if (this.props.lookupConfig?.uri === undefined) {
+    // Don't render if no property template yet
+    if (!this.props.propertyTemplate) {
+      return null
+    }
```
ndushay: stole this guard clause from InputLookupQA - it seemed apt.

src/components/editor/property/InputListLOC.jsx
```diff
@@ -62,11 +60,20 @@ class InputListLOC extends Component {
...
+
+    if (this.props.lookupConfig?.length > 1) {
+      alert(`There are multiple configured list lookups for ${this.props.propertyTemplate.propertyURI}`)
+    }
```
ndushay: there was a test ostensibly for this, so I put in the guard clause

src/components/editor/property/InputListLOC.jsx
```diff
@@ -97,7 +104,7 @@ class InputListLOC extends Component {
           ref={typeahead => this.typeahead = typeahead}
           onFocus={() => {
             this.setState({ isLoading: true })
-            fetch(`${this.props.lookupConfig.uri}.json`)
+            fetch(`${this.props.lookupConfig[0].uri}.json`)
```
ndushay: previously, what was passed was the first one, so this represents moving the hardcoding from PropertyComponent to this component

src/components/editor/property/InputListLOC.jsx
ndushay: i think you are right that some of these errors were for vestigial code.  In this PR, I decided to explicitly check when there was an undefined uri, and then created the explicit guard clause to match the test that was there for when the lookupConfig json has a property uri in there twice.  

You'll also notice that I had to redo the inline fixture props for the tests, and I decided to name them something helpful to help the readability of the tests.

I didn't delve far into "what, precisely is in the lookupConfig and why should this component get lookupConfig[0] when the other didn't ..." -- so yeah, there might be room for cleaner code there.  Hardcoding to "first one" hidden in the code is not my fav ... but I wanted to get the PR done.

src/components/editor/property/InputListLOC.jsx
ndushay: see comment above - this feels hinky but I let it slide.

ndushay:
thanks for the merge.  I think this PR is more explicit which errors are tested by which data and which error messages come from which errors.  So it gets my vote for that, not because I wrote it.

PR #862, opened 2019-07-01
Adds ability to download resource templates.

### LINKED ISSUE #637: allow user to download RT JSON from the available templates
in Available Resource Templates in Sinopia, allow user to download the Resource Template json.  For users who are very involved in creating profiles, this will be very helpful to learn how different Resource Templates are set up and to compare Resource Templates

(This feature requested by developers too)

justinlittman:
closes #637

### DIFF
--- a/__tests__/components/templates/Download.test.js
+++ b/__tests__/components/templates/Download.test.js
@@ -1,0 +1,25 @@
+// Copyright 2018 Stanford University see LICENSE for license
+
+import React from 'react'
+import { shallow } from 'enzyme'
+import Download from 'components/templates/Download'
+
+describe('<Download />', () => {
+  it('has a download link', () => {
+    const wrapper = shallow(<Download filename="example.json" blob={ {} }/>)
+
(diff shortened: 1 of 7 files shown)


justinlittman:
Thanks for the feedback @jmartin-sul. I merged your suggestions and was able to fix the problem in FF (and while I was at it, switched from Chrome to FF, so, bonus).

The reason I chose to go with the download instead of link to trellis is so that I didn't have to add special handling for development, when trellis isn't available. A bit of a toss up on cleanest approach.

jmartin-sul:
> switched from Chrome to FF, so, bonus

nice!

> The reason I chose to go with the download instead of link to trellis is so that I didn't have to add special handling for development, when trellis isn't available. A bit of a toss up on cleanest approach.

ah, totally, hadn't thought of that.  twas nice being able to test with the fixtures, now that you mention it, since spinning up trellis with usable data for the editor is still a bit of manual work.

PR #865, opened 2019-07-02
Removes property template as InputLiteral prop. Validates resource te…

### LINKED ISSUE #765: when a resource template has a property template with the same property URI as the preceding property template, the placeholder text for the second field incorrectly uses the property label for the first field

### LINKED ISSUE #849: InputLiteral component should retrieve propertyTemplate from redux state via reduxPath
analogous to issue #694 (for InputListLOC) and to 5a0090af  

The InputLiteral component currently uses a combination of propertyTemplate passed in as prop and redux state.  propertyTemplate should be derived from the reduxPath using the stored Resource Templates in the entities object in the Redux state.

justinlittman:
…mplates for repeated property URIs.

closes #765. closes #849.

### DIFF
--- a/src/ResourceTemplateValidator.js
+++ b/src/ResourceTemplateValidator.js
@@ -1,0 +1,26 @@
+// Copyright 2019 Stanford University see LICENSE for license
+
+/**
+ * Validates a resource template.
+ * Validation checks that does not contain repeated property templates.
+ * @param {Object>} resourceTemplate to validate
+ * @return {string|null} reasons that validation failed if invalid otherwise null if valid
+ */
+const validateResourceTemplate = (resourceTemplate) => {
+  const dupes = []
+  const propertyTemplateIds = []
+  resourceTemplate.propertyTemplates.forEach((propertyTemplate) => {
+    const propertyURI = propertyTemplate.propertyURI
+    if (propertyTemplateIds.indexOf(propertyURI) !== -1) {
+      dupes.push(propertyURI)
+    } else {
+      propertyTemplateIds.push(propertyURI)
+    }
+  })
+  if (dupes.length > 0) {
+    return `Repeated property templates with same property URI (${dupes}) are not allowed.`
+  }
+  return null
+}
+
+export default validateResourceTemplate
--- a/src/actionCreators.js
+++ b/src/actionCreators.js
@@ -8,6 +8,7 @@ import {
 } from 'actions/index'
 import { updateRDFResource, getResourceTemplate } from 'sinopiaServer'
 import { rootResourceId } from 'selectors/resourceSelectors'
+import validateResourceTemplate from 'ResourceTemplateValidator'
 import GraphBuilder from 'GraphBuilder'
 
 export const authenticationFailed = authenticationResult => authenticationFailure(authenticationResult)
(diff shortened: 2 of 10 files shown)


src/actionCreators.js
```diff
@@ -31,9 +31,33 @@ export const fetchRootResourceTemplate = resourceTemplateId => (dispatch) => {
...
+    } else {
+      dispatch(rootResourceTemplateLoaded(resourceTemplate))
+    }
   }).catch((err) => {
     console.error(err)
     dispatch(retrieveError(resourceTemplateId))
   })
 }
+
+export const validateResourceTemplate = (resourceTemplate) => {
```
jcoyne: Can you move this to a different file, because it is not an actionCreator (it does not return a function that takes dispatch and optionally getState)

src/reducers/index.js
```diff
@@ -99,10 +99,18 @@ export const resourceTemplateLoaded = (state, action) => {
...
+  const resourceTemplateId = action.payload.resourceTemplateId
+  const reason = action.payload.reason
   const newState = { ...state }
 
-  newState.editor.serverError = `There was a problem retrieving ${resourceTemplateId}`
+  let serverError = `There was a problem retrieving ${resourceTemplateId}`
+  if (reason) {
+    serverError += `: ${reason}`
+  } else {
+    serverError += '.'
```
ndushay: would it make sense to use something akin to the old verbiage here?  "Sinopia server may be offline or have no resource templates to display" ?

ndushay: Also, vis-a-vis this morning's conversation ... a test for this else branch might be appropriate ...

src/actionCreators.js
justinlittman: Done.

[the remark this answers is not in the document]
justinlittman: Add comment docs for the method.

Don't which whether null or undefined or other is idiomatic js.

src/reducers/index.js
justinlittman: Added tests.

The old verbiage was sometimes wrong. The reducers would specify an error message and this would override and display a different error message.

src/ResourceTemplateValidator.js
```diff
@@ -0,0 +1,26 @@
+// Copyright 2019 Stanford University see LICENSE for license
+
+/**
+ * Validates a resource template.
+ * Validation checks that does not contain repeated property templates.
+ * @param {<object>} resource template to validate
```
jcoyne: Add the parameter name here. I believe the convention for jsdoc is to use `{Object}` (capitalized, no `<>`)

src/ResourceTemplateValidator.js
```diff
@@ -0,0 +1,26 @@
...
+ * @return {<string> or null} reasons that validation failed if invalid otherwise null if valid
```
jcoyne: Multiple types should be written as `{(string|null)}`

src/reducers/index.js
ndushay: ja, agree and appreciate new logic.  I cleverly changed the old wording to "may" - like, if you don't have a reason, this MIGHT be why.  But not as important as the test - thanks for that.

PR #868, opened 2019-07-02
Proxy ElasticSearch service

### LINKED ISSUE #630: Create Search Express route in Editor for ElasticSearch queries

mjgiarlo:
Fixes #630

Includes:
* Add middleware to Express server to proxy ElasticSearch requests
  * Allow only GET and POST requests to a known path
* Add new config entry for index (*i.e.*, ElasticSearch) URL, backed by an environment variable
  * Add environment variable to the necessary spots to ensure it is properly fixed at build time (per README)
* Handle a FIXME in server.js, preferring `.find()` over `.forEach()`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -59,9 +59,9 @@ jobs:
           name: Build & Register Images
           command: |
             docker build -t ld4p/sinopia_editor:latest .
-            docker build -t ld4p/sinopia_editor:dev --build-arg TRELLIS_BASE_URL=$TRELLIS_BASE_URL_DEV --build-arg SINOPIA_URI=$SINOPIA_URI_DEV --build-arg AWS_COGNITO_DOMAIN=$AWS_COGNITO_DOMAIN_DEV --build-arg COGNITO_CLIENT_ID=$COGNITO_CLIENT_ID_DEV --build-arg COGNITO_USER_POOL_ID=$COGNITO_USER_POOL_ID_DEV .
-            docker build -t ld4p/sinopia_editor:stage --build-arg TRELLIS_BASE_URL=$TRELLIS_BASE_URL_STAGE --build-arg SINOPIA_URI=$SINOPIA_URI_STAGE --build-arg AWS_COGNITO_DOMAIN=$AWS_COGNITO_DOMAIN_STAGE --build-arg COGNITO_CLIENT_ID=$COGNITO_CLIENT_ID_STAGE --build-arg COGNITO_USER_POOL_ID=$COGNITO_USER_POOL_ID_STAGE .
-            docker build -t ld4p/sinopia_editor:prod --build-arg TRELLIS_BASE_URL=$TRELLIS_BASE_URL_PROD --build-arg SINOPIA_URI=$SINOPIA_URI_PROD --build-arg AWS_COGNITO_DOMAIN=$AWS_COGNITO_DOMAIN_PROD --build-arg COGNITO_CLIENT_ID=$COGNITO_CLIENT_ID_PROD --build-arg COGNITO_USER_POOL_ID=$COGNITO_USER_POOL_ID_PROD .
+            docker build -t ld4p/sinopia_editor:dev --build-arg TRELLIS_BASE_URL=$TRELLIS_BASE_URL_DEV --build-arg SINOPIA_URI=$SINOPIA_URI_DEV --build-arg AWS_COGNITO_DOMAIN=$AWS_COGNITO_DOMAIN_DEV --build-arg COGNITO_CLIENT_ID=$COGNITO_CLIENT_ID_DEV --build-arg COGNITO_USER_POOL_ID=$COGNITO_USER_POOL_ID_DEV --build-arg INDEX_URL=$INDEX_URL_DEV .
+            docker build -t ld4p/sinopia_editor:stage --build-arg TRELLIS_BASE_URL=$TRELLIS_BASE_URL_STAGE --build-arg SINOPIA_URI=$SINOPIA_URI_STAGE --build-arg AWS_COGNITO_DOMAIN=$AWS_COGNITO_DOMAIN_STAGE --build-arg COGNITO_CLIENT_ID=$COGNITO_CLIENT_ID_STAGE --build-arg COGNITO_USER_POOL_ID=$COGNITO_USER_POOL_ID_STAGE --build-arg INDEX_URL=$INDEX_URL_STAGE .
+            docker build -t ld4p/sinopia_editor:prod --build-arg TRELLIS_BASE_URL=$TRELLIS_BASE_URL_PROD --build-arg SINOPIA_URI=$SINOPIA_URI_PROD --build-arg AWS_COGNITO_DOMAIN=$AWS_COGNITO_DOMAIN_PROD --build-arg COGNITO_CLIENT_ID=$COGNITO_CLIENT_ID_PROD --build-arg COGNITO_USER_POOL_ID=$COGNITO_USER_POOL_ID_PROD --build-arg INDEX_URL=$INDEX_URL_PROD .
             echo $DOCKER_PASS | docker login -u $DOCKER_USER --password-stdin
             docker push ld4p/sinopia_editor:latest
             docker push ld4p/sinopia_editor:dev
--- a/package.json
+++ b/package.json
@@ -116,6 +116,7 @@
     "react-router-dom": "^4.3.1",
     "redux": "^4.0.1",
     "redux-thunk": "^2.3.0",
+    "request": "^2.88.0",
     "reselect": "^4.0.0",
     "shortid": "^2.2.14",
     "sinopia_server": "^3.0.0-beta4",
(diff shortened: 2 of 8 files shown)


.circleci/config.yml
```diff
@@ -59,9 +59,9 @@ jobs:
...
-            docker build -t ld4p/sinopia_editor:stage --build-arg TRELLIS_BASE_URL=$TRELLIS_BASE_URL_STAGE --build-arg SINOPIA_URI=$SINOPIA_URI_STAGE --build-arg AWS_COGNITO_DOMAIN=$AWS_COGNITO_DOMAIN_STAGE --build-arg COGNITO_CLIENT_ID=$COGNITO_CLIENT_ID_STAGE --build-arg COGNITO_USER_POOL_ID=$COGNITO_USER_POOL_ID_STAGE .
-            docker build -t ld4p/sinopia_editor:prod --build-arg TRELLIS_BASE_URL=$TRELLIS_BASE_URL_PROD --build-arg SINOPIA_URI=$SINOPIA_URI_PROD --build-arg AWS_COGNITO_DOMAIN=$AWS_COGNITO_DOMAIN_PROD --build-arg COGNITO_CLIENT_ID=$COGNITO_CLIENT_ID_PROD --build-arg COGNITO_USER_POOL_ID=$COGNITO_USER_POOL_ID_PROD .
+            docker build -t ld4p/sinopia_editor:dev --build-arg TRELLIS_BASE_URL=$TRELLIS_BASE_URL_DEV --build-arg SINOPIA_URI=$SINOPIA_URI_DEV --build-arg AWS_COGNITO_DOMAIN=$AWS_COGNITO_DOMAIN_DEV --build-arg COGNITO_CLIENT_ID=$COGNITO_CLIENT_ID_DEV --build-arg COGNITO_USER_POOL_ID=$COGNITO_USER_POOL_ID_DEV --build-arg INDEX_URL=$INDEX_URL_DEV .
```
mjgiarlo: FYI, I have already tracked down these values and added new env-specific variables to the CircleCI build.

package.json
```diff
@@ -116,6 +116,7 @@
     "react-router-dom": "^4.3.1",
     "redux": "^4.0.1",
     "redux-thunk": "^2.3.0",
+    "request": "^2.88.0",
```
mjgiarlo: It seems in Javascriptland there are a million options here, and it's not clear to me which is the best fit or the most idiomatic. Happy to consider alternatives.

server.js
```diff
@@ -5,17 +5,51 @@
...
 
 const port = 8000
 const app = express()
 
-app.use(express.static(`${__dirname}/`))
+// Required for ElasticSearch proxy middleware to parse response body as JSON
+app.use(express.json())
+
+app.use('/api/search', (req, res) => {
+  if (!['GET', 'POST'].includes(req.method)) {
```
mjgiarlo: We only allow the editor to hit the search part of ElasticSearch's API, which means either GET or POST could be used. Throw a 400 error if any other HTTP action is attempted.

server.js
```diff
@@ -5,17 +5,51 @@
...
+    res.status(400).json({ error: `unsupported method: ${req.method}` })
+    return
+  }
+
+  // Hard-coded for now since we only have the one use case for proxying ES,
+  // i.e., full-text search of resources
+  if (req.path !== '/sinopia_resources/sinopia/_search') {
```
mjgiarlo: I don't love this approach but I was loath to get into regex or otherwise parsing the request path, lest I leave some wrinkle that could be exploited. I am open to validating requests some other way, to prevent our ElasticSearch instance from getting owned or hosed.

server.js
```diff
@@ -5,17 +5,51 @@
...
 const express = require('express')
+const request = require('request')
+const Config = require('./src/Config')
 const versoSpoof = require('./src/versoSpoof')
 
 const port = 8000
 const app = express()
 
-app.use(express.static(`${__dirname}/`))
+// Required for ElasticSearch proxy middleware to parse response body as JSON
```
mjgiarlo: Note that `server.js` doesn't have test coverage. I'm not sure exactly how or what I'd test. I *could* consider extracting the middleware elsewhere so it's easier to test but that's not super clear to me at the moment.

jermnelson:
Hi @mjgiarlo, I am getting this error both when I rebuild the Editor container and try to run with `docker-compose up editor` as well as trying to run `npm start` directly to test this route: 

```javascript
sinopia_editor/src/Config.js:109
export default Config
^^^^^^
SyntaxError: Unexpected token export  
``` 
I'm hoping this my user error :-).

mjgiarlo:
> Hi @mjgiarlo, I am getting this error both when I rebuild the Editor container and try to run with `docker-compose up editor` as well as trying to run `npm start` directly to test this route:
> 
> ```js
> sinopia_editor/src/Config.js:109
> export default Config
> ^^^^^^
> SyntaxError: Unexpected token export  
> ```
> 
> I'm hoping this my user error :-).

@jermnelson That does not ring a bell. Weird! Happy to pair in the morning.

server.js
```diff
@@ -32,15 +66,9 @@ app.all('/profile-edit/server/whichrt', (req, res) => {
...
-      // FIXME:  there's probably a better way to find the value in array than forEach, but there are only 5 urls
-      let json
-
-      versoSpoof.owlOntUrl2JsonMappings.forEach((el) => {
-        if (reqUri === el.url) {
-          json = el.json
-        }
-      })
-      res.status(200).type('application/json').send(json)
+      const json = versoSpoof.owlOntUrl2JsonMappings.find(el => reqUri === el.url).json
```
ndushay: huzzah for getting rid of FIXME. 👍

server.js
```diff
@@ -5,17 +5,51 @@
...
+      res.status(500).json({ error: 'server error: could not make request to ElasticSearch' })
+    })
+    .pipe(res)
+    .on('error', (err) => {
+      console.error(`error returning ElasticSearch response: ${err}`)
+      res.status(500).json({ error: 'server error: could not send ElasticSearch response' })
+    })
 })
 
+app.use(express.static(`${__dirname}/`))
```
ndushay: I don't get what this line is doing

server.js
```diff
@@ -32,15 +66,9 @@ app.all('/profile-edit/server/whichrt', (req, res) => {
...
-
-      versoSpoof.owlOntUrl2JsonMappings.forEach((el) => {
-        if (reqUri === el.url) {
-          json = el.json
-        }
-      })
-      res.status(200).type('application/json').send(json)
+      const json = versoSpoof.owlOntUrl2JsonMappings.find(el => reqUri === el.url).json
+
+      res.json(json)
```
ndushay: can you say more about this line vs old line 43 ?   is status 200 the default?  Is application/json the default type?

server.js
mjgiarlo: :man_dancing:

server.js
mjgiarlo: @ndushay This middleware is for serving static assets to the browser, such as the stuff in `src/styles/` and `static/`. I'm not sure how "right" this code is given how we're using it, tbh, and I didn't add it—it was shifted around by my work. This code was added 4 years ago by LC: https://github.com/LD4P/sinopia_editor/commit/fb13bb2825df0e46e6e671d9258f4c957182cea0#diff-dc2704ac83baf8775944bab1251b45abR9

If we think we should tighten up how we serve static assets... new issue?

server.js
mjgiarlo: @ndushay 	Oh, shoot, I meant to leave a comment about this line but missed it. Sorry! Sure.

We can safely shorten `res.status(200).type('application/json').send(json)` to `res.json(json)` because Express will return an HTTP 200 OK by default, and calling `.json()` automatically sets the content type (to `application/json`), sends the content to the client, and ends the request/response cycle.

package.json
jermnelson: This module `request` is fine.

server.js
ndushay: how about just adding a comment:

```
//serve static assets to the browser, e.g. from src/styles/ and static/
```

And mayhaps making a ticket to move "styles" out of "src" directory??

mjgiarlo: I was/am hoping someone with more knowledge of express could write up the ticket. I'm reluctant to pile that on the backlog without a clearer sense of the problem or the solution. (Sorry, still recovering from Hyrax backlog stress, where anyone and everyone tossed issues that were ill defined and often just questions.)

ndushay: ok.  I can live w/o the ticket, but would appreciate the comment.

mjgiarlo: @ndushay ok.

mjgiarlo: @ndushay :boom: https://github.com/LD4P/sinopia_editor/pull/887 :boom:

PR #872, opened 2019-07-03
Creates resource state builder, for transforming RDF to resource state.

### LINKED ISSUE #867: Transform RDF graph to Redux state
Given a graph, transform to the appropriate Redux state structure. This is the inverse of GraphBuilder.

Supports #808

justinlittman:
closes #867.

### DIFF
--- a/__tests__/ResourceStateBuilder.test.js
+++ b/__tests__/ResourceStateBuilder.test.js
@@ -1,0 +1,119 @@
+// Copyright 2019 Stanford University see LICENSE for license
+
+import ResourceStateBuilder from 'ResourceStateBuilder'
+import shortid from 'shortid'
+import { rdfDatasetFromN3 } from 'Utilities'
+
+describe('ResourceStateBuilder', () => {
+  it('builds the state for literal properties', async () => {
+    const resource = `<http://example/123> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://id.loc.gov/ontologies/bibframe/Note> .
+  <http://example/123> <http://www.w3.org/ns/prov#wasGeneratedBy> "resourceTemplate:bf2:Note" .
+  <http://example/123> <http://www.w3.org/2000/01/rdf-schema#label> "foo"@en .
+  <http://example/123> <http://www.w3.org/2000/01/rdf-schema#label> "bar"@en .`
+
+    const dataset = await rdfDatasetFromN3(resource)
+
+    shortid.generate = jest.fn().mockReturnValue('abc123')
+
+    const builder = new ResourceStateBuilder(dataset, 'http://example/123')
+    const resourceState = builder.state
+    expect(resourceState).toEqual({
+      'resourceTemplate:bf2:Note': {
+        'http://www.w3.org/2000/01/rdf-schema#label': {
+          items: [
+            {
+              id: 'abc123',
+              content: 'foo',
+              lang: {
+                items: [
+                  {
+                    id: 'en',
+                  },
+                ],
+              },
+            },
+            {
+              id: 'abc123',
+              content: 'bar',
+              lang: {
+                items: [
+                  {
+                    id: 'en',
+                  },
+                ],
+              },
+            },
+          ],
+        },
+      },
+    })
+  })
+
+  it('builds the state for URI resource properties', async () => {
+    const resource = `<http://example/123> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://id.loc.gov/ontologies/bibframe/Note> .
+  <http://example/123> <http://www.w3.org/ns/prov#wasGeneratedBy> "resourceTemplate:bf2:Note" .
+  <http://example/123> <http://rdaregistry.info/Elements/i/P40021> <http://id.loc.gov/vocabulary/organizations/wauar> .`
+
+    const dataset = await rdfDatasetFromN3(resource)
+
+    shortid.generate = jest.fn().mockReturnValue('abc123')
+
+    const builder = new ResourceStateBuilder(dataset, 'http://example/123')
+    const resourceState = builder.state
+
+    expect(resourceState).toEqual({
+      'resourceTemplate:bf2:Note': {
+        'http://rdaregistry.info/Elements/i/P40021': {
+          items: [
+            {
+              id: 'abc123',
+              uri: 'http://id.loc.gov/vocabulary/organizations/wauar',
+            },
+          ],
+        },
+      },
+    })
+  })
+
+  it('builds the state for embedded resource property', async () => {
+    const resource = `<http://example/123> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://id.loc.gov/ontologies/bibframe/Note> .
+  <http://example/123> <http://www.w3.org/ns/prov#wasGeneratedBy> "resourceTemplate:bf2:Note" .
+  <http://example/123> <http://id.loc.gov/ontologies/bibframe/note> _:b1 .
+  _:b1 <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://id.loc.gov/ontologies/bibframe/Note> .
+  _:b1 <http://www.w3.org/ns/prov#wasGeneratedBy> "resourceTemplate:bf2:Note" .
+  _:b1 <http://www.w3.org/2000/01/rdf-schema#label> "foobar"@en .`
+
+    const dataset = await rdfDatasetFromN3(resource)
+
+    shortid.generate = jest.fn().mockReturnValue('abc123')
+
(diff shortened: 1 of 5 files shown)


__tests__/ResourceStateBuilder.test.js
```diff
@@ -0,0 +1,138 @@
+// Copyright 2019 Stanford University see LICENSE for license
+
+import ResourceStateBuilder from '../src/ResourceStateBuilder'
```
jcoyne: You should be able to use an absolute url here: `import ResourceStateBuilder from 'ResourceStateBuilder'`

src/ResourceStateBuilder.js
```diff
@@ -0,0 +1,104 @@
...
+   * @return {Object} the resource represented as Redux state
+   */
+  get state() {
+    this.resourceState = this.buildResource(rdf.namedNode(this.resourceURI))
+
+    return this.resourceState
+  }
+
+  /**
+   * @param {rdf.Term} NamedNode or BlankNode of resource to be built
```
jcoyne: add the parameter name

src/ResourceStateBuilder.js
```diff
@@ -0,0 +1,104 @@
...
+const rdf = require('rdf-ext')
+const _ = require('lodash')
+
+/**
+ * Builds Redux state from an RDF graph
+ */
+export default class ResourceStateBuilder {
+  /**
+   * @param {Dataset} dataset the RDF graph
+   * @param {string} URI of the resource
```
jcoyne: add the parameter name

__tests__/ResourceStateBuilder.test.js
```diff
@@ -0,0 +1,138 @@
...
+import shortid from 'shortid'
+
+const rdf = require('rdf-ext')
+const ParserN3 = require('@rdfjs/parser-n3')
```
jcoyne: Did you consider using `n3/lib/N3Parser` instead?  That's what I used here: https://github.com/LD4P/sinopia_editor/pull/808/files#diff-a8ce107dbeb5f2fb8b01abcfc5e7fa02R10 so if we use the same libraries it'll be easier to stitch these together.

mjgiarlo: `n3/lib/N3Parser` is also what we use in `sinopia_acl` so :+1:

src/ResourceStateBuilder.js
```diff
@@ -0,0 +1,104 @@
+// Copyright 2019 Stanford University see LICENSE for license
+
+import shortid from 'shortid'
+
+const rdf = require('rdf-ext')
```
mjgiarlo: Is `rdf-ext` one of the libraries that doesn't function when loaded via `import rdf from 'rdf-ext'`?

src/ResourceStateBuilder.js
mjgiarlo: e.g., `@param {string} resourceURI - the URI of the resource`

src/ResourceStateBuilder.js
```diff
@@ -0,0 +1,104 @@
...
+        const item = this.buildItem(quad)
+
+        thisResourceState[rtId][propertyURI].items.push(item)
+      }
+    })
+    return thisResourceState
+  }
+
+  /**
+   * @param {rdf.Term} NamedNode or BlankNode of resource to find
```
mjgiarlo: as above, add the param name after the type declaration

src/ResourceStateBuilder.js
justinlittman: Not sure. Just cribbed this from elsewhere in the codebase.

__tests__/ResourceStateBuilder.test.js
justinlittman: Fixed.

src/ResourceStateBuilder.js
justinlittman: Added.

__tests__/ResourceStateBuilder.test.js
justinlittman: Switched and added a helper function to Utilities.js for loading.

package.json
```diff
@@ -94,6 +94,7 @@
     "@fortawesome/fontawesome-svg-core": "^1.2.17",
     "@fortawesome/free-solid-svg-icons": "^5.8.1",
     "@fortawesome/react-fontawesome": "^0.1.4",
+    "@rdfjs/parser-n3": "^1.1.2",
```
jcoyne: Remove this now?

src/Utilities.js
```diff
@@ -64,3 +66,23 @@ export const getLookupConfigItems = (propertyTemplate) => {
 
   return templateConfigItems
 }
+
+/**
+ * Loads N3 into a dataset.
+ * @param {string} data that is the N3
+ * @return {Promise<rdf.Dataset>} a promise that resolves to the loaded dataset
+ */
+export const loadN3 = data => new Promise((resolve) => {
```
mjgiarlo: Is this only, and will this only be, used in the test suite? If so, would you consider moving this to a new helper file in `__tests__/`? I think we have 1-2 examples of those in there.

justinlittman: It currently is used only once. However, I anticipate that @jcoyne will use it in Trellis loading, as he has a similar method in https://github.com/LD4P/sinopia_editor/pull/808

package.json
justinlittman: Fixed.

src/Utilities.js
mjgiarlo: okie dokie. let's leave it here then!

src/ResourceStateBuilder.js
justinlittman: Done.

mjgiarlo:
@jcoyne 	Can you have another look at the PR?

src/ResourceStateBuilder.js
```diff
@@ -0,0 +1,104 @@
...
+    return this.resourceState
+  }
+
+  /**
+   * @param {rdf.Term} resourceTerm NamedNode or BlankNode of resource to be built
+   * @return {Object} the resource represented as Redux state
+   */
+  buildResource(resourceTerm) {
+    const thisResourceState = {}
+    // Find the resource template id. Should be only 1.
```
ndushay: some of these comments don't pass this Sandi Metz test:  if you read the code line out loud, and it says what your comment says, the comment is superfluous.   I think this could be like below `// expect exactly 1 rtid` and you can get rid of `//find the properties` comment below and `// add each property`

src/Utilities.js
```diff
@@ -64,3 +66,23 @@ export const getLookupConfigItems = (propertyTemplate) => {
 
   return templateConfigItems
 }
+
+/**
+ * Loads N3 into a dataset.
```
ndushay: into an rdf.Dataset?  into an rdf-ext Dataset object?  And it's actually loading ttl as a string, no?

src/Utilities.js
ndushay: maybe add a comment re what you discussed, so we can take the knowledge forward:  ` * expect this to be used in Trellis loading, but currently only used for tests`  as that may help with future refactoring?

src/ResourceStateBuilder.js
```diff
@@ -0,0 +1,104 @@
...
+      }
+    })
+    return thisResourceState
+  }
+
+  /**
+   * @param {rdf.Term} resourceTerm NamedNode or BlankNode of resource to find
+   * @return {string} the resource template id
+   */
+  findResourceTemplateId(resourceTerm) {
```
ndushay: if some of these methods are, effectively, private methods and not part of the API of this class, can they be so denoted?  We don't have enough context to understand how this class should be expected to be used.

src/Utilities.js
ndushay: I wonder if improving on this name would be useful.  It *seems* like it is expecting ttl as a string to be passed in, and what comes out is an rdf-ext.Dataset object.  So "loadN3" isn't terribly obvious to me.  `rdfDatasetFromTtl` ??  "N3" is both an rdf format and the name of a package, which is even more confusing.

src/Utilities.js
jcoyne: an rdf-ext Dataset is an implementation of rdf.Dataset.

src/ResourceStateBuilder.js
jcoyne: https://babeljs.io/blog/2018/12/03/7.2.0

justinlittman: This doesn't seem fully baked yet. I (and others) couldn't get eslint to handle the `#` syntax.

justinlittman: As an alternative, I added example usage to the class docs. This makes it clear what the API of the class is.

PR #878, opened 2019-07-03
Remove defaults on template that has valueTemplateRefs

jcoyne:
It doesn't make sense to have both properties set for a template


jcoyne:
@justinlittman I created https://github.com/LD4P/sinopia_editor/issues/879 for that.

ndushay:
@jcoyne @justinlittman -- does this impact the older blocked PR #829?  Should that be closed and tickets created for its pieces?

jcoyne:
@ndushay we'll just rebase that one.

PR #885, opened 2019-07-03
Rename refreshResourceTemplate to the more appropriate refreshPropert…

jcoyne:
…yTemplate


jcoyne:
@ndushay I think the problem is that the tests don't pass a realistic payload.  If you read the code you can see the method requires a property template in the payload: https://github.com/LD4P/sinopia_editor/blob/507fc1cd725dd41cb56b6d7d5efcd08bec98c211/src/reducers/index.js#L45

PR #899, opened 2019-07-08
Extract the model from the view logic

jcoyne:
This introduces a UI regression, where entering a value in a collapsable component, collapses the component rather than keeping it open.


jermnelson:
Merging and created issue https://github.com/LD4P/sinopia_editor/issues/900 to note the regression.

PR #906, opened 2019-07-09
simplify findNode conditional logic

jmartin-sul:
the current intent seems to be safe access to `obj`, and return of `undefined` if `obj[key]` is undefined.  it seems like the logic replaced by this commit does that, albeit likely in an unintended way (if `obj[key]` was undefined, it'd have been returned, but by the if branch of the replaced ternary operation instead of the else branch)..


jmartin-sul:
h/t to @mjgiarlo for noticing this

mjgiarlo:
@jmartin-sul :clap:

PR #859, opened 2019-06-28
882 show version

peetucket:
fixes #882

PR #909, opened 2019-07-10
NewsItem: get the version number from package.json

jmartin-sul:
so presentation on home page is always up to date, instead of hardcoding it independently

(noticed this when reviewing https://github.com/LD4P/sinopia_editor/pull/892/files#diff-5083dfea3973f64f536edf318a38b904R9, had it in my notes as a small todo)

### DIFF
--- a/__tests__/components/home/NewsItem.test.js
+++ b/__tests__/components/home/NewsItem.test.js
@@ -3,6 +3,7 @@
 import React from 'react'
 import { shallow } from 'enzyme'
 import NewsItem from 'components/home/NewsItem'
+import { version } from '../../../package.json'
 
 describe('<NewsItem />', () => {
   const wrapper = shallow(<NewsItem />)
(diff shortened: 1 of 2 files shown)


jmartin-sul:
i just ripped off what @peetucket did here:  https://github.com/LD4P/sinopia_editor/pull/859/files#diff-f880b57906a072cdffa62da2c954ddd4R31

__tests__/components/home/NewsItem.test.js
```diff
@@ -16,4 +17,14 @@ describe('<NewsItem />', () => {
...
     })
   })
+
+  it('renders the currently released version', () => {
+    /*
+      not a concern here, since the input is ours:
+      "Detects RegExp(variable), which might allow an attacker to DOS your server with a long-running regular expression"
+      - https://github.com/nodesecurity/eslint-plugin-security#detect-non-literal-regexp
+     */
+    // eslint-disable-next-line security/detect-non-literal-regexp
```
jmartin-sul: yeah, this is kinda gross, and i am totally open to cleaner suggestions.  i know @ndushay in particular dislikes inline style check exceptions, and this is a particularly bad specimen.

but... i didn't want to disable the rule in general (i think it's a good one), i thought it was totally fine here (per explanation about us controlling the var), and i wanted to explain why, so that people wouldn't copypasta this exception into places where the var contained user input.

🤷‍♂

PR #923, opened 2019-07-11
Validates that literal property templates do not have default URIs.

justinlittman:


jcoyne:
This has conflicts now😟

PR #930, opened 2019-07-11
remove subauthorities for LCSH

elrayle:
Per issue https://github.com/LD4P/qa_server/issues/141, subauthorities are being removed for LCSH.


hudajkhan:
One of the files now moved to test fixtures also specifies the LCSH subjects persons subauthority: the following would need to be changed: https://github.com/LD4P/sinopia_editor/blob/master/__tests__/__fixtures__/MonographInstance.json#L165  .  This could be changed to use just the subject authority instead of the persons subauthority for LCSH.

elrayle:
@jermnelson This is rebased and tests are green.  Are you ok with it merging?

Issue #927, opened 2019-07-11
Embedded Resource Templates Defaults not working


jermnelson:
Defaults in an embedded Resource Template are not populating the Redux state and therefore are not showing up in the React components.

For example in the Fixture Resource Template `resourceTemplate:bf2:Monograph:Instance`:

<img width="599" alt="Screen Shot 2019-07-11 at 10 43 05 AM" src="https://user-images.githubusercontent.com/71847/61072725-bcedd280-a3c8-11e9-99ee-e44dca048570.png">
 
The BIBFRAME Work `hasInstance` property uses the `resourceTemplate:bf2:Monograph:Instance` resource template as an embedded resource but the defaults for the property templates like **Mode of Issuance** and **Carrier Type** defaults are not showing up.

justinlittman:
closed by https://github.com/LD4P/sinopia_editor/pull/938

PR #938, opened 2019-07-12
Fix defaults for nested properties and adds tests for stubProperty.

justinlittman:


jcoyne:
@justinlittman does this fix #927 ?

Issue #535, opened 2019-05-20
repeatable resource properties should allow removal 


ndushay:
Right now, you can add additional instances of a repeatable property of type resource ... but you can't delete them once you add them.

PR #953, opened 2019-07-15
Remove button

justinlittman:


justinlittman:
refs #535

justinlittman:
Specifically:
* If there is more than one instance of a resource property, a remove button will be available for all instances of the resource property.
* If a resource property is repeatable, there will be an add button for just the first instance.

PR #957, opened 2019-07-15
INDEX_URL has to be set as an environment variable

jcoyne:
Otherwise the server ends up using the default value in here: https://github.com/LD4P/sinopia_editor/blob/master/src/Config.js#L44

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,6 @@ ARG SINOPIA_URI
 ARG AWS_COGNITO_DOMAIN
 ARG COGNITO_CLIENT_ID
 ARG COGNITO_USER_POOL_ID
-ARG INDEX_URL
 
 # This is the directory the user in the circleci/node image can write to
 WORKDIR /home/circleci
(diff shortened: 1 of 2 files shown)


mjgiarlo:
@jcoyne @aaron-collier Isn't this going to cause a problem given that `process.env` variables are fixed in React apps at build time rather than at runtime? We've been down this road a couple times before and it seems like we're now using two different patterns for env vars in the same app. cc: @jermnelson @jmartin-sul

jmartin-sul:
yeah, if an env var is needed outside of a developer's local laptop instance, i'd expect deployed instances to hit the same snag we've hit in the past w.r.t. what @mjgiarlo is describing.  instructions for whatall needs to be done:
https://github.com/LD4P/sinopia_editor/pull/553/commits/f9806db271e5433a28bb7b2202fcdcb0f84347ab

though i think the `ARG` part in the Dockerfile will be needed too, per the instructions in that commit?

PR #959, opened 2019-07-15
Adds add and remove buttons for toggling properties.

### LINKED ISSUE #956: Remove button for expanded properties
As discussed, the initial toggle of a property involves adding the embedded property to the state, which may have defaults. This ticket is for adding the corresponding remove button.

Note that this does not cover the UX enhancements that Astrid is working on, just adding the remove functionality.

justinlittman:
closes #956

### DIFF
--- a/__tests__/components/editor/property/OutlineHeader.test.js
+++ b/__tests__/components/editor/property/OutlineHeader.test.js
@@ -3,9 +3,7 @@
 import React from 'react'
 import { shallow } from 'enzyme'
 import { faMinusSquare, faPlusSquare } from '@fortawesome/free-solid-svg-icons'
-import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import OutlineHeader from 'components/editor/property/OutlineHeader'
-import PropertyLabel from 'components/editor/property/PropertyLabel'
 
 describe('<OutlineHeader />', () => {
   const property = {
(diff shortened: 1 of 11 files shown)


__tests__/components/editor/property/OutlineHeader.test.js
```diff
@@ -3,44 +3,141 @@
...
     propertyURI: 'http://id.loc.gov/ontologies/bibframe/instanceOf',
     mandatory: 'false',
   }
-  const headerProps = {
-    spacer: 0,
-    collapsed: true,
-    pt: property,
-  }
-  const wrapper = shallow(<OutlineHeader {...headerProps} />)
+  // const mockEvent = { preventDefault: () => {} }
```
jcoyne: Cruft?

src/components/editor/property/OutlineHeader.jsx
```diff
@@ -40,10 +42,41 @@ OutlineHeader.propTypes = {
...
+
+  return {
+    resourceModel,
+    property,
+    collapsed: !isExpanded(state.selectorReducer, ourProps.reduxPath),
+  }
+}
+
+const mapDispatchToProps = (dispatch, ownProps) => ({
+  handleTogglePlusButton: (event) => {
```
jcoyne: As this is not a "plus" button anymore, can you please change the name to "handleToggleCollapse" or something without "plus" in it.

src/components/editor/property/PropertyTemplateOutline.jsx
```diff
@@ -60,7 +54,7 @@ PropertyTemplateOutline.propTypes = {
   property: PropTypes.object.isRequired,
   resourceModel: PropTypes.object,
   collapsed: PropTypes.bool,
-  handleTogglePlusButton: PropTypes.func,
+  // handleTogglePlusButton: PropTypes.func,
```
jcoyne: Cruft?

src/reducers/index.js
```diff
@@ -54,8 +54,8 @@ export const appendResource = (state, action) => {
 export const removeResource = (state, action) => {
   const newState = _.cloneDeep(state)
   const reduxPath = action.payload
-  const key = reduxPath.slice(-2)[0]
-  const parentReduxPath = reduxPath.slice(0, reduxPath.length - 2)
+  const key = reduxPath.slice(-1)[0]
```
jcoyne: It's odd that this didn't require a change to the test for removeResource.   Do we have a good test there?

__tests__/components/editor/property/OutlineHeader.test.js
```diff
@@ -3,44 +3,141 @@
...
+    })
 
-    expect(faWrapper.getElement(0).props.icon).toEqual(faMinusSquare)
+    it('has a remove button', () => {
+      expect(wrapper.exists('button.btn-remove')).toEqual(true)
+    })
   })
+
+
+  // it('contains a FontAwesomeIcon', () => {
```
jcoyne: Cruft?

src/reducers/index.js
justinlittman: See the change in __tests__/reducers/index.test.js

justinlittman:
@jcoyne Removed cruft and renamed function.

src/components/editor/property/OutlineHeader.jsx
```diff
@@ -40,10 +42,41 @@ OutlineHeader.propTypes = {
...
   handleCollapsed: PropTypes.func,
   id: PropTypes.string,
-  isRequired: PropTypes.any,
-  isAdd: PropTypes.bool,
-  pt: PropTypes.object,
-  spacer: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
+  property: PropTypes.object,
+  reduxPath: PropTypes.array,
+  resourceModel: PropTypes.object,
+  handleToggleAddButton: PropTypes.func,
```
jcoyne: Why is it `handleToggleAddButton` rather than `handleAddButton`?  Is it toggleable?

justinlittman: It actually handles 2 buttons: the toggle (plus) and the add button (which is add and then toggle).

jcoyne: To me toggle means open and close.  So, `handleAddAndOpen` is preferable in that case.  I think we could also call this `handleAdd` as that is the users intent.  The fact that "add" has two actions associated with it is the business of a the reducer.

justinlittman: Renamed.

PR #964, opened 2019-07-16
Extract the PropertyTemplate shape to a module

jcoyne:


jcoyne:
@justinlittman Sounds good. I've made that change.

PR #978, opened 2019-07-17
Sets correct container when saving rdf.

### LINKED ISSUE #973: Save is not using the selected container
This only happens for cornell (the first element in the list)

It's using something like this:
http://platform:8080/repository/ld4p/c36b114f-3b3b-4fce-811c-f657e604ce1f

justinlittman:
closes #973

### DIFF
--- a/__tests__/components/editor/GroupChoiceModal.test.js
+++ b/__tests__/components/editor/GroupChoiceModal.test.js
@@ -17,14 +17,15 @@ describe('<GroupChoiceModal />', () => {
       },
     },
   })
-  const rdfFunc = jest.fn()
+  const rdfFunc = jest.fn().mockReturnValue('some triples')
   const closeFunc = jest.fn()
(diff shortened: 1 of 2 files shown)


__tests__/components/editor/GroupChoiceModal.test.js
```diff
@@ -91,6 +92,7 @@ describe('<GroupChoiceModal />', () => {
     it('attempts to save the RDF content with group choice when save is clicked and closes the modal', () => {
       wrapper.find('.btn-primary', { text: 'Save' }).simulate('click')
       expect(server.publishRDFResource).toHaveBeenCalled()
```
jcoyne: I think you can remove this line as you're testing the same + more on the following line.

justinlittman: Removed.

PR #976, opened 2019-07-17
Add search tab to editor

aaron-collier:
Closes #583

This replaces the browse tab with a search tab that allows for searching against elasticsearch. A non-paged list of resources is displayed (if any found) and when clicking on a result it loads the resource into the editor.

PR #985, opened 2019-07-17
Add tls and net to the webpack node config so that request can be used

aaron-collier:
This is something I ran into while working on search and may be an issue elsewhere so pulling into a separate PR.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -35,6 +35,8 @@ module.exports = {
   },
   node: {
     fs: 'empty',
+    net: 'empty',
+    tls: 'empty',
   },
   resolve: {
     extensions: ['*', '.js', '.jsx'],


jcoyne:
Can you explain more why we need this and how it fixes the problem?

aaron-collier:
@jcoyne sure. I was seeing a lot of `Error: Cannot resolve module net` and `Error: Cannot resolve module tls` when using `request` in sinopiaServer (https://github.com/LD4P/sinopia_editor/pull/976/files#diff-66941f9cc553e538ea7a216facf360b6R188) in order to execute the search. I found this issue on request: https://github.com/request/request/issues/1529

Perhaps there is a better way to execute the query?

aaron-collier:
@jcoyne it looks like I may be able to use `fetch` instead of `request` (as you did in the InputLookup), I'm happy to try that before this is merged to see if it can be avoided.

aaron-collier:
Not having any luck with fetch, yet...

jcoyne:
@aaron-collier request is a node module (for server side requests).  You wouldn't want to use that in client side code.  Maybe we should split the server into it's own repository (or directory at least)

aaron-collier:
Ah... Ok, I think I've figured out how to query via parameter vs. form, so I think fetch will work and I can use the pattern you have in sinopialookup.

aaron-collier:
Closing per discussion above.

PR #986, opened 2019-07-17
allow a user to enter a literal on InputQA

### LINKED ISSUE #866: Allow user to enter a URI or Literal for InputLookupQA
The `InputLookupQA` component does a typeahead search that returns a JSON payload from the QA server. If the returns results are zero or the results aren't what the user wants, the user should be able to enter a known URI as a value. 

First we search for a value in QA and it isn't found:
Screen Shot 2019-07-02 at 10 47 30 AM

Next, enter a URI (which is possible) but hitting enter should make the value persistent in the component:
Screen Shot 2019-07-02 at 10 48 18 AM

peetucket:
fixes #866 

WIP

note: remove console logging when done

### DIFF
--- a/src/components/editor/property/InputLookupQA.jsx
+++ b/src/components/editor/property/InputLookupQA.jsx
@@ -28,18 +28,29 @@ class InputLookupQA extends Component {
 
   // Render menu function to be used by typeahead
   renderMenuFunc = (results, menuProps) => {
+
     const items = []
+    const custom_options = []
     let menuItemIndex = 0
 
+    console.log('RESULTS',results)
     /*
      * Returning results per each promise
      * If error is returned, it will be used to display for that source
      */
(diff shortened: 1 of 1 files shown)


peetucket:
Ideas:

- [ ] set the `allowNew` attribute to true for the typeahead component
- [ ] catch the `.customOption` on the result and push a new menu item with appropriate values

Currently doesn't work when `allowNew` is set to true.  Possible leads to get this to work:

- [ ] ensure that the `options` state value gets added to correctly with an `authLabel` and `authURI` for the custom option?  Because right now the value of `options` state value is set in the `Promise` (near line 180) and so any custom option is not added.  When we render the menu, we get a blowup from missing URIs?

peetucket:
superseded by #995

PR #987, opened 2019-07-17
Clear errors when loading a resource template

### LINKED ISSUE #945: error message when opening RT is not being cleared when you try to open another RT
on DEV, I tried to open the RT sinopia:resourceTemplate:bf2:Agents:Addresses and got an error, as expected:
Screen Shot 2019-07-12 at 11 45 58 AM
then I clicked back to the Resource Templates tab and tried to open a different RT, my:resourcetemplate, and got the same error, referring to the RT I had just tried to open before

if I do a hard-refresh and then try opening my:resourcetemplate again, it opens successfully

I'm using Safari

jcoyne:
Fixes #945

### DIFF
--- a/src/reducers/index.js
+++ b/src/reducers/index.js
@@ -95,6 +95,11 @@ export const setRetrieveError = (state, action) => {
   return newState
 }
 
+export const clearRetrieveError = (state) => {
+  const newEditor = { ...state.editor, serverError: undefined }
+  return { ...state, editor: newEditor }
+}
+
 export const saveAppVersion = (state, action) => {
   const newState = { ...state }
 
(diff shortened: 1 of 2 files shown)


src/reducers/index.js
```diff
@@ -125,6 +130,7 @@ const handlers = {
   SET_ITEMS: setItemsOrSelections,
   CHANGE_SELECTIONS: setItemsOrSelections,
   RETRIEVE_ERROR: setRetrieveError,
+  RETRIEVE_RESOURCE_TEMPLATE_STARTED: clearRetrieveError,
```
justinlittman: This could potentially cause confusion. Would prefer to have these aligned.

jcoyne: @justinlittman I disagree.   The message is "what happened" and the method is "what to do now that that event occurred".   When X happens, then do Y.

justinlittman: OK, then we can agree to disagree.

PR #989, opened 2019-07-18
Use the URL from the stdlib

jcoyne:


aaron-collier:
I'm getting an `INVALID URL` error on this. Looking now.

PR #993, opened 2019-07-18
Include the searchUri in new URL()

aaron-collier:
Getting an `Invalid URL` on new URL() without this.

### DIFF
--- a/server.js
+++ b/server.js
@@ -45,7 +45,7 @@ app.use('/api/search', (req, res) => {
 
   let searchUri = `${Config.indexUrl}${req.path}`
   if (!_.isEmpty(req.query)) {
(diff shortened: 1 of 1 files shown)


server.js
```diff
@@ -45,7 +45,7 @@ app.use('/api/search', (req, res) => {
 
   let searchUri = `${Config.indexUrl}${req.path}`
   if (!_.isEmpty(req.query)) {
-    searchUri += new URL(req.originalUrl).search
+    searchUri += new URL(`${searchUri}${req.originalUrl}`).search
```
jcoyne: I'm not understanding why this is necessary.  Can you explain?

jcoyne: Ah, I see. I've used this idea, but using the req uri components instead in my branch.

PR #994, opened 2019-07-18
Fixes loading of existing resources.

### LINKED ISSUE #979: Loading Trellis URI through Browse tab not using correct Resource Template
Staring with the following ISMN display with a saved Trellis URI, all properties have at least one value:
Screen Shot 2019-07-17 at 10 39 17 AM

Taking the Trellis URL [https://trellis.development.sinopia.io/repository/pcc/35b3069d-28c3-4e66-a375-bf8517d9b8bd](https://trellis.development.sinopia.io/repository/pcc/35b3069d-28c3-4e66-a375-bf8517d9b8bd), clicking on the Browse tab and loading this URL:
Screen Shot 2019-07-17 at 10 46 38 AM

Results in the following Editor display using the Note `sinopia:resourceTemplate:bf2:Identifiers:Note` template instead of the  
`sinopia:resourceTemplate:bf2:Identifiers:ISMN` resource template.
Screen Shot 2019-07-17 at 10 47 42 AM

justinlittman:
closes #979

### DIFF
--- a/__tests__/ResourceStateBuilder.test.js
+++ b/__tests__/ResourceStateBuilder.test.js
@@ -117,3 +117,63 @@ describe('ResourceStateBuilder', () => {
     })
   })
 })
+
+it('handles null root nodes', async () => {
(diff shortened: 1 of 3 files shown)


justinlittman:
The problem is that rdf.dataset.match() does not match correctly when a null named node is provided as the subject. (It matches both null named nodes and blank nodes.)

To address this, additional filtering on subjects is added when performing a match.

PR #995, opened 2019-07-18
Adds option for new uri to InputLookupQA.

### LINKED ISSUE #866: Allow user to enter a URI or Literal for InputLookupQA

justinlittman:
closes #866


peetucket:
- [ ] Do we need to validate entry as a URI before rendering in menu?
- [ ] A test that generates RDF is failing, suggesting there may be a problem here

PR #1003, opened 2019-07-18
Extract LanguageButton from InputLiteral

jcoyne:


jcoyne:
@justinlittman Fixed now that we're using redux

PR #1012, opened 2019-07-19
Load the language data once into entities 

jcoyne:
Previously it was loaded every time the language popup was activated.

### DIFF
--- a/src/actionCreators/languages.js
+++ b/src/actionCreators/languages.js
@@ -1,0 +1,19 @@
+// Copyright 2019 Stanford University see LICENSE for license
+
+import { loadingLanguages, languagesReceived } from 'actions/entities'
+
+const loadLanguages = () => (dispatch, getState) => {
+  if (getState().selectorReducer.entities.languages) {
+    return // Languages already loaded
+  }
+
+  dispatch(loadingLanguages())
+  fetch('https://id.loc.gov/vocabulary/iso639-1.json')
+    .then(resp => resp.json())
+    .then((json) => {
+      dispatch(languagesReceived(json))
+    })
+    .catch(() => false)
+}
+
+export default loadLanguages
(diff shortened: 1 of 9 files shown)


src/actionCreators/languages.js
```diff
@@ -0,0 +1,19 @@
...
+    return // Languages already loaded
+  }
+
+  dispatch(loadingLanguages())
+  fetch('https://id.loc.gov/vocabulary/iso639-1.json')
+    .then(resp => resp.json())
+    .then((json) => {
+      dispatch(languagesReceived(json))
+    })
+    .catch(() => false)
```
justinlittman: Log the error?

src/components/editor/property/InputLang.jsx
```diff
@@ -29,66 +24,19 @@ class InputLang extends Component {
...
-                    isLoading: false,
-                    options: this.createOptions(json),
-                  })
-                })
-                .catch(() => false)
-            }}
-            onBlur={() => {
-              this.setState({ isLoading: false })
-            }}
+            onFocus={() => this.props.loadLanguages()}
```
justinlittman: How about loading in App.componentDidMount()? That way it is (1) preloaded and (2) available for rendering language button when loading a resource.

src/actionCreators/languages.js
jcoyne: This is just moving this code here. This is exactly how it was written when it was in https://github.com/LD4P/sinopia_editor/pull/1012/files#diff-2435c181d99c6352e3e1322a7ce3eb43L85

src/components/editor/property/InputLang.jsx
jcoyne: That's fine as a subsequent change, but this is how it was done previously.

PR #1020, opened 2019-07-19
Get rid of items in redux state

jcoyne:
This would allow us to address individual literal items using redux path (without `find()`)

This is only 1/2 done.


jcoyne:
Superseded by #1108 1108

PR #1021, opened 2019-07-19
Adds recursive validation. Fixes display of validation errors.

justinlittman:


jermnelson:
Looks good; smoke testing works as expected. So an edge case oddity, using the BIBFRAME Instance fixture, after loading and clicking the **Save and Publish** button the expected validation errors appear:
Screen Shot 2019-07-23 at 10 40 11 AM

The Item Information **Barcode** is displaying the Required label as expect:
Screen Shot 2019-07-23 at 10 41 58 AM

Clicking the **Add** button for the Barcode adds and expands the Barcode but the disabled field is highlighted with the Required error message even though by clicking and adding the Barcode resource template I added a value. 
Screen Shot 2019-07-23 at 10 43 04 AM

This isn't a blocker because clicking on the **Save and Publish** again now clears out the barcode error
Screen Shot 2019-07-23 at 10 45 18 AM
 from the list of errors:

PR #1027, opened 2019-07-22
Clears the resourceURIMessage when clicking back into the editor screen

### LINKED ISSUE #1014: ResourceURIMessage persists after loading new Resource Template
Creating an publishing an ISSN displays the expected `ResourceURIMessage` as seen here:
Screen Shot 2019-07-19 at 11 06 28 AM

Loading a different resource template still have that message persisting:
Screen Shot 2019-07-19 at 11 09 10 AM

aaron-collier:
Fixes #1014


package-lock.json
```diff
@@ -6560,28 +6560,27 @@
       "dependencies": {
         "abbrev": {
           "version": "1.1.1",
-          "resolved": false,
+          "resolved": "",
```
justinlittman: Why are there changes to package-lock.json?

src/actionCreators/messages.js
```diff
@@ -0,0 +1,13 @@
+// Copyright 2019 Stanford University see LICENSE for license
+
+import { clearingResourceURIMessage } from 'actions/index'
+
+const clearResourceURIMessage = () => (dispatch, getState) => {
```
justinlittman: Why is there an actionCreator? The action isn't asynchronous.

package-lock.json
aaron-collier: This happens ALL THE TIME... I accidentally checked it in... ugh...

src/actionCreators/messages.js
aaron-collier: I think I can simplify this into just the template. I was following the pattern used in `loadLanguages` but it seems unnecessary.

package-lock.json
aaron-collier: Removed these changes.

src/reducers/index.js
```diff
@@ -3,8 +3,8 @@
 import { combineReducers } from 'redux'
 import authenticate from './authenticate'
 import {
-  removeMyItem, setItemsOrSelections, setBaseURL, showResourceURIMessage, setMyItemsLang,
-  showGroupChooser, closeGroupChooser, showRdfPreview,
+  removeMyItem, setItemsOrSelections, setBaseURL, showResourceURIMessage, clearResourceURIMessage,
+  setMyItemsLang, showGroupChooser, closeGroupChooser, showRdfPreview,
```
ndushay: for some reason, "setMyItemsLang" makes me think of "how's my driving?"

PR #1028, opened 2019-07-22
Correctly handles nested properties that are mandatory and have defau…

### LINKED ISSUE #1022: Nested defaults not added to redux state
Steps to reproduce:
1. Open a new Bibframe instance.
2. Instance the redux state.

Expected result:
`resource > resourceTemplate:bf2:Monograph:Instance >http://id.loc.gov/ontologies/bibframe/itemPortion > <key> > resourceTemplate:bf2:Identifiers:Barcode > http://www.w3.org/1999/02/22-rdf-syntax-ns#value > items` should include the default _12345_.

Actual result:
State is only `resource > resourceTemplate:bf2:Monograph:Instance >http://id.loc.gov/ontologies/bibframe/itemPortion > <key> > resourceTemplate:bf2:Identifiers:Barcode`.

justinlittman:
…lts.

closes #1022

### DIFF
--- a/src/actionCreators/resources.js
+++ b/src/actionCreators/resources.js
@@ -1,5 +1,5 @@
 // Copyright 2019 Stanford University see LICENSE for license
-/* eslint max-params: ["warn", 6] */
+/* eslint max-params: ["warn", 7] */
 
 import {
   assignBaseURL, updateStarted, updateFinished,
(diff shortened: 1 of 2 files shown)


justinlittman:
Here is a nested property for a resource that is mandatory and has defaults:

src/actionCreators/resources.js
```diff
@@ -136,46 +137,63 @@ const stubResourceProperties = async (resourceTemplateId, existingResourceTempla
...
+
+        if (newResource[propertyTemplate.propertyURI] === undefined) {
+          newResource[propertyTemplate.propertyURI] = {}
+        }
+        // For each value template
+        // Since these are promises, using Promise.all for https://stackoverflow.com/questions/37576685/using-async-await-with-a-foreach-loop
+        await Promise.all(
+          propertyTemplate.valueConstraint.valueTemplateRefs.map(async (resourceTemplateId) => {
+            const nestedResourceTemplatePromise = fetchResourceTemplate(resourceTemplateId, dispatch)
+            // See if there is alread a <key> > <resource template id> for this resource template id
```
jermnelson: misspelling, should be `already`

justinlittman: Fixed @jermnelson

PR #1030, opened 2019-07-22
wrap all columns in resource template table

### LINKED ISSUE #847: Resource Template list: id column should wrap
per https://github.com/LD4P/sinopia_editor/issues/802#issuecomment-506085591, a long ID string will bleed into the author column:

Screen Shot 2019-06-26 at 7 51 55 PM

would be nice to style the text so that it wraps instead, and stays in its own col.

peetucket:
fixes #847


peetucket:
fixed by #1029

PR #1038, opened 2019-07-23
Oops, adds validation error display to InputURI.

justinlittman:


justinlittman:
I forgot to do as part of #1021.

PR #1050, opened 2019-07-24
Adds error message when user enters a literal in InputURI.

### LINKED ISSUE #944: when cataloger enters a literal where a URI is expected, give cataloger some kind of feedback telling them to enter a URI
in DEV, I made a Resource Template my:resourcetemplate
with the first property template of type = resource, and no target/nested RT specified, so it expects the cataloger to enter a URI

when cataloger sees this field, it looks like a literal field. but when cataloger enters a literal value, and hits enter, the value is not "entered" like one would expect nor does it show up in preview

the cataloger has no way of knowing they're supposed to be entering a URI. 

can we tell them somehow? I don't think a tooltip is best solution because profile creator may already have added their own tooltip (remark) specific to this field

justinlittman:
closes #944


justinlittman:
Screen Shot 2019-07-24 at 6 07 00 PM

PR #1052, opened 2019-07-25
Adds notification when saving and publishing a resource.

### LINKED ISSUE #1032: Provide notification when resource is re-saved
1. Create a new resource.
2. Populate values for the resource.
3. Click Save & Publish and proceed with saving and publishing.
4. Make a change to the resource.
5. Click Save & Publish again.

Expected result:
User is notified that resource is being saved again.

justinlittman:
closes #1032


aaron-collier:
@justinlittman when testing this locally, the message shows up for about 3 seconds, then clears. Is that expected?

justinlittman:
Yes. I should have made that clear in PR.

PR #1054, opened 2019-07-25
Also validates when adding / removing properties.

### LINKED ISSUE #1036: Remove validation error messages when User performs necessary action
When a user performs an action that fixes a validation error, we should clear UI error artifacts. Below is a more complex example, but ideally this should be done for all components.

## Example
Using the BIBFRAME Instance fixture, after loading and clicking the **Save and Publish** button the expected validation errors appear:
Screen Shot 2019-07-23 at 10 40 11 AM

The Item Information **Barcode** is displaying the Required label as expect:
Screen Shot 2019-07-23 at 10 41 58 AM

Clicking the **Add** button for the Barcode adds and expands the Barcode but the disabled field is highlighted with the Required error message even though by clicking and adding the Barcode resource template I added a value. 
Screen Shot 2019-07-23 at 10 43 04 AM

This isn't a blocker because clicking on the **Save and Publish** again now clears out the barcode error from the list of errors:
Screen Shot 2019-07-23 at 10 45 18 AM

justinlittman:
closes #1036


justinlittman:
Once you are in "validation mode" (tried to Save & Publish but validation failed), validation notifications should be removed as the errors are corrected.

peetucket:
Looks like a rebase is needed.

PR #1057, opened 2019-07-25
Allows literal in InputLookupQA.

### LINKED ISSUE #1049: allow user to enter a literal in a lookup field
see previous issue https://github.com/LD4P/sinopia_editor/issues/866, the part about entering a URI was already solved by https://github.com/LD4P/sinopia_editor/pull/995

but catalogers still have use cases for needing to enter a literal
--We need to be able to enter a literal for cases where there is no known URI, as most institutions don't have the bandwidth to do authority work for everything that isn't in a controlled vocabulary unless there is a conflict for the string used to represent it.
--there are cases where we have both a vocabulary that is available as linked data, but also another vocabulary we also want to use that we can currently only give as literals

per dev team discussions, need to check user's input to see if it's a URI or literal and serialize accordingly

justinlittman:
closes #1049

### DIFF
--- a/__tests__/components/editor/property/InputLookupQA.test.js
+++ b/__tests__/components/editor/property/InputLookupQA.test.js
@@ -95,7 +95,7 @@ const validNewURIResults = [{
   label: 'http://id.loc.gov/authorities/subjects/123456789',
 }]
 
-const invalidNewURIResults = [{
+const validNewLiteralResults = [{
   customOption: true,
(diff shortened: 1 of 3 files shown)


peetucket:
LGTM to me too.  Do we need to be concerned with RDF generation?

justinlittman:
@peetucket I verified that RDF generation works correctly (as well as loading).

PR #1061, opened 2019-07-25
Patch 0.3.6 along with Release notes

peetucket:


michelleif:
@jermnelson can we change this process for future so that PO writes the NewsItem

PR #1077, opened 2019-07-29
Removed duplicate lookups

### LINKED ISSUE #1035: Sinopia Editor pop-up message with "base material" QA lookup
See stage.sinopia.io "harvard:rt:test:single_lookup"

This template has a simple property with QA lookup for lc list "base material"  when opening rt the browser generates an infinite popup message:  "There are multiple configured list lookups for http://qa_test1"

Looking at the lookup config file: https://github.com/LD4P/sinopia_editor/blob/31bb926d80ece456b612e4465ea3befb830d9f7c/static/lookupConfig.json#L783  
it appears as thought 2 lists "base material" and "applied material" point to the same source list uri: https://id.loc.gov/vocabulary/mmaterial   Could this be the problem?

I also notice in the profile editor, when selecting the "value uri" from the QA list that when one tries to select "applied material" it automatically switches to "base material"

jcoyne:
Lookups must have a unique URI. Fixes #1035

### DIFF
--- a/static/lookupConfig.json
+++ b/static/lookupConfig.json
@@ -846,18 +846,13 @@
     "component": "list"
   },
   {
-    "label": "applied material",
+    "label": "material",
     "uri": "https://id.loc.gov/vocabulary/mmaterial",
     "component": "list"
   },
   {
     "label": "aspect ratio",
     "uri": "https://id.loc.gov/vocabulary/maspect",
-    "component": "list"
-  },
-  {
-    "label": "base material",
-    "uri": "https://id.loc.gov/vocabulary/mmaterial",
     "component": "list"
   },
   {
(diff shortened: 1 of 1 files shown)


static/lookupConfig.json
```diff
@@ -930,11 +925,6 @@
     "uri": "https://id.loc.gov/vocabulary/msoundcontent",
     "component": "list"
   },
-  {
-    "label": "special playback characteristics",
-    "uri": "https://id.loc.gov/vocabulary/mplayback",
-    "component": "list"
-  },
```
ndushay: this isn't a duplicate -- is it unused??

jcoyne: It's a duplicate of https://github.com/LD4P/sinopia_editor/blob/dc995f042c1452872bbbeb077b2546826415d0a5/static/lookupConfig.json#L900

jcoyne: See the issue for context: https://github.com/LD4P/sinopia_editor/issues/1035#issuecomment-514347925

ndushay: gotcha - thanks.

Issue #1060, opened 2019-07-25
Update UI for Add/Expand


astridu:
![image](https://user-images.githubusercontent.com/42216982/61908595-e3d9f780-aee4-11e9-848b-8030b3f4a542.png)

![image](https://user-images.githubusercontent.com/42216982/61908613-ed635f80-aee4-11e9-9930-7df46b4e93de.png)

astridu:
1. Change "Add" button next to BIBFRAME Work to "Add another BIBFRAME Work" (or whatever the item name is) and place it next to the title. 
2. For items that have not been added yet:
• make them a gray outlined button with white fill 
• Add a "+ Add" at the beginning (hex color: 4A4A4A)
• Make the item name bold in black
3. Once item has been added, it gets an expand/contract icon (fontawesome arrow that points to the right when item is not expanded, and down arrow when item is expanded)
4. For items with required fields, open all required fields when user expands the item.
• Required fields with a default value have the expand/contract icon.
• Required fields without a default value use the + Add button style with the field exposed so user can add a value. Once value is added, the item uses the expand/contract icon to indicate a value has been added.
5. For items like "Had BIBFRAME instance" where user can enter multiple instances, put the "+ Add [instance.name]" button under the title.

PR #1085, opened 2019-07-29
Aligned InputLiteral and InputLookup.

justinlittman:
refs #1060

### DIFF
--- a/src/components/editor/property/InputLiteral.jsx
+++ b/src/components/editor/property/InputLiteral.jsx
@@ -96,16 +96,22 @@ export class InputLiteral extends Component {
       const itemId = obj.id || shortid.generate()
 
       return <div id="userInput" key = {itemId} >
-        {obj.content}
-        <button
-          id="deleteItem"
-          type="button"
-          onClick={this.handleDeleteClick}
-          key={`delete${obj.id}`}
-          data-item={itemId}
-          data-label={this.props.formData.uri}
-        >X
-        </button>
+        <div
+          className="rbt-token rbt-token-removeable">
+          {obj.content}
+          <button
+            id={`delete${obj.id}`}
+            type="button"
+            onClick={this.handleDeleteClick}
+            key={`delete${obj.id}`}
+            data-item={obj.id}
+            data-label={this.props.formData.uri}
+            className="close rbt-close rbt-token-remove-button">
+            <span
+                aria-hidden="true"
+                data-item={obj.id}>×</span>
+          </button>
+        </div>
         <button
           id="editItem"
           type="button"
@@ -113,7 +119,8 @@ export class InputLiteral extends Component {
           key={`edit${obj.id}`}
           data-item={itemId}
           data-label={this.props.formData.uri}
-        >Edit
+          className="btn btn-sm btn-literal btn-default">
+          Edit
         </button>
         <LanguageButton id={obj.id} reduxPath={this.props.reduxPath}/>
       </div>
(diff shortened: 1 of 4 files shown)


src/components/editor/property/InputLiteral.jsx
```diff
@@ -96,24 +96,31 @@ export class InputLiteral extends Component {
...
-          key={`delete${obj.id}`}
-          data-item={itemId}
-          data-label={this.props.formData.uri}
-        >X
-        </button>
+        <div
+          className="rbt-token rbt-token-removeable">
+          {obj.content}
+          <button
+            id={`delete${obj.id}`}
```
jcoyne: 👏

src/components/editor/property/InputLiteral.jsx
```diff
@@ -96,24 +96,31 @@ export class InputLiteral extends Component {
...
+            type="button"
+            onClick={this.handleDeleteClick}
+            key={`delete${obj.id}`}
+            data-item={obj.id}
+            data-label={this.props.formData.uri}
+            className="close rbt-close rbt-token-remove-button">
+            <span
+                aria-hidden="true"
+                data-item={obj.id}>×</span>
```
jcoyne: is this needed here too?

src/components/editor/property/InputLiteral.jsx
```diff
@@ -96,24 +96,31 @@ export class InputLiteral extends Component {
...
-        >X
-        </button>
+        <div
+          className="rbt-token rbt-token-removeable">
+          {obj.content}
+          <button
+            id={`delete${obj.id}`}
+            type="button"
+            onClick={this.handleDeleteClick}
+            key={`delete${obj.id}`}
```
jcoyne: is `obj.id` enough of a key here?

src/components/editor/property/LanguageButton.jsx
```diff
@@ -48,7 +48,8 @@ const LanguageButton = (props) => {
       <Button
         id="language"
         bsSize="small"
-        onClick = { () => setShow(true) }>
+        onClick = { () => setShow(true) }
+        className="btn btn-literal btn-default">
```
jcoyne: Button should not have a `btn-default` className.  Instead use the `bsStyle="default"`

src/components/editor/property/InputLiteral.jsx
justinlittman: Yes, because depending on where you click, you either get the span or the button passed into the event.

src/components/editor/property/InputLiteral.jsx
justinlittman: Yes-ish. The key for the delete button is prefixed with _delete_; the key for the edit button is prefixed with _edit_. So it could go either way, but this is consistent with the edit button.

src/components/editor/property/LanguageButton.jsx
justinlittman: Fixed @jcoyne

src/components/editor/property/InputLiteral.jsx
jcoyne: Why did you move away from `itemId`?

justinlittman: Clearer. `const itemId = obj.id || shortid.generate()`. If, for some reason, obj.id were missing, then a generated shortid would be meaningless for deleting. So what you really need is obj.id.

PR #1088, opened 2019-07-29
Only render property panel body when it has been added.

justinlittman:
refs #1060


__tests__/components/editor/property/PropertyPanel.test.js
```diff
@@ -56,6 +59,9 @@ describe('<PropertyPanel />', () => {
     it('does not render Add button', () => {
       expect(wrapper.exists('button.btn-add')).toBeFalsy()
     })
+    it('renderd panel body', () => {
```
jcoyne: typo

justinlittman: Fixed.

PR #1096, opened 2019-07-30
Remove language label from the resource state

jcoyne:
Language is now an entity, we only use its identifier

### DIFF
--- a/src/Utilities.js
+++ b/src/Utilities.js
@@ -7,6 +7,7 @@ import _ from 'lodash'
 import shortid from 'shortid'
 import CryptoJS from 'crypto-js'
 
+export const defaultLanguageId = 'en'
 
 export const isResourceWithValueTemplateRef = property => property?.type === 'resource'
     && property?.valueConstraint?.valueTemplateRefs?.length > 0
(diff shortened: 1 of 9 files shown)


src/Utilities.js
```diff
@@ -7,6 +7,7 @@ import _ from 'lodash'
 import shortid from 'shortid'
 import CryptoJS from 'crypto-js'
 
+export const defaultLangTemplate = 'en'
```
ndushay: this var name -- it's not a Template, is it?

src/components/editor/property/InputLiteral.jsx
```diff
@@ -42,7 +42,7 @@ export class InputLiteral extends Component {
     userInputArray.push({
       content: currentcontent,
       id: shortid.generate(),
-      lang: defaultLangTemplate(),
+      lang: defaultLangTemplate,
```
ndushay: ditto on var name?

src/components/editor/property/LanguageButton.jsx
```diff
@@ -70,7 +70,7 @@ const mapStateToProps = (state, ourProps) => {
   const node = findNode(state.selectorReducer, ourProps.reduxPath)
   const item = node.items.find(item => item.id === ourProps.id)
   return {
-    language: languageLabel(state, item.lang.id),
+    language: languageLabel(state, item.lang),
```
ndushay: is this function name still appropriate?

jcoyne: yes, this is looking up the language label from the entity store.

jcoyne:
@ndushay changes made.

PR #1098, opened 2019-07-30
Make the layout responsive.

jcoyne:
Use a balanced 2-column layout on large screens and a single column layout on small screens.
Avoid having to precalculate left or right

Before:
Screen Shot 2019-07-30 at 12 24 25 PM

After:
Screen Shot 2019-07-30 at 12 26 52 PM


ndushay:
Can we see a before and after picture to show how this changes display (or doesn't)

jcoyne:
@ndushay I've added that.

jcoyne:
No, it was just bothering me that we were calculating how to render pieces on the page, when the browser engines are very good at doing that for us.

PR #1108, opened 2019-07-31
Convert items to a hash instead of an array

### LINKED ISSUE #886: InputLang Component uses Redux state
Refactor `InputLang` Component to use more of the Redux state (currently has placeholders for `mapStateToProps` and `mapDispatchToProps`.

jcoyne:
so that the items can be addressed by redux path

Fixes #886

### DIFF
--- a/src/components/editor/property/InputListLOC.jsx
+++ b/src/components/editor/property/InputListLOC.jsx
@@ -145,13 +145,16 @@ const mapStateToProps = (state, ownProps) => {
   const errors = findErrors(state.selectorReducer, ownProps.reduxPath)
 
   // Make sure that every item has a label
+  // TODO: Move this to the ResourceStateBuilder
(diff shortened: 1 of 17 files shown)


src/components/editor/property/InputListLOC.jsx
```diff
@@ -145,13 +145,16 @@ const mapStateToProps = (state, ownProps) => {
   const errors = findErrors(state.selectorReducer, ownProps.reduxPath)
 
   // Make sure that every item has a label
+  // TODO: Move this to the ResourceStateBuilder
```
jermnelson: Could you create a ticket for moving this to the ResourceStateBuilder and reference the issue here?

jcoyne: https://github.com/LD4P/sinopia_editor/issues/1117

PR #1110, opened 2019-07-31
Make InputLang a functional component

jcoyne:
Please merge #1108 first as this would otherwise result in a complicated rebase.

### DIFF
--- a/__tests__/components/editor/property/InputLang.test.js
+++ b/__tests__/components/editor/property/InputLang.test.js
@@ -29,33 +29,12 @@ describe('<InputLang />', () => {
     )
   })
 
-  it('typeahead component should use selectHintOnEnter', () => {
-    expect(wrapper.find('#langComponent').props().selectHintOnEnter).toBeTruthy()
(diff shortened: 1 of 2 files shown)


__tests__/components/editor/property/InputLang.test.js
```diff
@@ -29,33 +29,12 @@ describe('<InputLang />', () => {
...
-    const opts = { id: 'URI', label: 'LABEL', uri: 'URI' }
-
-    wrapper.instance().opts = opts
-    const event = (wrap) => {
-      global.fetch = jest.fn().mockImplementation(async () => await { ok: true, resp: wrapper.instance().opts })
-      wrap.setState({ options: [wrapper.instance().opts] })
-      wrap.setState({ selected: [wrapper.instance().opts] })
-    }
-
-    wrapper.find('#langComponent').simulate('focus', event(wrapper))
```
jcoyne: This test is not helpful.  here it calls the method defined on line 45, and then it checks that the method made changes to the state.   It is testing the test code itself, not the component under test.

Issue #1089, opened 2019-07-29
Add integration test for search


aaron-collier:
Related to #1082, and more specifically #1083, a recent change in the saving and resource template loading caused loading the resource from search to fail at in part because of https://github.com/LD4P/sinopia_editor/blob/master/src/actionCreators/resources.js#L45

jermnelson:
More generally, here are the specific things I think we would want a search integration test to accomplish:
- Search term with and without a wildcard
- List of Results
- Load a search result into the editor
- Edit and save back to Trellis

PR #1112, opened 2019-07-31
Add search paging

### LINKED ISSUE #1082: Search Results only shows the first 10
**Describe the bug**
When I search for a term that has more than 10 results only the first 10 are displayed. There is no way to load an item that is not in the top 10 results.

aaron-collier:
Fixes #1082 

Note: Integration testing of search will come in a follow up PR addressing #1089 

Screen Shot 2019-07-31 at 2 13 16 PM

UI Notes:
- Paging includes a first page, previous page, next page, and last page button.
- First, previous, next, and last buttons are disabled as necessary
- Current page is shown as active
- Total result count and current result span are displayed.

### DIFF
--- a/src/components/search/SearchResultsPaging.jsx
+++ b/src/components/search/SearchResultsPaging.jsx
@@ -5,11 +5,8 @@ import { connect } from 'react-redux'
 import PropTypes from 'prop-types'
 import Config from 'Config'
 import 'react-bootstrap-table-next/dist/react-bootstrap-table2.min.css'
-import { getCurrentUser } from 'authSelectors'
-import Button from 'react-bootstrap/lib/Button'
 import Pagination from 'react-bootstrap/lib/Pagination'
 import { showSearchResults } from 'actions/index'
-
 
 const SearchResultsPaging = (props) => {
   const [currentPage, setCurrentPage] = useState(1) // initialize currentPage to 1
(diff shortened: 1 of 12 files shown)


aaron-collier:
A follow up ticket will be written to add "ellipsis" functionality (see: https://react-bootstrap.github.io/components/pagination/#pagination-more) in order to address possible navigation issues with very large result sets, but that can currently be addressed by a higher `searchResultsPerPage` configuration option, so I'm inclined not to include it in this PR.

src/components/search/SearchResultsPaging.jsx
```diff
@@ -0,0 +1,117 @@
...
+    totalHits: json.hits.total,
+    results: json.hits.hits.map(row => ({
+      uri: row._id,
+      title: row._source.title,
+    })),
+  })
+
+
+  const search = (query, queryFrom) => {
+    const uri = `${Config.searchHost}${Config.searchPath}?q=title:${query}%20OR%20subtitle:${query}&from=${queryFrom}&size=${Config.searchResultsPerPage}`
```
jcoyne: It seems odd that we have this knowledge in here and in `components/search/Search`. Can this functionality be extracted?

src/components/search/SearchResultsPaging.jsx
```diff
@@ -0,0 +1,117 @@
...
+    fetch(uri)
```
jcoyne: This should be in an action or actionCreator, not a component.

aaron-collier: Yeah, likely. I was thinking about this as well, but wasn't sure an action/actionCreator was correct. I can extract it pretty easily I think.

PR #1116, opened 2019-07-31
[Refactor] extract Literal from InputLiteral

### LINKED ISSUE #1126: Fix formatting of InputURI
Such that it appears like InputLiteral.

Currently:
Screen Shot 2019-08-01 at 6 30 08 AM

jcoyne:
Fixes #1126

### DIFF
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -1,3 +1,4 @@
+
 // Copyright 2019 Stanford University see LICENSE for license
 
 // export const newResource = resourceTemplateId => ({
@@ -47,9 +48,9 @@ export const languageSelectOpened = () => ({
   type: 'LANGUAGE_SELECT_OPENED',
 })
 
-export const removeItem = (reduxPath, id) => ({
+export const removeItem = reduxPath => ({
   type: 'REMOVE_ITEM',
-  payload: { reduxPath, id },
+  payload: reduxPath,
 })
 
 export const removeResource = reduxPath => ({
--- a/src/reducers/inputs.js
+++ b/src/reducers/inputs.js
@@ -156,17 +156,10 @@ export const clearResourceURIMessage = (state) => {
 
 export const removeMyItem = (state, action) => {
   const newState = { ...state }
-  const reduxPath = action.payload.reduxPath
-  let level = 0
+  const reduxPath = action.payload
 
-  reduxPath.reduce((obj, key) => {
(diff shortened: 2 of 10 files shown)


ndushay:
Can we get a little context as to why this is being done, and done now?

jcoyne:
Motivation: was because I was having trouble understanding how to do my other ticket, because this one component had too many responsibilities

src/components/editor/property/Literal.jsx
```diff
@@ -0,0 +1,52 @@
...
+
+const Literal = (props) => {
+  const handleEditClick = () => {
+    props.handleEdit(props.item.content)
+    props.removeItem(props.reduxPath)
+  }
+
+  const itemId = props.item.id
+
+  return (<div id="userInput">
```
justinlittman: ids here and below should be unique.

src/reducers/inputs.js
```diff
@@ -156,17 +156,10 @@ export const clearResourceURIMessage = (state) => {
...
 
-  reduxPath.reduce((obj, key) => {
-    level++
-    if (level === reduxPath.length) {
-      delete obj[key].items[action.payload.id]
-    }
-
-    return obj[key]
-  }, newState)
+  const node = findNode(newState, reduxPath.slice(0, -1))
```
justinlittman: I prefer the original way of doing this. You know, because the original way was more confusing and complicated.

src/actions/index.js
```diff
@@ -47,9 +48,9 @@ export const languageSelectOpened = () => ({
   type: 'LANGUAGE_SELECT_OPENED',
 })
 
-export const removeItem = (reduxPath, id) => ({
+export const removeItem = reduxPath => ({
```
justinlittman: This will also affect InputURI, which uses removeItem.

src/components/editor/property/Literal.jsx
jcoyne: This ID was here before and the existing tests depend on it

jcoyne:
I think inputURI will have its own similar extraction. It uses different labels and doesn’t have language

justinlittman:
Still off ...
Screen Shot 2019-08-01 at 3 06 26 PM

jcoyne:
@justinlittman that's the current behavior in master.  There is no change in this PR.

PR #1119, opened 2019-07-31
Lookup context api

hudajkhan:
Based on discussions here (https://github.com/LD4P/sinopia_editor/pull/1023), we are breaking out work for integrating context into smaller portions, starting with this pull request.

The updates here enable a call to the Discogs authority to be made using QA and the ability to pass along a context parameter to QA if "subtype" in the resource template uses the value "context".  Two example templates, Discogs and ShareVDE, have been added to the fixtures directory and to the list of templates loaded when the application is configured to use the fixtures templates.  

Suggestions on updates to InputLookupQA.test.js would be welcome.  In this PR, the input's display functionality will not change but it will be able to make calls to discogs and pass along the context parameter (which are all part of the search function).  

Another option may be to always pass along "context:true" to the QA API.  This may have an effect on performance (as more info is being requested in that case) but it may be fine.


src/components/editor/property/InputLookupQA.jsx
```diff
@@ -156,6 +157,23 @@ class InputLookupQA extends Component {
...
           const language = lookupConfig.language
 
+          /*
+           *  There are two types of lookup: linked data and non-linked data. The API calls
+           *  for each type are different, so check the nonldLookup field in the lookup config.
+           *  If the field is not set, assume false.
+           */
+          const nonldLookup = lookupConfig.nonldLookup ? lookupConfig.nonldLookup : false
+
+          // default the API calls to their linked data values
```
jermnelson: Not sure if these specific QA parameters should be hardcoded in the component but fine with leaving them here for now.

hudajkhan: In case it's not clear here, these specific parameters are being read in from the lookup config to signal that the URLs for this specific vocabulary will be different from the regular /search/linked_data/vocab_name path (it will be /search/vocab_name instead).  
Also, to not require resource templates to always have the property for whether or not the lookup requires the linked data or non-linked data QA paths, the code uses the regular linked data path except if the resource template indicates otherwise. https://github.com/LD4P/sinopia_editor/blob/lookup_context_api/src/components/editor/property/InputLookupQA.jsx#L172

hudajkhan: I can break out the context parameter check into a separate function perhaps and then test if the function picks up the parameter from the resource template correctly.

jermnelson: I'm fine if you want to mark this resolved. When everything is merged, I think we could create an integration test to test for expectations arising from ShareVDE and Discogs sources.

hudajkhan:
Merged from master and added a test or two to check if context parameter is being picked up by a function that checks the resource template.

PR #1111, opened 2019-07-31
Fixes setting last save checksum so that it is called sync rather tha…

justinlittman:
…n async.

PR #1123, opened 2019-08-01
[Refactor] GroupChoiceModal uses thunk for saving to trellis

### LINKED ISSUE #1051: Move saving out of GroupChoiceModal into thunk
This doesn't belong in a component.

ndushay:
Closes #1051

This is a refactor;  no functionality change expected.

Please review carefully;  I'm pretty sure I got it right but it's my first thunk.

Also wondering if GroupChoiceModal component tests adequately assert the appropriate plumbing methods are called

### DIFF
--- a/__tests__/actionCreators/resources.test.js
+++ b/__tests__/actionCreators/resources.test.js
@@ -2,7 +2,7 @@
 
 import {
   update, existingResource, retrieveResource, newResource,
-  stubResourceProperties,
+  stubResourceProperties, publishResource,
 } from 'actionCreators/resources'
 /* eslint import/namespace: 'off' */
 import * as server from 'sinopiaServer'
@@ -15,19 +15,12 @@ import * as resourceTemplateThunks from 'actionCreators/resourceTemplates'
 const mockStore = configureMockStore([thunk])
 shortid.generate = jest.fn().mockReturnValue('abc123')
 
-describe('update', () => {
-  const state = {
-    selectorReducer: {
-      entities: {
-        resourceTemplates: {
-          'sinopia:profile:bf2:Place': {
-            resourceURI: 'http://id.loc.gov/ontologies/bibframe/place',
-          },
-        },
-      },
-      resource: {
+const state = {
+  selectorReducer: {
+    entities: {
+      resourceTemplates: {
         'sinopia:profile:bf2:Place': {
-          resourceURI: 'http://example.com/repository/stanford/12345',
+          resourceURI: 'http://id.loc.gov/ontologies/bibframe/place',
         },
       },
     },
@@ -38,6 +31,7 @@ describe('update', () => {
   }
   const store = mockStore(state)
 
+describe('update', () => {
   it('dispatches actions when started and finished', async () => {
     server.updateRDFResource = jest.fn().mockResolvedValue(true)
 
@@ -51,9 +45,6 @@ describe('update', () => {
 })
 
 describe('retrieveResource', () => {
-  const currentUser = {
-    getSession: jest.fn(),
-  }
   const uri = 'http://sinopia.io/repository/stanford/123'
   const received = `<> <http://www.w3.org/2000/01/rdf-schema#label> "splendid"@en .
 <> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://id.loc.gov/ontologies/bibframe/Note> .
@@ -112,6 +103,39 @@ describe('retrieveResource', () => {
       expect(actions[4]).toEqual({ type: 'SET_LAST_SAVE_CHECKSUM' })
       expect(actions[5]).toEqual({ type: 'SET_LAST_SAVE_CHECKSUM', payload: 'd7ee91e78c7065b55a0e7df016bd1622' })
     })
+  })
+})
+
+describe('publishResource', () => {
+  const group = 'myGroup'
+  const received = `<http://sinopia.io/repository/myGroup/myResource> <http://www.w3.org/2000/01/rdf-schema#label> "splendid"@en .
+<> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://id.loc.gov/ontologies/bibframe/Note> .
+<> <http://www.w3.org/ns/prov#wasGeneratedBy> "profile:bf2:Note" .`
+
+  it('dispatches actions for happy path', async () => {
+    server.publishRDFResource = jest.fn().mockResolvedValue({
+      response: {
+        headers: { location: 'http://sinopia.io/repository/myGroup/myResource' },
+        text: received,
+      },
+    })
+
+    const dispatch = jest.fn()
+    await publishResource(currentUser, group)(dispatch, getState)
+    expect(dispatch).toHaveBeenCalledTimes(4)
+    expect(dispatch).toBeCalledWith({ type: 'PUBLISH_STARTED' })
+    expect(dispatch).toBeCalledWith({ type: 'SET_BASE_URL', payload: 'http://sinopia.io/repository/myGroup/myResource' })
+    expect(dispatch).toBeCalledWith({ type: 'SHOW_RESOURCE_URI_MESSAGE', payload: 'http://sinopia.io/repository/myGroup/myResource' })
+    expect(dispatch).toBeCalledWith({ type: 'UPDATE_FINISHED', payload: '53ce99f9e4b1132733bae37801cd8000' })
+  })
+  it('dispatches actions for error path', async () => {
+    const dispatch = jest.fn()
+    server.publishRDFResource = jest.fn().mockRejectedValue(new Error('publish error'))
+
+    await publishResource(currentUser, group)(dispatch, getState)
+    expect(dispatch).toHaveBeenCalledTimes(2)
+    expect(dispatch).toBeCalledWith({ type: 'PUBLISH_STARTED' })
+    expect(dispatch).toBeCalledWith({ type: 'PUBLISH_ERROR', payload: 'Unable to save resource: Error: publish error' })
   })
 })
 
--- a/__tests__/actionCreators/resources.test.js
+++ b/__tests__/actionCreators/resources.test.js
@@ -24,12 +24,18 @@ const state = {
         },
       },
     },
-  }
+    resource: {
+      'sinopia:profile:bf2:Place': {
+        resourceURI: 'http://example.com/repository/stanford/12345',
+      },
+    },
+  },
+}
 
-  const currentUser = {
-    getSession: jest.fn(),
-  }
-  const store = mockStore(state)
+const currentUser = {
+  getSession: jest.fn(),
+}
+const store = mockStore(state)
 
 describe('update', () => {
   it('dispatches actions when started and finished', async () => {
@@ -113,6 +119,7 @@ describe('publishResource', () => {
 <> <http://www.w3.org/ns/prov#wasGeneratedBy> "profile:bf2:Note" .`
 
   it('dispatches actions for happy path', async () => {
+    const store = mockStore(state)
     server.publishRDFResource = jest.fn().mockResolvedValue({
       response: {
         headers: { location: 'http://sinopia.io/repository/myGroup/myResource' },
@@ -120,22 +127,23 @@ describe('publishResource', () => {
       },
     })
 
-    const dispatch = jest.fn()
-    await publishResource(currentUser, group)(dispatch, getState)
-    expect(dispatch).toHaveBeenCalledTimes(4)
-    expect(dispatch).toBeCalledWith({ type: 'PUBLISH_STARTED' })
-    expect(dispatch).toBeCalledWith({ type: 'SET_BASE_URL', payload: 'http://sinopia.io/repository/myGroup/myResource' })
-    expect(dispatch).toBeCalledWith({ type: 'SHOW_RESOURCE_URI_MESSAGE', payload: 'http://sinopia.io/repository/myGroup/myResource' })
-    expect(dispatch).toBeCalledWith({ type: 'UPDATE_FINISHED', payload: '53ce99f9e4b1132733bae37801cd8000' })
+    await store.dispatch(publishResource(currentUser, group))
+    const actions = store.getActions()
+    expect(actions.length).toEqual(4)
+    expect(actions[0]).toEqual({ type: 'PUBLISH_STARTED' })
+    expect(actions[1]).toEqual({ type: 'SET_BASE_URL', payload: 'http://sinopia.io/repository/myGroup/myResource' })
+    expect(actions[2]).toEqual({ type: 'SHOW_RESOURCE_URI_MESSAGE', payload: 'http://sinopia.io/repository/myGroup/myResource' })
+    expect(actions[3]).toEqual({ type: 'UPDATE_FINISHED', payload: 'cd2a00d75801b859a7d6228beefb6ace' })
   })
   it('dispatches actions for error path', async () => {
-    const dispatch = jest.fn()
+    const store = mockStore(state)
     server.publishRDFResource = jest.fn().mockRejectedValue(new Error('publish error'))
 
-    await publishResource(currentUser, group)(dispatch, getState)
-    expect(dispatch).toHaveBeenCalledTimes(2)
-    expect(dispatch).toBeCalledWith({ type: 'PUBLISH_STARTED' })
-    expect(dispatch).toBeCalledWith({ type: 'PUBLISH_ERROR', payload: 'Unable to save resource: Error: publish error' })
+    await store.dispatch(publishResource(currentUser, group))
+    const actions = store.getActions()
+    expect(actions.length).toEqual(2)
+    expect(actions[0]).toEqual({ type: 'PUBLISH_STARTED' })
+    expect(actions[1]).toEqual({ type: 'PUBLISH_ERROR', payload: 'Unable to save resource: Error: publish error' })
   })
 })
 
--- a/__tests__/components/editor/GroupChoiceModal.test.js
+++ b/__tests__/components/editor/GroupChoiceModal.test.js
@@ -10,24 +10,17 @@ import GroupChoiceModal from 'components/editor/GroupChoiceModal'
 import * as server from 'sinopiaServer'
 
 describe('<GroupChoiceModal />', () => {
-  server.publishRDFResource = jest.fn().mockResolvedValue({
-    response: {
-      headers: {
-        location: 'http://example.com/resource/123',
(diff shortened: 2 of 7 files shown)


src/components/editor/GroupChoiceModal.jsx
```diff
@@ -102,15 +93,8 @@ const mapDispatchToProps = dispatch => ({
...
-    dispatch(assignBaseURL(url))
-    dispatch(showResourceURIMessage(url))
-  },
-  saveStarted() {
-    dispatch(updateStarted())
-  },
-  saveFinished(rdf) {
-    dispatch(updateFinished(generateMD5(rdf)))
+  publishMyResource: (user, group) => {
+    dispatch(publishResource(user, group))
```
ndushay: is this needed, or could i just call publishResource() directly in saveAndClose() ?  It seems like a silly wrapper.

__tests__/actionCreators/resources.test.js
```diff
@@ -13,42 +13,38 @@ import shortid from 'shortid'
...
-    selectorReducer: {
-      entities: {
-        resourceTemplates: {
-          'sinopia:profile:bf2:Place': {
-            resourceURI: 'http://id.loc.gov/ontologies/bibframe/place',
-          },
-        },
-      },
-      resource: {
+const state = {
```
ndushay: refactoring these test vars out made the other tests shorter and DRYed it up

__tests__/components/editor/GroupChoiceModal.test.js
```diff
@@ -91,16 +84,27 @@ describe('<GroupChoiceModal />', () => {
...
+      },
+    })
+    it('attempts to save the RDF content with group choice when save is clicked and then closes the modals', () => {
+      const selectedGroup = 'cornell' // default is first choice, which is cornell
       wrapper.find('[bsStyle="primary"]').simulate('click')
-      expect(server.publishRDFResource).toHaveBeenCalledWith(currentUser, 'some triples', 'cornell')
-      expect(mockSaveStarted).toHaveBeenCalled()
+      expect(mockPublishMyResource).toHaveBeenCalledWith(currentUser, selectedGroup)
+      expect(mockCloseRdfPreview).toHaveBeenCalled()
+      expect(mockCloseGroupChooser).toHaveBeenCalled()
```
ndushay: this is where I wonder if this is sufficient testing

src/components/editor/GroupChoiceModal.jsx
```diff
@@ -28,18 +24,13 @@ const GroupChoiceModal = (props) => {
...
-      props.setBaseURL(result.response.headers.location)
-      // Need to regenerate RDF now that have baseURL
-      props.saveFinished(props.rdf())
-    }).catch((err) => {
-      alert('Unable to save resource')
-      console.error('unable to save resource')
-      console.error(err)
-    })
     props.closeRdfPreview()
     props.close()
```
ndushay: keeping these modal close actions here made sense to me, as they are pertinent to the logic of the GroupChoiceModal.  In fact, I originally closed them before showing the error message and an integration test failed.

src/components/editor/GroupChoiceModal.jsx
```diff
@@ -28,18 +24,13 @@ const GroupChoiceModal = (props) => {
   }
 
   const saveAndClose = () => {
-    props.saveStarted()
-    const request = publishRDFResource(props.currentUser, props.rdf(), selectedValue)
+    props.publishMyResource(props.currentUser, selectedValue)
+
+    if (props.error) {
+      alert(props.error)
+      console.error(props.error)
```
ndushay: I was trying to figure out if we do this sort of reporting back to the user when retrieving an existing resource, and it didn't appear to be true to me ... did I misunderstand the code?

__tests__/reducers/index.test.js
```diff
@@ -80,6 +80,25 @@ describe('createReducer', () => {
   })
 })
 
+describe('setPublishError', () => {
+  it('adds error to editor state', () => {
+    const newState = setPublishError(initialState.selectorReducer, {
+      type: 'RETRIEVE_ERROR',
```
jcoyne: Wouldn't this be `PUBLISH_ERROR`?

__tests__/reducers/index.test.js
```diff
@@ -80,6 +80,25 @@ describe('createReducer', () => {
...
+      payload: { resourceTemplateId: 'abc123' },
+    })
+
+    expect(newState.editor.serverError).toEqual('There was a problem publishing abc123.')
+  })
+  it('adds error with reason to editor state', () => {
+    const newState = setPublishError(initialState.selectorReducer, {
+      type: 'RETRIEVE_ERROR',
```
jcoyne: `PUBLISH_ERROR`?

__tests__/reducers/index.test.js
```diff
@@ -80,6 +80,25 @@ describe('createReducer', () => {
   })
 })
 
+describe('setPublishError', () => {
+  it('adds error to editor state', () => {
+    const newState = setPublishError(initialState.selectorReducer, {
+      type: 'RETRIEVE_ERROR',
+      payload: { resourceTemplateId: 'abc123' },
```
jcoyne: why would it have a `resourceTemplateId` here? Shouldn't the payload just be a string with the reason?

__tests__/reducers/index.test.js
```diff
@@ -80,6 +80,25 @@ describe('createReducer', () => {
...
+    })
+
+    expect(newState.editor.serverError).toEqual('There was a problem publishing abc123.')
+  })
+  it('adds error with reason to editor state', () => {
+    const newState = setPublishError(initialState.selectorReducer, {
+      type: 'RETRIEVE_ERROR',
+      payload: { resourceTemplateId: 'abc123', reason: 'Because it is broken.' },
```
jcoyne: `resourceTemplateId` ? (see above)

src/components/editor/GroupChoiceModal.jsx
```diff
@@ -7,13 +7,9 @@ import Modal from 'react-bootstrap/lib/Modal'
 import PropTypes from 'prop-types'
 import GraphBuilder from 'GraphBuilder'
 import Config from 'Config'
-import {
-  closeGroupChooser, showRdfPreview, assignBaseURL, showResourceURIMessage,
-  updateStarted, updateFinished,
-} from 'actions/index'
-import { publishRDFResource } from 'sinopiaServer'
```
jcoyne: 👏

src/components/editor/GroupChoiceModal.jsx
```diff
@@ -28,18 +24,13 @@ const GroupChoiceModal = (props) => {
   }
 
   const saveAndClose = () => {
-    props.saveStarted()
-    const request = publishRDFResource(props.currentUser, props.rdf(), selectedValue)
+    props.publishMyResource(props.currentUser, selectedValue)
+
+    if (props.error) {
+      alert(props.error)
```
jcoyne: `props.error` should be shown when rendering, not when doing `saveAndClose`, because otherwise it might not be show until the *next* time you call `saveAndClose`.

I think the standard way we've been showing errors is in an bootstrap styled alert div, not a browser alert.  Can you use that method for consistency?

src/reducers/index.js
```diff
@@ -57,6 +57,24 @@ export const removeResource = (state, action) => {
   return validate(newState)
 }
 
+// TODO: could combine setPublishError() and setRetrieveError() with arg for single word diff in error message
+//   for now, shameless green.  (Sandi Metz says hold your nose for 3 occurrences before refactoring)
+export const setPublishError = (state, action) => {
+  const resourceTemplateId = action.payload.resourceTemplateId
+  const reason = action.payload.reason
```
jcoyne: Isn't the payload itself the reason? See https://github.com/LD4P/sinopia_editor/pull/1123/files#diff-4e3037116dbc2bc69a473b7ace3d02abR94

src/reducers/index.js
```diff
@@ -57,6 +57,24 @@ export const removeResource = (state, action) => {
   return validate(newState)
 }
 
+// TODO: could combine setPublishError() and setRetrieveError() with arg for single word diff in error message
```
jermnelson: Could you create a ticket for this TODO and move some of detail from this comment to the ticket?

__tests__/reducers/index.test.js
ndushay: copy pasta - fixing.

__tests__/reducers/index.test.js
ndushay: I was a bit muzzy on this.  It can certainly just have the error.

ndushay: is there a better way to test these that avoids mocking errors passing?

src/components/editor/GroupChoiceModal.jsx
ndushay: thanks - will do.

src/reducers/index.js
ndushay: So the action is this, as your link shows:

```js
export const publishError = reason => ({
  type: 'PUBLISH_ERROR',
  payload: reason,
})
```

But that is some notation foo for 

```
export const publishError = reason => ({
  type: 'PUBLISH_ERROR',
  payload: { reason: reason }
})
```

Which I can show empirically by testing diff references in reducer:

```js
const reason = action.payload
```

results:

image

while the following passes:

```js
const reason = action.payload?.reason
```

src/reducers/index.js
ndushay: I'm vastly shortening the comment - is that your main objection to it?

I think making an issue is premature, but will create one if you think it's worthy.

ndushay:
@jcoyne @jermnelson the changes I just pushed up should address both of your concerns.

ndushay:
Note that if PR #1111 is merged first, there should be one more update.

__tests__/reducers/index.test.js
```diff
@@ -80,8 +80,26 @@ describe('createReducer', () => {
...
+    const newState = setPublishError(initialState.selectorReducer, {
+      type: 'PUBLISH_ERROR',
+    })
+
+    expect(newState.editor.serverError).toEqual('There was a problem saving the resource.')
+  })
+  it('adds error with reason to editor state', () => {
+    const newState = setPublishError(initialState.selectorReducer, {
+      type: 'PUBLISH_ERROR',
+      payload: { reason: 'publishing error msg' },
```
jcoyne: I believe the payload here would be a string.

__tests__/reducers/index.test.js
```diff
@@ -80,8 +80,26 @@ describe('createReducer', () => {
...
+    })
+
+    expect(newState.editor.serverError).toEqual('There was a problem saving the resource: publishing error msg')
+  })
+})
+
 describe('setRetrieveError', () => {
-  it('adds error to editor state', () => {
+  it('adds error to editor state wen no reason given', () => {
```
jcoyne: typo

src/components/editor/GroupChoiceModal.jsx
```diff
@@ -28,22 +24,16 @@ const GroupChoiceModal = (props) => {
...
-      console.error(err)
-    })
+    props.publishMyResource(props.currentUser, selectedValue)
     props.closeRdfPreview()
     props.close()
   }
 
+  if (props.error) {
+    // the error will be displayed via Editor -> ErrorMessages components
+    return (<div />)
```
jcoyne: I don't think that saveAndClose is expected to return any values ever.

src/components/editor/GroupChoiceModal.jsx
```diff
@@ -79,20 +69,20 @@ const GroupChoiceModal = (props) => {
...
-  setBaseURL: PropTypes.func,
   closeRdfPreview: PropTypes.func,
   choose: PropTypes.func,
   show: PropTypes.bool,
   rdf: PropTypes.func,
   currentUser: PropTypes.object,
-  saveStarted: PropTypes.func,
-  saveFinished: PropTypes.func,
+  publishMyResource: PropTypes.func,
+  error: PropTypes.string,
```
jcoyne: why do we need `error` as a prop?

src/components/editor/GroupChoiceModal.jsx
```diff
@@ -79,20 +69,20 @@ const GroupChoiceModal = (props) => {
...
 }
 
 const mapStateToProps = state => ({
   show: state.selectorReducer.editor.groupChoice.show,
   rdf: () => new GraphBuilder(state.selectorReducer).graph.toCanonical(),
   currentUser: getCurrentUser(state),
+  error: state.selectorReducer.editor.serverError,
```
jcoyne: Can this be discarded?

src/components/editor/GroupChoiceModal.jsx
jcoyne: Needed because you need `dispatch`.  That is not available in saveAndClose.

src/reducers/index.js
```diff
@@ -57,9 +57,27 @@ export const removeResource = (state, action) => {
   return validate(newState)
 }
 
+// shameless green -- nearly duplicates setRetrieveError()
```
jcoyne: I don't know what this means.

src/reducers/index.js
```diff
@@ -57,9 +57,27 @@ export const removeResource = (state, action) => {
...
+export const setPublishError = (state, action) => {
+  // Use safe navigation for dynamic properties: https://github.com/tc39/proposal-optional-chaining#syntax
+  const reason = action.payload?.reason
```
jcoyne: I don't think you need safe nav hear because `action.payload` is guaranteed to be present.

src/reducers/index.js
```diff
@@ -57,9 +57,27 @@ export const removeResource = (state, action) => {
...
+
+  newState.editor.serverError = serverError
+  return newState
+}
+
 export const setRetrieveError = (state, action) => {
-  const resourceTemplateId = action.payload.resourceTemplateId
-  const reason = action.payload.reason
+  // Use safe navigation for dynamic properties: https://github.com/tc39/proposal-optional-chaining#syntax
+  const resourceTemplateId = action.payload?.resourceTemplateId
```
jcoyne: Again, you don't need this.  Remember that this expands to a lot of low level code.

src/reducers/index.js
```diff
@@ -57,9 +57,27 @@ export const removeResource = (state, action) => {
...
+  const reason = action.payload?.reason
```
jcoyne: Same here.

src/components/editor/GroupChoiceModal.jsx
ndushay: it doesn't, now.  Am i missing something?

src/components/editor/GroupChoiceModal.jsx
ndushay: oh - that went away.  Good 👀 .  thanks

__tests__/reducers/index.test.js
ndushay: Hrmm. I changed it and got this:

image

I tried to make it a string before (see comment train above).  What am I missing?

src/components/editor/GroupChoiceModal.jsx
ndushay: oui

src/reducers/index.js
ndushay: hrmm.  Jeremy didn't like the longer comment;  this is too terse for you.  Lemme try a slight revision.

src/reducers/index.js
ndushay: It's because of this:
image

ndushay: I will try adding empty payload to test.  :-)

src/actionCreators/resources.js
```diff
@@ -47,6 +48,24 @@ export const retrieveResource = (currentUser, uri) => (dispatch) => {
...
+  dispatch(publishStarted()) // clears possible residual server error
+
+  const rdf = new GraphBuilder(getState().selectorReducer).graph.toCanonical()
+
+  return publishRDFResource(currentUser, rdf, group).then((result) => {
+    const resourceUrl = result.response.headers.location
+    dispatch(assignBaseURL(resourceUrl))
+    dispatch(showResourceURIMessage(resourceUrl))
+    // Need to regenerate RDF now that we have baseURL
+    const updatedRdf = new GraphBuilder(getState().selectorReducer).graph.toCanonical()
```
justinlittman: There is going to be a race condition here between line 59 and line 62. If assignBaseURL hasn't completed in time, it will produce an incorrect MD5. I'd recommend putting the base URL in a local copy of state and pass that to the GraphBuilder.

ndushay: @justinlittman   I hear you ... but this is straight copied from what was in code -- this is a refactor PR.  Can we do the fix separately, or maybe merge this and put fix in your PR #1111?  🤞

src/components/editor/GroupChoiceModal.jsx
ndushay: I've removed these useless lines, which weren't actually in saveAndClose() in any case.

src/reducers/index.js
ndushay: this is from the Sandi Metz class -- implementing something that works, but that isn't necessarily optimal code, that passes tests ... is `shameless green`.  This phrase appears in various code base comments.

The bit after explains how this may ultimately be suboptimal code.

src/reducers/index.js
ndushay: You are correct, sir!

src/actionCreators/resources.js
ndushay: or maybe you can push a commit up to this branch if it's super easy for you?

ndushay:
@jcoyne @justinlittman more changes ... are we done yet?  :-)

src/actionCreators/resources.js
```diff
@@ -48,6 +49,30 @@ export const retrieveResource = (currentUser, uri) => (dispatch) => {
...
+export const publishResource = (currentUser, group) => (dispatch, getState) => {
+  dispatch(publishStarted()) // clears possible residual server error
+
+  // Make a copy of state to prevent changes that will affect the publish.
+  const state = _.cloneDeep(getState())
+  const rdf = new GraphBuilder(state.selectorReducer).graph.toCanonical()
+
+  return publishRDFResource(currentUser, rdf, group).then((result) => {
+    const resourceUrl = result.response.headers.location
+    dispatch(assignBaseURL(resourceUrl))
```
ndushay: is this `assignBaseURL` dispatch still appropriate given lines 65-66

PR #1129, opened 2019-08-01
Prop component refactor

jcoyne:
I've split out PropertyTemplate related functionality from `src/Utilities.js` into `src/utilities/propertyTemplates.js`.

I've also extracted `getTagNameForPropertyTemplate` into that new `src/utilities/propertyTemplates.js` and out of PropertyComponent.


__tests__/components/editor/ResourceTemplateForm.test.js
```diff
@@ -76,10 +76,6 @@ describe('<ResourceTemplateForm /> after fetching data from sinopia server', ()
 
       const wrapper = shallow(<ResourceTemplateForm.WrappedComponent {...rtProps} />)
       expect(wrapper.find('div.ResourceTemplateForm PropertyComponent').length).toEqual(1)
-      const inputType = wrapper.find('PropertyComponent').dive()
-
-      expect(inputType.find('Connect(InputLookupQA)').length).toEqual(1)
-      expect(inputType.find('Connect(InputListLOC)').length).toEqual(0)
```
ndushay: getting rid of these lines because ???

I feel like we are really shy on tests that check component interaction ... but I dunno if this is a great way to go.

__tests__/components/App.test.js
```diff
@@ -1,6 +1,7 @@
 // Copyright 2019 Stanford University see LICENSE for license
 
 import 'jsdom-global/register'
+import 'isomorphic-fetch'
```
ndushay: why do we need this import now?

jcoyne: I don't really know why.  If we omit it, we get "fetch is not defined".

__tests__/components/editor/ResourceTemplateForm.test.js
jcoyne: Because the ResourceTemplateForm test should be about the resource template form.  The PropertyComponent test is handling these cases as are *all* of the integration tests.  Furthermore, since the loading is dynamic now, enzyme won't be able to see these components as they're actually nested in a Suspense/lazy load.

PR #1133, opened 2019-08-01
Make InputLiteral a functional component

jcoyne:


__tests__/components/editor/property/InputLiteral.test.js
```diff
@@ -40,45 +46,31 @@ describe('<InputLiteral />', () => {
...
   it('contains required="false" attribute on input tag when mandatory is false', () => {
-    wrapper.instance().props.propertyTemplate.mandatory = 'false'
-    wrapper.instance().forceUpdate()
+    const propertyTemplate = { propertyTemplate: { ...plProps.propertyTemplate, mandatory: 'false' } }
+    wrapper.setProps({ ...plProps, ...propertyTemplate })
     expect(wrapper.find('input').prop('required')).toBeFalsy()
   })
 })
 
-describe('checkMandatoryRepeatable', () => {
```
ndushay: why don't we want / need this test any more?  I suspect we will see all the variants of mandatory true/false and repeatable true/false ...

Is this duplicated elsewhere?

jcoyne: There is no such thing as "checkMandatoryRepeatable." I think this is a relic from another era. We are already testing `mandatory` on 48-59

ndushay: I was interpreting the test as "check when both mandatory AND repeatable are true" but whatever.

PR #1137, opened 2019-08-02
Fixes typehead bug when hitting backspace key.

### LINKED ISSUE #581: Firefox only: hitting delete button in a QA lookup field to remove a value has same effect as browser back button
In Editor, in a lookup field that uses QA, when the user hits the "delete" key with the intention of backspacing to remove one of the previously selected values (or to remove the last typed character in a type ahead), it's as if you hit the browser back button. You are taken off of the cataloging screen and back to the list of Resource Templates.
[delete_key.swf.zip](https://github.com/LD4P/sinopia_editor/files/3234661/delete_key.swf.zip)

justinlittman:
closes #581

So this isn't perfect (you have to hit backspace twice to delete a value in a typeahead box), but better than the reported bug.

### DIFF
--- a/src/components/editor/property/InputListLOC.jsx
+++ b/src/components/editor/property/InputListLOC.jsx
@@ -25,6 +25,17 @@ class InputListLOC extends Component {
     }
   }
 
+  // From https://github.com/ericgio/react-bootstrap-typeahead/issues/389
+  onKeyDown = (e) => {
+    // 8 = backspace
+    if (e.keyCode === 8
+        && e.target.value === '') {
+      // Don't trigger a "back" in the browser on backspace
+      e.returnValue = false
+      e.preventDefault()
+    }
+  }
+
   selectionChanged(items) {
     const payload = {
       id: this.props.propertyTemplate.propertyURI,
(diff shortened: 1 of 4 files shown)


src/components/editor/property/InputListLOC.jsx
```diff
@@ -25,6 +25,17 @@ class InputListLOC extends Component {
     }
   }
 
+  // From https://github.com/ericgio/react-bootstrap-typeahead/issues/389
+  onKeyDown = (e) => {
```
jcoyne: Can this code be extracted and shared among all 3 components?

justinlittman: It could, but (1) these components need to be refactored in general to remove redundancy and (2) we don't share component action functions anywhere else, so this would be a one-off.

If you feel strongly, I will extract but otherwise my gut is to leave as is.

ndushay: I'm fine with it as is;  Sandi Metz told us:

- repeat code once -- hold your nose and do it
- repeat code twice - still hold your nose and do it, esp. if code has low churn

PR #1150, opened 2019-08-02
Components do not need to know about the internals of state

jcoyne:


jcoyne:
@ndushay the idea is that a component is aware of `state` only, but not it's internals (e.g. `selectorReducer`).  `findObjectAtPath` was needed because `findNode` was previously used for things other than finding input values in the state.

src/ResourceValidator.js
```diff
@@ -1,9 +1,10 @@
-import { getResourceTemplate, findNode } from 'selectors/resourceSelectors'
+import { getResourceTemplate } from 'selectors/resourceSelectors'
 import _ from 'lodash'
 
 export default class Validator {
-  constructor(selectorReducer) {
+  constructor(selectorReducer, findNode) {
```
justinlittman: Why pass findNode to Validator?

jcoyne: This refactor enabled me to change the implementation easily.  Now that the refactor is done, we could back this out if you care about that.

PR #1158, opened 2019-08-02
Display error for bad prop template

### LINKED ISSUE #1122: Display Error Message for edge case in PropertyComponent
**Describe the bug**

If a user has a resource template with a propertyTemplate like this:

```javascript
  {
            "mandatory": "false",
            "repeatable": "true",
            "type": "lookup",
            "resourceTemplates": [],
            "valueConstraint": {
              "valueTemplateRefs": [
                "PMO:SoundRecording:MOP:MediumOfPerformance"
              ],
              "useValuesFrom": [],
              "defaults": [],
              "valueDataType": {}
            },
            "propertyLabel": "PMO Medium of Performance",
            "propertyURI": "http://performedmusicontology.org/ontology/hasMediumOfPerformance"
          }
```
results in this error described in #1101. We should reflect the error up through the user interface that is return at the console.error here https://github.com/LD4P/sinopia_editor/blob/d0979e79f6162998cc95b25bfe0a85bd6702b0f3/src/components/editor/property/PropertyComponent.jsx#L37-L39

ndushay:
Fixes #1122.

# Before

note no display of error except in console ... (you can tell I clicked the Add button because it's darker blue ...)

1122 before

# After

1122 after

### DIFF
--- a/src/ResourceValidator.js
+++ b/src/ResourceValidator.js
@@ -35,16 +35,22 @@ export default class Validator {
   validateResource(reduxPath, labelPath) {
     const resourceTemplateId = reduxPath.slice(-1)[0]
     const resourceTemplate = getResourceTemplate({ selectorReducer: this.selectorReducer }, resourceTemplateId)
-    resourceTemplate.propertyTemplates.forEach((propertyTemplate) => {
-      const newReduxPath = [...reduxPath, propertyTemplate.propertyURI]
-      const newLabelPath = [...labelPath, resourceTemplate.resourceLabel, propertyTemplate.propertyLabel]
-      if (propertyTemplate.mandatory === 'true') {
-        this.validateMandatoryProperty(newReduxPath, newLabelPath)
-      }
-      if (!_.isEmpty(propertyTemplate.valueConstraint?.valueTemplateRefs)) {
-        this.validateNestedResourceProperty(newReduxPath, newLabelPath, propertyTemplate)
-      }
-    })
+    if (resourceTemplate) {
+      resourceTemplate.propertyTemplates.forEach((propertyTemplate) => {
+        const newReduxPath = [...reduxPath, propertyTemplate.propertyURI]
+        const newLabelPath = [...labelPath, resourceTemplate.resourceLabel, propertyTemplate.propertyLabel]
+        if (propertyTemplate.mandatory === 'true') {
+          this.validateMandatoryProperty(newReduxPath, newLabelPath)
+        }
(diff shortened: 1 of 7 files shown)


src/ResourceValidator.js
```diff
@@ -35,16 +35,22 @@ export default class Validator {
...
+          this.validateMandatoryProperty(newReduxPath, newLabelPath)
+        }
+        if (!_.isEmpty(propertyTemplate.valueConstraint?.valueTemplateRefs)) {
+          this.validateNestedResourceProperty(newReduxPath, newLabelPath, propertyTemplate)
+        }
+      })
+    } else {
+      // not sure this error is ever surfaced anywhere, but we do need to trap for undefined resourceTemplate
+      //  possibly will be surfaced when issue #1148 is addressed
+      this.addError(reduxPath, labelPath, `unable to retrieve ${resourceTemplateId} from local store`)
```
ndushay: needed to add this else block to avoid "resourceTemplate undefined" error, which wasn't even being trapped for.

src/components/editor/property/PropertyComponent.jsx
```diff
@@ -10,6 +10,19 @@ import { getTagNameForPropertyTemplate } from 'utilities/propertyTemplates'
 
 const PropertyComponent = (props) => {
   const tag = getTagNameForPropertyTemplate(props.propertyTemplate)
+  if (!tag) {
+    return (
+      <div className="row">
+        <div className="col-md-12" style={{ marginTop: '10px' }}>
+          <div className="alert alert-danger alert-dismissible">
+            <button className="close" data-dismiss="alert" aria-label="close">&times;</button>
```
jermnelson: Should the error be dismissible? Could clicking close here for alert give the false impression that the error is in someway addressed?

jermnelson: If you don't think it is a problem; just resolve this conversation and I'll approve so we can merge the PR

ndushay: mebbe we can address with follow-on issue/PR?  This is certainly an improvement over former behavior ...

ndushay: Also - we really would like to get folks using resourceTemplates that adhere to 0.2.0 schemas.  So not sure how often we'll hit this?

Another approach would be to have it display at the top of the screen, via ErrorMessages component.

This seemed potentially more useful but 🤷‍♀ .

I hope it's ok if I leave the ticket making to you, as I'm not really sure what the desired behavior would be.

PR #1159, opened 2019-08-02
Simplifies resource uri message, now that resource uri is available i…

justinlittman:
…n resource state.


justinlittman:
Before and after:

PR #1161, opened 2019-08-02
Extract QA search to a thunk

jcoyne:


src/actionCreators/qa.js
```diff
@@ -0,0 +1,75 @@
...
+
+const getSearchResults = (query, propertyTemplate) => Swagger({ spec: swaggerSpec })
+  .then((client) => {
+    const lookupConfigs = getLookupConfigItems(propertyTemplate)
+
+    // Create array of promises based on the lookup config array that is sent in
+    const lookupPromises = createLookupPromises(client, query, lookupConfigs)
+
+    /*
+     * If undefined, add info - note if error, error object returned in object
```
justinlittman: If defined?

jcoyne: I have no idea. This wasn't something I added, just moved from https://github.com/LD4P/sinopia_editor/pull/1161/files#diff-93b04186261f1a6fa9e8f243d588ef94L199

PR #1165, opened 2019-08-05
Changes label and uri fields expected from ES.

### LINKED ISSUE #1142: InputLookupSinopia / indexing mismatch
Steps to reproduce
In a full local dev environment, not using fixtures:
1. Create and publish a new Work Title.

In a full local dev environment, using fixtures:
1. Open the Testing Sinopia Lookup resource template.
2. Enter a search term in the lookup.

Expected result:
Lookup returns matches.

Actual result:
```
nvariant Violation: "One or more options does not have a valid label string. Check the `labelKey` prop to ensure that it matches the correct option key and provides a string for filtering and display."
```

The problem is that InputLookupSinopia is expecting the result to have certain fields:
```
const responseToOptions = json => json
    .hits.hits.map(row => ({ uri: row._source['@id'], label: row._source.label }))
```

However, the result does not have those fields:
```
{
	"took": 166,
	"timed_out": false,
	"_shards": {
		"total": 5,
		"successful": 5,
		"skipped": 0,
		"failed": 0
	},
	"hits": {
		"total": 1,
		"max_score": 0.2876821,
		"hits": [{
			"_index": "sinopia_resources",
			"_type": "sinopia",
			"_id": "repository/cornell/33e483c2-0fe7-49d8-be96-a615e996f00a",
			"_score": 0.2876821,
			"_source": {
				"document": {
					"@id": "",
					"@type": "http://id.loc.gov/ontologies/bibframe/Title",
					"mainTitle": {
						"@language": "en",
						"@value": "The Foo Chronicles"
					},
					"wasGeneratedBy": "sinopia:resourceTemplate:bf2:WorkTitle",
					"@context": {
						"mainTitle": {
							"@id": "http://id.loc.gov/ontologies/bibframe/mainTitle"
						},
						"wasGeneratedBy": {
							"@id": "http://www.w3.org/ns/prov#wasGeneratedBy"
						},
						"schema": "http://schema.org/",
						"owl": "http://www.w3.org/2002/07/owl#",
						"xsd": "http://www.w3.org/2001/XMLSchema#",
						"skos": "http://www.w3.org/2004/02/skos/core#",
						"memento": "http://mementoweb.org/ns#",
						"rdfs": "http://www.w3.org/2000/01/rdf-schema#",
						"acl": "http://www.w3.org/ns/auth/acl#",
						"geo": "http://www.w3.org/2003/01/geo/wgs84_pos#",
						"dc11": "http://purl.org/dc/elements/1.1/",
						"as": "https://www.w3.org/ns/activitystreams#",
						"rdf": "http://www.w3.org/1999/02/22-rdf-syntax-ns#",
						"ldp": "http://www.w3.org/ns/ldp#",
						"time": "http://www.w3.org/2006/time#",
						"prov": "http://www.w3.org/ns/prov#",
						"dc": "http://purl.org/dc/terms/"
					}
				},
				"title": ["The Foo Chronicles"],
				"title-suggest": ["the", "foo", "chronicles"],
				"subtitle": []
			}
		}]
	}
}
```

justinlittman:
closes #1142

### DIFF
--- a/__tests__/components/editor/property/InputLookupSinopia.test.js
+++ b/__tests__/components/editor/property/InputLookupSinopia.test.js
@@ -85,7 +85,7 @@ describe('<InputLookupSinopia />', () => {
           hits: [
             {
               _source: {
-                '@id': 'http://example.com/fakeThing',
+                uri: 'http://example.com/fakeThing',
                 label: 'This is a thing',
               },
             },
--- a/package-lock.json
+++ b/package-lock.json
@@ -11446,7 +11446,7 @@
       "dependencies": {
         "minimist": {
           "version": "0.0.10",
-          "resolved": "https://registry.npmjs.org/minimist/-/minimist-0.0.10.tgz",
+          "resolved": "http://registry.npmjs.org/minimist/-/minimist-0.0.10.tgz",
           "integrity": "sha1-3j+YVD2/lggr5IrRoMfNqDYwHc8=",
           "dev": true
         },
--- a/src/components/editor/property/InputLookupSinopia.jsx
+++ b/src/components/editor/property/InputLookupSinopia.jsx
@@ -29,7 +29,7 @@ const InputLookupSinopia = (props) => {
   const isRepeatable = booleanPropertyFromTemplate(props.propertyTemplate, 'repeatable', true)
 
   const responseToOptions = json => json
-    .hits.hits.map(row => ({ uri: row._source['@id'], label: row._source.label }))
+    .hits.hits.map(row => ({ uri: row._source.uri, label: row._source.label }))
 
   const search = (query) => {
     setLoading(true)
--- a/src/sinopiaServer.js
+++ b/src/sinopiaServer.js
@@ -183,13 +183,14 @@ export const loadRDFResource = async (currentUser, uri) => {
 export const getSearchResults = async (query, queryFrom = 0) => {
   const uri = `${Config.searchHost}${Config.searchPath}?q=title:${query}%20OR%20subtitle:${query}&from=${queryFrom}&size=${Config.searchResultsPerPage}`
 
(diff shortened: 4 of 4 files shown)


justinlittman:
This depends on https://github.com/LD4P/sinopia_indexing_pipeline/pull/31 to work correctly.

docker-compose.yml
```diff
@@ -1,20 +1,22 @@
...
-        USE_FIXTURES: 'false'
-        TRELLIS_BASE_URL: http://localhost:8080
-    environment:
-      INDEX_URL: http://search:9200
-    ports:
-      - 8000:8000
-    depends_on:
-      - platform
-      - pipeline
+  # editor:
```
jermnelson: Why is the editor image commented out? This is needed when deploying on AWS.

docker-compose.yml
```diff
@@ -1,20 +1,22 @@
...
+  #   environment:
+  #     INDEX_URL: http://search:9200
+  #   ports:
+  #     - 8000:8000
+  #   depends_on:
+  #     - platform
+  #     - pipeline
   pipeline:
-    image: ld4p/sinopia_indexing_pipeline:latest
+    # image: ld4p/sinopia_indexing_pipeline:latest
```
jermnelson: Is there a reason for the switch from a published image for the indexing pipeline? Using a relative path assumes that the indexing pipeline is available up a directory and that it is up to date. I think we want to keep using published Docker images.

justinlittman: This is was errant commit. Will fix.

docker-compose.yml
aaron-collier: How is docker-compose being used for AWS deployment?

jermnelson: > How is docker-compose being used for AWS deployment?

Likely I'm remembering an earlier iteration; this would break testing locally with the editor image forwarding traffic for search to the local containers.

Issue #556, opened 2019-05-23
signout interaction is flaky


jmartin-sul:
my experience is that if you sign into the editor, and then immediately sign out (without navigating anywhere or reloading the page), the following happens:
* signout calls out to AWS and invalidates your refresh token...
  * ...but for some reason, the component blinks on the sign-in form, and then goes back to displaying the logged in state.  the id token hangs around too, because local storage doesn't get cleared.
  * and since the refresh token's been invalidated, subsequent requests to AWS for signout fail (because you're already signed out from their perspective, and you need a valid refresh token to globally sign out your user)

and so, it just sort of looks like signout doesn't work sometimes, even when it's actually half-working.

in my experience, doing any sort of page refresh or navigation between the sign-in and the attempt to signout alleviates this problem.

### Workaround to make sure the user is globally signed out
* Clear all sinopia-related local storage in the browser.
* Sign into the editor application.
* Hard refresh the page (`Cmd+Shift+R` on Mac, `Ctrl+Shift+R` on Windows).
* Attempt to sign out.
  * If that works, all done.
  * If the same issue occurs, hit the signout button again.
    * If the error message says that the refresh token is no longer valid or is expired, you can go ahead and just clear Sinopia related local storage.
      * This error indicates that the token was invalidated properly on amazon's end, and that it will no longer be usable.  the bug here is that there browser state has become inconsistent, and still thinks that the now-invalid token will work.  simply clearing the dead token from the browser should get everything in sync.

michelleif:
see also #432

michelleif:
on Chrome, DEV, click sign out from Editor page, sometimes the click does nothing, sometimes it looks as if you've signed out, with blank username and password boxes appearing, but those immediately disappear and you are still signed in with the 
[sign_out_issue_chrome.swf.zip](https://github.com/LD4P/sinopia_editor/files/3230430/sign_out_issue_chrome.swf.zip)

peetucket:
some debugging suggests that the `props.currentUser` is not getting cleared by any of the signout() methods being called from https://github.com/LD4P/sinopia_editor/blob/master/src/components/LoginPanel.jsx#L53-L63, and since the `currentUserInfoPanel`  (https://github.com/LD4P/sinopia_editor/blob/master/src/components/LoginPanel.jsx#L70-L74) uses the value of that prop to decide how to render, its showing the wrong state

i am still trying to puzzle through where the logout methods are defined though (or how/where to alter the state of that prop)

[...]

PR #1167, opened 2019-08-05
Avoid 404 flash.

justinlittman:
This avoids a brief 404 that was occurring before a session was available.

It also changes routing so if a user is not logged in, all routes go to the home page.

### DIFF
--- a/__tests__/components/App.test.js
+++ b/__tests__/components/App.test.js
@@ -57,7 +57,7 @@ describe('#routes', () => {
     getState: () => ({ ...state }),
   })
 
-  describe('public routes', () => {
+  describe('when the user is not authenticated', () => {
     const unauthenticatedStoreFake = makeStoreFake({
       selectorReducer: {
         appVersion: {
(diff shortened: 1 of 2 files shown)


justinlittman:
@jermnelson Bringing this to your attention in case it changes something intentional and significant.

__tests__/components/App.test.js
```diff
@@ -88,50 +88,68 @@ describe('#routes', () => {
...
+    //   const component = renderRoutes('/templates')
+    //
+    //   expect(component.find(ImportResourceTemplate).length).toEqual(0)
+    // })
+    //
+    // it('renders a 404 for an unknown path', () => {
+    //   const component = renderRoutes('/blah')
+    //
+    //   expect(component.contains(<h1>404</h1>)).toEqual(true)
+    // })
```
ndushay: if this PR goes in, these commented out lines should go away? or??

__tests__/components/App.test.js
```diff
@@ -155,14 +173,65 @@ describe('#routes', () => {
       expect(component.contains(<h1>404</h1>)).toEqual(true)
     })
 
+    // it('redirects to /templates when "/editor" is visited without resource', () => {
+    //   const component = renderRoutes('/editor')
+    //
+    //   expect(component.find(ImportResourceTemplate).length).toEqual(1)
+    // })
```
ndushay: commented out lines left in by accident?

__tests__/components/App.test.js
justinlittman: Uggh, yes. Removed.

jermnelson:
I'm seeing different behavior when testing the functionality with this PR. The 404 splash screen is not showing up (good!) but now after I am logged in with my user name and log out, I get an alert message that my Access Token has been revoked:
Screen Shot 2019-08-05 at 5 00 23 PM

But I'm actually not logged out as I can still access all of the routes (`/search`, `/templates`, etc.) and the display doesn't reflect that I'm logged out either (see below)

Screen Shot 2019-08-05 at 5 02 44 PM

The 404 message flash may have had an effect to make sure the user is logged out using the Cognito SDK @jmartin-sul may have a better explanation about why the 404 is necessary.

__tests__/components/App.test.js
ndushay: what about this commented out section?

justinlittman:
The signout interaction is known flaky: https://github.com/LD4P/sinopia_editor/issues/556

The 404 was unrelated to this. It was introduced when I made some earlier changes to the routes: https://github.com/LD4P/sinopia_editor/commit/7cbe9f787900f74d046921ce2a98fefee10e13e3#diff-f880b57906a072cdffa62da2c954ddd4

jmartin-sul:
yup, sorry for chiming in late, but agreed with @justinlittman that the flakiness seen here is probably the known issue described in #556.

PR #1176, opened 2019-08-06
Reverts change to title field for searches

### LINKED ISSUE #1171: Search Results missing Title field
**Describe the bug**
When doing a wildcard search ("*") returns a list of results but the Title field is missing or not being displayed.

- What is the impact on your work? Needed for M4
- What environment are you in? Locally and tested in development
- What OS and browser are you using? Mac/Firefox & Chrome

Screen Shot 2019-08-05 at 6 10 29 PM

aaron-collier:
fixes #1171 

This change from #1165 doesn't appear to have changed the results from a title & subtitle search, so `title` is still required.

### DIFF
--- a/src/sinopiaServer.js
+++ b/src/sinopiaServer.js
@@ -190,7 +190,7 @@ export const getSearchResults = async (query, queryFrom = 0) => {
       totalHits: json.hits.total,
       results: json.hits.hits.map(row => ({
         uri: row._id,
-        title: row._source.label,
+        title: row._source.title,
       })),
     }))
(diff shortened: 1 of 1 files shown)


justinlittman:
So if `label` wasn't working, it was because of old pipeline code being deployed. I'd suggest that this be reverted. Displaying the `label` is better than the `title` because it optionally includes the `subtitle`.

aaron-collier:
@justinlittman confirmed. When I pulled an fresh image of pipeline and reindexed locally, I get `label` where `title` was in the past.

PR #1190, opened 2019-08-07
Added wildcard help text to InputLookupSinopia and Search.

### LINKED ISSUE #1186: add hint text to inputLookupSinopia component to give users tip on how to search
since inputLookupSinopia behaves differently than other Lookups (which themselves may have inconsistent behavior due to how implemented in QA), add a line of text to the UI for that component that says:

Use a * to wildcard your search.

@justinlittman lmk if we need a separate ticket to include same text on the Search tab

justinlittman:
closes #1186


jcoyne:
There appears to be a test failure here.

PR #1192, opened 2019-08-07
Give a more clear error when the lookup is misconfigured

### LINKED ISSUE #1187: property templates giving error messages, cannot figure out why
On DEV, resource template Harvard:rt:test:lookups, many of the fields are showing the error message "This property template should not be of type lookup"

Why are these causing an error? These property templates have what looks to be a valid urn in useValuesFrom.

Screen Shot 2019-08-06 at 10 11 52 PM

jcoyne:
Fixes #1187


michelleif:
@jcoyne thanks for this! so much easier to understand

Issue #1200, opened 2019-08-09
InputQALookup context typeahead should show label when values are objects


hudajkhan:
![image](https://user-images.githubusercontent.com/2677207/62786097-ec5b3200-ba8f-11e9-8756-570f45cf1479.png)

hudajkhan:
As an example, broader concepts in the LC genre form results (when additional context is requested) are passed back as objects.  For now, in the typeahead, the code should check if the value in the context is an object and then output the label value (possibly linking to the URI if possible).

hudajkhan:
Updated code in branch now has lookup context looking like this for genre form 
![image](https://user-images.githubusercontent.com/2677207/62794668-a9ef2080-baa2-11e9-8c84-b89c12f5088b.png)

jermnelson:
Fixed in PR #1201

PR #1201, opened 2019-08-09
#1200 handle objects in context

hudajkhan:
Updating context rendering component to handle objects that come back as values in context.
For example, broader concepts in genre form lookup come back as
values": [
    {
        "uri": "http://id.loc.gov/authorities/genreForms/gf2011026049",
        "id": "gf2011026049",
        "label": "Animated films"
    }
]

The code now looks for "label" in the value object and displays that.  If there is no label but there is a uri, it will display that instead, and if neither, then it will display a stringified version of the json object


hudajkhan:
Attached pictures in the issue #1200 should show a before and after view.

PR #1253, opened 2019-08-16
Adding resource templates that are being used for User Feedback testing

jermnelson:
The Sinopia user council is doing user feedback testing based on the `Harvard:rt:test:userfeedback` and all of the referenced resource templates in this and child resource templates. This should help in determining if the user issues are bugs in the Sinopia Editor code base, issues related to specific templates, or the need for user education in the current features.

This will also assist in doing any unit or integration testing based on these issues.


jermnelson:
> I'm trying to connect the dots here. Do I have this right?
> 
>     * the user council is trying to exercise the BFF and surface bugs, concerns etc.
> 
>     * to do this testing, the user council is working with a specific resourceTemplate
> 
>     * as the user council files bug reports, it will facilitate debugging and bug report assessment if we have the same resourceTemplate (and all its descendents) available as a fixture -- so we can easily go through the steps described in a bug report.
> 
> 
> The downside:
> 
>     * typically, "fixtures" are objects used by automated tests, and occasionally for developers as they are working.  Fixtures become something maintained in the code base, and there is usually an effort to keep them targeted towards testing.
> 
> 
> I'm ok with this tradeoff, and I approve this PR if I understood the context correctly.

Yes, thanks @ndushay! Your understand of the context for this PR is correct. We are receiving a lot of bug/enhancement requests based on this RT by the user council testings and having these as fixures would be a value.

PR #1272, opened 2019-08-19
Updating packages

jermnelson:
Updates packages being tracked with `npm` in this repository.


ndushay:
I came to look, but the build failed.  :-(

jermnelson:
> I came to look, but the build failed. :-(

Really weird eslint error that I haven't been able to find anything about and nothing I tried fixed it. The eslint error is:
```javascript 
Error: airbnb-base/whitespace:                                                            
        Configuration for rule "import/no-unused-modules" is invalid:                     
        Value [] should NOT have fewer than 1 items. 
```
Adding an entry to the rules for `import/no-unused-modules` and even using a copied rule from the [documentation](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-unused-modules.md) like `import/no-unused-modules": [1, {"unusedExports": true}])` results in the same error.

PR #1294, opened 2019-08-28
List lookups based on cohort requests LC Profiles

sfolsom:
Assuming https is the desired URI based on existing list entries, even though the canonical is http.

### DIFF
--- a/static/lookupConfig.json
+++ b/static/lookupConfig.json
@@ -824,11 +824,6 @@
     "component": "list"
   },
   {
-    "label": "carrier",
-    "uri": "https://id.loc.gov/vocabulary/carriers",
-    "component": "list"
-  },
-  {
     "label": "class schedule used",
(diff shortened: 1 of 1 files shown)


jermnelson:
Hi @sfolsom, investigating why this PR is failing some of the integration tests. The PR changes are so minor that it shouldn't have these test failures.

sfolsom:
Thanks! Let me know if I did something wrong... that I can help with or improve on in the future.

static/lookupConfig.json
```diff
@@ -823,6 +823,11 @@
     "uri": "https://id.loc.gov/vocabulary/carriers",
     "component": "list"
   },
+  {
```
jermnelson: Duplicate needs removing.

sfolsom: oops. I've made the change to sfolsom-patch-1. (I think I did it correctly, first time updating a patch.)

PR #1304, opened 2019-09-11
add remaining sharevde lookups

elrayle:
The only one not yet completed is for Harry Ransom.  We are still working on some access issues for that one.

This also removes access to the old sharevde data for Stanford and UCSD.


michelleif:
should we also be removing from the Profile Editor lookup dropdown the old ShareVDE lookups:
Screen Shot 2019-09-11 at 1 38 54 PM

PR #1316, opened 2019-09-13
Add latest resource templates and instructions for transforming profi…

justinlittman:
…les to rts.


justinlittman:
This requires loading too many files, so closing this PR until I fix.

PR #1327, opened 2019-09-17
Introduces react-testing-library and replaces InputLiteral tests.

justinlittman:


justinlittman:
Also, separates integration tests, as I couldn't get react-testing-library to work with jest-puppeteer.

PR #1342, opened 2019-09-19
Introduces Cypress and adds a basic end-to-end test.

justinlittman:


src/components/search/Search.jsx
```diff
@@ -81,33 +81,34 @@ const Search = (props) => {
...
-          <div className="form-group" style={{ width: '750px', paddingLeft: '5px' }}>
-            <label className="sr-only" htmlFor="searchInput">Query</label>
-            <div className="input-group" style={{ width: '100%' }}>
-              <input id="searchInput" type="text" className="form-control"
-                     onChange={ event => setQueryString(event.target.value) }
-                     onKeyPress={handleKeyPress} />
-              <span className="input-group-btn">
-                <button className="btn btn-default" type="submit" aria-label="submit search"><FontAwesomeIcon className="fa-search" icon={faSearch} /></button>
-              </span>
+      <div className="container">
```
jcoyne: It's unclear to me why we need to change the source on a PR that is about introducing a new test.

justinlittman: To make it testable. In this case, the reflow of the search box with smaller width screen was pushing it into a location in which it was hidden. https://github.com/LD4P/sinopia_editor/issues/1567 is for completely addressing this problem.

PR #1345, opened 2019-09-20
Removed some resource template fixtures.

justinlittman:


justinlittman:
@jermnelson Did I remove anything critical?

jermnelson:
Hi @justinlittman, removing these RT from the mocks makes sense and I don't think there is anything critical although I liked having the `DiscogsExample.json` for demo purposes but I think it makes more sense to reflect the actual RTs the cohort will be using.

justinlittman:
Yeah @jermnelson I was wondering about Discogs, but suggest that we add a more current version when we do related work this cycle.

PR #1347, opened 2019-09-23
Address all existing lint warnings.

justinlittman:


justinlittman:
This will fix #172, as it addresses the last of the a11y warnings.

PR #1365, opened 2019-09-24
Authorities in QA always request context

### LINKED ISSUE #1338: Determine context configuration from `lookupConfig.json`
currently to show context on a QA lookup, we require the Resource Template to state explicitly that context should be used for a particular source, using the "subtype": "context" parameter

a better way to do this would be to simply show context when available. one strategy would be to include in the lookupconfig file an indication of which lookups have context

the problem with the current method is that:
a) schemas used by most resource templates don't support "subtype"; only the recent 2.0.1 schema supports it, and in order to upgrade a profile from 0.0.9 to 2.0.1 a lot of other cleanup is required, with no easy way to do it (Profile Editor does not support some of the cleanup required). -- there is another ticket to make a 0.0.10 schema to include subtype
b) if context becomes available for an existing authority, users have to go back and modify all resource templates if they want to use that context
c) as pointed out by @kirkhess at LC, whether or not to show context is more of a UI instruction than a data shape. If we want our profiles to be more like shareable application profiles, then the profile isn't really the place for this
d) if context is available, we should show it. so we are not really gaining anything by giving profile creators the choice of whether to show context

if we implement this request, then these tickets are not needed:
https://github.com/LD4P/sinopia_profile_editor/issues/245
https://github.com/LD4P/sinopia/issues/252

jermnelson:
If the QA authority doesn't support context still returning expected results. The `InputLookupQA` component handles the presence of context property in the returned result.  Setting this parameter in the Swagger client to always true may have an impact performance but it also means we don't have to set a context property in each authority in `lookupConfig`. 

Closes #1338

### DIFF
--- a/__tests__/utilities/qa.test.js
+++ b/__tests__/utilities/qa.test.js
@@ -1,17 +1,6 @@
 // Copyright 2019 Stanford University see LICENSE for license
 
-import getSearchResults, { isContext } from 'utilities/qa'
-
-describe('isContext()', () => {
-  // check context property is picked up and when property is not present, method returns false
-  it('detects context subproperty in resource template', () => {
-    expect(isContext({ subtype: 'context' })).toEqual(true)
-  })
-
-  it('detects absence of context subproperty in resource template', () => {
-    expect(isContext({})).toBeFalsy()
-  })
-})
+import getSearchResults from 'utilities/qa'
 
 describe('getSearchResults()', () => {
   const template = {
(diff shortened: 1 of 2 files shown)


jermnelson:
> Can point me to some examples for testing?
Hi @justinlittman, I created the following Resource Template that has three properties, the first using the Discogs authorit (that has context), the second with three ShareVDE Instance authorities that all have context, and the last propertyTemplate contains a non-context authority from QA.

[TestingQAContexts.json.zip](https://github.com/LD4P/sinopia_editor/files/3653161/TestingQAContexts.json.zip)

michelleif:
@jermnelson I like this solution, so simple! In the test template on DEV, the third example LCNAF is showing context, which is expected per https://wiki.duraspace.org/display/LD4P2/Authority+Status+for+Lookups+through+QA

I added a 4th field in a template called jpn:qa:tests_additional and put in on DEV, to test an authority that doesn't support context, and it works

michelleif:
@jermnelson @justinlittman does this impact schema 2.0.1? should we deprecate or remove that schema?

PR #1366, opened 2019-09-24
Display flash message instead of console error

jgreben:
When a DropZone file load fails to execute readAsText, save the error message to the `message` state instead of logging it to the console.


jgreben:
ImportProfiles

PR #1370, opened 2019-09-26
Added stable keys and removed use of suspense to stop remounting of c…

### LINKED ISSUE #1306: Nested defaults for literals don't allow for editing
When a PT with a default literal is in a nested RT, the default literal disappears when edit is clicked.

image

In the image above, "n" disappears from the PT when attempting to edit. 
Browser: chrome, opera
RT: WAU:RT:RDA:Work:monograph

justinlittman:
…omponents.

closes #1306


justinlittman:
This problem was caused by components being unnecessarily remounted, which was resetting state. By stopping the remounting, the components correctly maintain state.

michelleif:
asked user who originally reported to verify that it's fixed on DEV, and it is.

PR #1372, opened 2019-09-26
Copy existing Sinopia Resource

### LINKED ISSUE #1361: Copy Sinopia resource
The user should be able to copy an existing Sinopia resource. This means the resource should be loaded, but be unsaved and not have a Sinopia URI.

The copy can be performed from the search results and from the editor window. For the design from search results see https://github.com/LD4P/sinopia/issues/263. For the editor window, a link should be placed next to Preview RDF. (A separate ticket will change the design of that section of the editor.)

jermnelson:
Adds a new Copy link to the Editor tab:
Screen Shot 2019-09-27 at 9 33 08 PM

Clicking on the **Copy** link creates a new message that the resource is copied:
Screen Shot 2019-09-27 at 9 35 34 PM

When a pre-existing (or saved) resource is loaded into the editor tab:
Screen Shot 2019-09-27 at 9 36 36 PM

Clicking on the **Copy** link removes the resourceURI from state and the old URI is displayed in the tab:
Screen Shot 2019-09-27 at 9 37 37 PM

Getting testing errors in `__tests__/components/editor/property/InputValue.test.jsx` and `__tests__/components/editor/property/PropertyPanel.test.js` that also are present in master branch. 

Would close #1361


package.json
```diff
@@ -94,7 +94,7 @@
     "react-dropzone": "^7.0.1",
     "react-hot-loader": "^4.8.4",
     "react-offcanvas": "^0.4.0",
-    "react-redux": "^6.0.0",
+    "react-redux": "^7.1.1",
```
jcoyne: Should this sort of change be in a separate PR?

src/components/editor/ResourceTemplate.jsx
```diff
@@ -33,6 +34,7 @@ class ResourceTemplate extends Component {
         <div id="resourceTemplate" style={{ marginTop: '-30px' }}>
           <section>
             <h1><em>{this.props.resourceTemplate.resourceLabel}</em></h1>
+            <CopyToNewMessage />
```
jcoyne: I wonder if we only want this to show up for a "saved" (has a URI) resource.

jermnelson: I was debating to only enable copy with "saved" (has a URI) but I thought that a user may have filled out a Resource Template first and then would want to do a copy of their work. @michelleif do you have any opinions on this? I'm happy to change the Copy to only be enabled if the source resource has a URI.

package.json
jermnelson: I'm not sure, I needed to upgrade to the `7.1.1` version to use the React hooks `useDispatch`, `useSelector` from this module.

src/components/editor/ResourceTemplate.jsx
jcoyne: Isn't copying an unsaved resource just going to be a no-op?  How could the original source data be preserved if it wasn't first saved?

justinlittman: I'm in agreement as well -- The link should be disabled for an unsaved resource.

__tests__/components/editor/CopyToNewButton.test.js
```diff
@@ -0,0 +1,71 @@
...
+
+import React from 'react'
+import { createStore } from 'redux'
+import CopyToNewButton from 'components/editor/CopyToNewButton'
+import appReducer from 'reducers/index'
+/* eslint import/no-unresolved: 'off' */
+import { renderWithRedux } from 'testUtils'
+import { fireEvent, wait } from '@testing-library/react'
+
+test('Link button value of Copy', () => {
```
justinlittman: test is OK, but can also use decribe / it.

__tests__/components/editor/CopyToNewMessage.test.js
```diff
@@ -0,0 +1,64 @@
...
+        copyToNewMessage: {
+          show: true,
+        },
+      },
+    },
+  })
+  const { container } = renderWithRedux(
+    <CopyToNewMessage />, store,
+  )
+  expect(container.querySelector('.alert-info')).toBeInTheDocument()
```
justinlittman: Would be preferable to search for text rather than class.

justinlittman: I'm confused. What is being displayed here?

src/components/editor/Editor.jsx
```diff
@@ -35,21 +36,22 @@ const Editor = (props) => {
       <AuthenticationMessage />
       <div className="row">
         <section className="col-md-3" style={{ float: 'right', width: '320px' }}>
+          <CopyToNewButton id="copy-new" />
```
justinlittman: Why are you passing id as a prop?

src/reducers/entities.js
```diff
@@ -21,6 +21,23 @@ export const clearResourceTemplates = (state) => {
   return newState
 }
 
+// Takes current user and optional uri and removes resourceURI
+export const copyResourceToEditor = (state, action) => {
+  const newState = { ...state }
+  const uri = action.uri
+
+  if (uri !== undefined) {
```
justinlittman: Copy from search is going to have to be handled a different way, as loading the resource should not be done in reducer. I suspect it can simply be done as an additional parameter on retrieveResource thunk that would omit setting the URI.

justinlittman:
Also, I think you might need a few test tweaks as well.

src/components/editor/ResourceTemplate.jsx
jermnelson: 👍

__tests__/components/editor/CopyToNewMessage.test.js
jermnelson: This just displays a short message of what the source resource URI used for the copy.

src/components/editor/ResourceTemplate.jsx
jermnelson: Done, commit push up to only enable the Copy button when a resource has an existing URI.

src/components/editor/Editor.jsx
jermnelson: Removed id

src/reducers/entities.js
jermnelson: 👍

__tests__/components/editor/CopyToNewMessage.test.js
jermnelson: Removed this particular test as it no longer being displayed if a user hits copy on an unsaved resource.

PR #1373, opened 2019-09-26
Pass search error message from sinopiaServer function to view

jgreben:
Pass error message and zero search results to redux search state when search fetch fails in order to display a failed search connection message. UPDATED GIF!:

SearchAlert

### DIFF
--- a/src/sinopiaServer.js
+++ b/src/sinopiaServer.js
@@ -193,5 +193,9 @@ export const getSearchResults = async (query, queryFrom = 0) => {
         title: row._source.label,
       })),
     }))
-    .catch(error => console.error(error))
+    .catch(err => ({
+      totalHits: 0,
+      results: [],
+      error: err,
+    }))
 }
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -215,6 +215,12 @@ module.exports = {
       "rules": {
         "import/no-unresolved": "off"
       }
+    },
+    {
+      "files": ["src/actions/index.js"],
+      "rules": {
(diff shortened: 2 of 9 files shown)


src/sinopiaServer.js
```diff
@@ -193,5 +193,11 @@ export const getSearchResults = async (query, queryFrom = 0) => {
         title: row._source.label,
       })),
     }))
-    .catch(error => console.error(error))
+    .catch(error => {
+      return ({
+        totalHits: 0,
+        results: [],
+        error: 'There was a problem retrieving search results. Please try again later.',
```
justinlittman: Please provide the actual error. And remove `Please try again later` as it may not work later depending on the error.

jgreben: > Also, should display message similar to other warnings. (Top of screen, dismissable, full width.)
> 
top of screen is fine, but If the error message is dismissed, and then the user tries subsequent search (without refreshing the page) the error will not reappear. This drawback can maybe overcome by changing SearchResultsMessage() to a Component with constructor props (I tried and failed to make it work using react state hooks), but I ultimately left out the dismiss-able element because (as seen in the GIF) if you perform a successful search (or even just a successful connection to search with zero results) the message clears itself. That said, I am fine trying to re-work the component if making it dismiss-able is a priority.

justinlittman: Dismissable is not a top priority.

jgreben: I found a way to do it using react hooks and drawing from the example in the Editor component.

PR #1376, opened 2019-09-27
Handle QA searches async and independently.

### LINKED ISSUE #1356: Evaluate improving response speed on QA lookups
Review our code and look for performanc bottlenecks. (Up to 2 hour evaluating.)

Note in particular, the use of Promises.all(). 
Create additional followup tickets and link to #1346.

justinlittman:
closes #1356

### DIFF
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -1,5 +1,5 @@
 // Copyright 2019 Stanford University see LICENSE for license
-/* eslint max-params: ["warn", 4] */
+/* eslint max-params: ["warn", 5] */
 
 export const loadingQaResults = () => ({
   type: 'LOADING_QA_RESULTS',
--- a/src/components/editor/property/InputLookupQA.jsx
+++ b/src/components/editor/property/InputLookupQA.jsx
@@ -1,6 +1,6 @@
 // Copyright 2019 Stanford University see LICENSE for license
 
-import React, { useState } from 'react'
+import React, { useState, useRef } from 'react'
 import { Typeahead, asyncContainer } from 'react-bootstrap-typeahead'
 import PropTypes from 'prop-types'
 import SinopiaPropTypes from 'SinopiaPropTypes'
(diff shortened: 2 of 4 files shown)


justinlittman:
In local dev, an example of a InputLookupQA configured with multiple authorities is Monograph Work (BIBFRAME) > Subject of the Work.

You should expect:
* Results are displayed from authorities as they come in. (Rather than wait for all authorities to come in.)
* You should only see results from one search at a time. (So, as you modify your query, results from different searches shouldn't be mixed.)

src/actions/index.js
```diff
@@ -1,5 +1,5 @@
 // Copyright 2019 Stanford University see LICENSE for license
-/* eslint max-params: ["warn", 4] */
+/* eslint max-params: ["warn", 5] */
```
jcoyne: I don't understand why this should change if nothing else in the file changed.

src/components/editor/property/InputLookupQA.jsx
```diff
@@ -23,16 +23,39 @@ const AsyncTypeahead = asyncContainer(Typeahead)
 //  based on values in propertyTemplate.valueConstraint.useValuesFrom
 //  and the lookupConfig for the URIs has component value of 'lookup'
 const InputLookupQA = (props) => {
-  const [authorities, setAuthorities] = useState([])
+  const triggerRenderState = useState('')
```
jcoyne: can this be `const [_, setTriggerRender] = ...`

src/components/editor/property/InputLookupQA.jsx
```diff
@@ -23,16 +23,39 @@ const AsyncTypeahead = asyncContainer(Typeahead)
...
+
+    // resultPromises is an array of Promise<authority>
+    const resultPromises = getSearchResults(query, props.propertyTemplate)
+    resultPromises.forEach((resultPromise) => {
+      resultPromise.then((authority) => {
+        // Only use these results if not cancelled.
+        if (!token.cancel) {
+          allAuthorities.current.push(authority)
+          // Changing state triggers re-render.
+          setTriggerRender(authority)
```
jcoyne: Would there ever be a problem here if the `authority` for the last search is the same as the first authority for the next search? Particularly when a search only uses one authority?

src/utilities/qa.js
```diff
@@ -66,7 +54,7 @@ export const createLookupPromises = (client, query, lookupConfigs) => lookupConf
    */
   const actionFunction = subauthority ? subAuthCall : authorityCall
 
-  return client
+  return Swagger({ spec: swaggerSpec }).then(client => client
```
jcoyne: Does this cause the swagger to be reloaded every time this is called? Would that be slow?

PR #1371, opened 2019-09-26
Loads resource based on resource template and returns unmatched triples

justinlittman:
closes #1357

PR #1385, opened 2019-09-30
Clarified Load by RDF page.

### LINKED ISSUE #1182: Clarify Load RDF tab
**Describe the bug**
When I go "Load RDF" and I enter the following data:

```
<http://example.com/foo> <http://id.loc.gov/ontologies/bibframe/mainTitle> "foo"@en .
<http://example.com/foo> <http://sinopia.io/vocabulary/hasResourceTemplate> "profile:bf2:Title:AbbrTitle" .
<http://example.com/foo> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://id.loc.gov/ontologies/bibframe/AbbreviatedTitle> .
```
and I set `http://example.com/foo` as the root.

Then when I click "Submit" and preview rdf, the root subject (http://example.com/foo) has been lost.

justinlittman:
closes #1182

### DIFF
--- a/src/components/load/LoadByRDFForm.jsx
+++ b/src/components/load/LoadByRDFForm.jsx
@@ -40,19 +40,31 @@ const LoadByRDFForm = (props) => {
     event.preventDefault()
   }
 
+  const n3PlaceHolder = `For example:
+<> <http://id.loc.gov/ontologies/bibframe/mainTitle> "Tractatus Logico-Philosophicus"@en .
+<> <http://sinopia.io/vocabulary/hasResourceTemplate> "resourceTemplate:bf2:WorkTitle" .
+<> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://id.loc.gov/ontologies/bibframe/Title> .
+  `
+
+  const baseURIPlaceholder = 'For example: https://trellis.sinopia.io/repository/stanford/e111a712-5a45-4c2a-9201-289b98d7452e.'
+
   return (
     <div>
-      <h4>Load RDF</h4>
+      <h4>Load RDF into Editor</h4>
       <form id="loadForm" onSubmit={handleSubmit}>
         <div className="form-group">
           <label htmlFor="resourceTextArea">N3 RDF</label>
-          <textarea className="form-control" id="resourceTextArea" rows="15" value={resourceN3} onChange={handleResourceN3Change}></textarea>
+          <textarea className="form-control" id="resourceTextArea" rows="15" value={resourceN3}
+                    onChange={handleResourceN3Change} placeholder={n3PlaceHolder}></textarea>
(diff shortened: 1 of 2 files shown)


justinlittman:
Screen Shot 2019-09-30 at 2 39 17 PM

jgreben:
The clarification text looks good to me. But If the browser window is too small there is no way to see the entire Base URI example text.

Screen Shot 2019-09-30 at 3 08 16 PM

jgreben:
Also, FWIW, when I try to submit the form nothing happens and I get a console error:

`TypeError: this.match(...).toArray(...)[0] is undefined`

Maybe that needs to become a new issue?

justinlittman:
@jgreben Google didn't know how to make it so that the entire Base URI example can show up even if the browser window is made smaller. Do you?

justinlittman:
@jgreben Regarding the `TypeError`: Can you try to replicate on https://github.com/LD4P/sinopia_editor/pull/1371? This PR includes improved error handling for this form.

jgreben:
> @jgreben Google didn't know how to make it so that the entire Base URI example can show up even if the browser window is made smaller. Do you?

How about just making it a `<textarea>` instead?

justinlittman:
I didn't want to make it a text area, as it might confuse the user into thinking that the input should be more than one line. Split into placeholder and help block instead.
Screen Shot 2019-10-01 at 5 17 20 AM

jcoyne:
I agree that `<textarea>` is not an appropriate widget for a URI. I think this change is good.

src/components/load/LoadByRDFForm.jsx
```diff
@@ -40,19 +40,31 @@ const LoadByRDFForm = (props) => {
...
         <div className="form-group">
           <label htmlFor="resourceTextArea">N3 RDF</label>
-          <textarea className="form-control" id="resourceTextArea" rows="15" value={resourceN3} onChange={handleResourceN3Change}></textarea>
+          <textarea className="form-control" id="resourceTextArea" rows="15" value={resourceN3}
+                    onChange={handleResourceN3Change} placeholder={n3PlaceHolder}></textarea>
         </div>
         <div className="form-group">
           <label htmlFor="uriInput">Base URI</label>
-          <input type="text" className="form-control" id="uriInput" value={baseUri} onChange={handleBaseUriChange} />
+          <input type="text" className="form-control" id="uriInput" value={baseUri} onChange={handleBaseUriChange}
```
jcoyne: `type` should be `url` See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/url

justinlittman: This is why I love code review. Always learning something new. (Or reminded of something I once knew.)

Changed.

PR #1387, opened 2019-10-01
Added QA authorities as search options.

### LINKED ISSUE #1377: Add search choices to search tab
* Update search tab to match design described in https://github.com/LD4P/sinopia/issues/255
* Provide configuration of QA authorities to be listed. This can use the same / similar configuration approach as `lookupConfig.json`.
* Note that one of the search options should be "Sinopia" and should perform a search of Sinopia as it currently does.

justinlittman:
closes #1377

### DIFF
--- a/src/components/search/Search.jsx
+++ b/src/components/search/Search.jsx
@@ -5,26 +5,41 @@ import { bindActionCreators } from 'redux'
 import { connect } from 'react-redux'
 import PropTypes from 'prop-types'
 import Header from '../Header'
-import FormGroup from 'react-bootstrap/lib/FormGroup'
-import FormControl from 'react-bootstrap/lib/FormControl'
-import Grid from 'react-bootstrap/lib/Grid'
-import Row from 'react-bootstrap/lib/Row'
-import Form from 'react-bootstrap/lib/Form'
 import Alert from 'react-bootstrap/lib/Alert'
 import fetchSearchResults from 'actionCreators/search'
 import SearchResults from './SearchResults'
 import SearchResultsPaging from './SearchResultsPaging'
 import SearchResultsMessage from './SearchResultsMessage'
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import { faSearch } from '@fortawesome/free-solid-svg-icons'
+import searchConfig from '../../../static/searchConfig.json'
 
 const Search = (props) => {
   const [queryString, setQueryString] = useState('')
   const [showAlert, setShowAlert] = useState(props.error)
-
+  const [authority, setAuthority] = useState('sinopia')
 
   const handleKeyPress = (event) => {
     if (event.key === 'Enter') {
-      props.fetchSearchResults(queryString)
+      search()
       event.preventDefault()
+    }
+  }
+
+  const handleSubmit = (event) => {
+    search()
+    event.preventDefault()
+  }
+
+  const search = () => {
+    if (queryString === '') {
+      return
+    }
+    if (authority === 'sinopia') {
+      props.fetchSearchResults(queryString)
+    } else {
+      // TODO: Implement QA search.
+      console.error(`${authority} not yet supported`)
     }
   }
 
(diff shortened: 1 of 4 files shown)


justinlittman:
Screen Shot 2019-09-30 at 10 47 52 PM
Screen Shot 2019-09-30 at 10 48 05 PM

static/searchConfig.json
```diff
@@ -0,0 +1,82 @@
+[
+  {
+    "label": "SHAREVDE ALBERTA",
```
jcoyne: https://media1.giphy.com/media/ii8rlXOlRFp5LKAVUR/giphy.gif

justinlittman: Or that he copied the existing config ...

src/components/search/Search.jsx
```diff
@@ -43,27 +58,41 @@ const Search = (props) => {
...
-            </FormGroup>
-          </Form>
-        </Row>
-        <SearchResults {...props} key="search-results" />
-        <SearchResultsPaging {...props} pageSize="1"/>
-        <SearchResultsMessage />
-      </Grid>
+      <div className="row">
+        <form className="form-inline" onSubmit={handleSubmit}>
+          <div className="form-group col-md-4" style={{ paddingRight: '5px' }}>
```
jcoyne: form-group elements should not have grid styles (e.g. `col-md-4`)
```

$ bootlint << EOF
>
> <form class="form-inline">
>   <div class="form-group col-md-3"></div>
> </form>
> EOF
...
<stdin>:3:3 E014 Columns (`.col-*-*`) can only be children of `.row`s or `.form-group`s
```

src/components/search/Search.jsx
```diff
@@ -43,27 +58,41 @@ const Search = (props) => {
...
+            <label htmlFor="searchType" style={{ marginRight: '5px' }}>Search</label>
+            <select className="form-control" id="searchType"
+                    value={authority} style={{ paddingRight: '0px', width: '85%' }}
```
jcoyne: In an `.form-inline` aren't select's using `width: auto`?  Why are you diverging from that?

src/components/search/Search.jsx
```diff
@@ -43,27 +58,41 @@ const Search = (props) => {
...
+                    onChange={ event => setAuthority(event.target.value) }
+                    onBlur={ event => setAuthority(event.target.value) }>
+              <option value="sinopia">Sinopia</option>
+              {options}
+            </select>
+          </div>
+          <div className="form-group col-md-6" style={{ paddingLeft: '5px' }}>
+            <div className="input-group" style={{ width: '100%' }}>
```
jcoyne: Isn't the width always 100%?

src/components/search/Search.jsx
```diff
@@ -43,27 +58,41 @@ const Search = (props) => {
...
+            <label htmlFor="searchType" style={{ marginRight: '5px' }}>Search</label>
+            <select className="form-control" id="searchType"
+                    value={authority} style={{ paddingRight: '0px', width: '85%' }}
+                    onChange={ event => setAuthority(event.target.value) }
+                    onBlur={ event => setAuthority(event.target.value) }>
+              <option value="sinopia">Sinopia</option>
+              {options}
+            </select>
+          </div>
+          <div className="form-group col-md-6" style={{ paddingLeft: '5px' }}>
```
jcoyne: can't use columns here.

src/components/search/Search.jsx
```diff
@@ -43,27 +58,41 @@ const Search = (props) => {
...
+            <div className="input-group" style={{ width: '100%' }}>
+              <label className="sr-only" htmlFor="searchInput">Query</label>
+              <input id="searchInput" type="text" style={{ width: '100%' }} className="form-control"
+                     onChange={ event => setQueryString(event.target.value) }
+                     onKeyPress={handleKeyPress} />
+              <span className="input-group-btn">
+                <button className="btn btn-default" type="submit"><FontAwesomeIcon className="fa-search" icon={faSearch} /></button>
```
jcoyne: Need aria (or `.sr-only`) text for screen readers

src/components/search/Search.jsx
```diff
@@ -43,27 +58,41 @@ const Search = (props) => {
...
+                    onChange={ event => setAuthority(event.target.value) }
+                    onBlur={ event => setAuthority(event.target.value) }>
+              <option value="sinopia">Sinopia</option>
+              {options}
+            </select>
+          </div>
+          <div className="form-group col-md-6" style={{ paddingLeft: '5px' }}>
+            <div className="input-group" style={{ width: '100%' }}>
+              <label className="sr-only" htmlFor="searchInput">Query</label>
+              <input id="searchInput" type="text" style={{ width: '100%' }} className="form-control"
```
jcoyne: I believe `.form-control` already sets `width: 100%`

static/searchConfig.json
jcoyne: But it's hard to read. Can we improve it by using normal case?

src/components/search/Search.jsx
jcoyne: `type` should be `url` See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/url

static/searchConfig.json
justinlittman: @michelleif might want these to say something entirely different, e.g., different case, different names, no SHAREVDE. Would prefer to handle in a separate PR in which she can edit as she feels appropriate.

jcoyne:
@justinlittman need to re-eslint this.

PR #1391, opened 2019-10-01
fix getty tgn and ulan pointing to aat

elrayle:
Fix several configs for getty that were pointing to aat when they should point to tgn or ulan.

Identified in https://github.com/LD4P/qa_server/issues/121

### DIFF
--- a/static/lookupConfig.json
+++ b/static/lookupConfig.json
@@ -376,7 +376,7 @@
   {
     "label": "GETTY_TGN (QA)",
     "uri": "urn:ld4p:qa:gettytgn",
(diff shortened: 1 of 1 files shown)


sfolsom:
@elrayle yes, both tgn and ulan are now pointing to the proper vocabs, but I noticed that ulan doesn't seem to be passing context through to Sinopia:

Screen Shot 2019-11-15 at 3 35 31 PM

https://lookup.ld4l.org/authorities/search/linked_data/getty_ulan_ld4l_cache/person?q=washington&maxRecords=4&context=true

PR #1407, opened 2019-10-02
Upgrade to eslint6

jcoyne:


jcoyne:
Looks like this is blocked by https://github.com/airbnb/javascript/issues/2105

PR #1424, opened 2019-10-03
Adds aria-label for close button on CanvasMenu

### LINKED ISSUE #1389: Button on CanvasMenu component is not accessable
**Describe the bug**
No text on the button when using a screen reader. We need `aria-label` or `sr-only` text

jermnelson:
Fixes #1389

### DIFF
--- a/__tests__/components/menu/CanvasMenu.test.js
+++ b/__tests__/components/menu/CanvasMenu.test.js
@@ -25,4 +25,8 @@ describe('<CanvasMenu />', () => {
       expect(node.prop('rel')).toEqual('noopener noreferrer')
     })
   })
+
+  it('has an aria-label for accessability', () => {
+    expect(wrapper.find('button[aria-label="Close Help Menu"]')).toBeTruthy()
+  })
 })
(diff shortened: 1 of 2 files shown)


__tests__/components/menu/CanvasMenu.test.js
```diff
@@ -25,4 +25,8 @@ describe('<CanvasMenu />', () => {
       expect(node.prop('rel')).toEqual('noopener noreferrer')
     })
   })
+
+  it('has an aria-label for accessability', () => {
+    expect(wrapper.find('button[aria-label="Close Help Menu"]')).toBeTruthy()
```
jmartin-sul: `toBeDefined()` seems like it'd be more consistent with the rest of the tests, and makes a bit more sense to me semantically, but `toBeTruthy()` also seems like it'd do the same job just fine in this case, so this isn't something i'd ask for changes over.

feel free to merge if you think the change isn't worth it, or if you do like the idea and wind up implementing it.
```suggestion
    expect(wrapper.find('button[aria-label="Close Help Menu"]')).toBeDefined()
```

PR #1426, opened 2019-10-04
Allows user to select resource template when loading by RDF. Adds use…

### LINKED ISSUE #1358: Allow selecting resource template when loading RDF
When the user loads RDF from the Load RDF tab, the user should have the option of selecting a resource template to load into.

This is an alternative to providing it as a triple like `<> <http://sinopia.io/vocabulary/hasResourceTemplate> "ld4p:RT:bf2:Monograph:Instance" .`

In the case that both are provided, the selected resource template should be used.

In the case that none are provided, an error should be displayed to the user on the Load RDF tab.

https://github.com/LD4P/sinopia_editor/blob/resource_state_builder/src/components/load/LoadByRDFForm.jsx is a reference implementation.

### LINKED ISSUE #1415: Use useResource hook in LoadByRDF
To remove redundant code.

justinlittman:
…Resource hook to LoadByRDFForm.

closes #1415
closes #1358

### DIFF
--- a/__tests__/components/load/LoadByRDFForm.test.js
+++ b/__tests__/components/load/LoadByRDFForm.test.js
@@ -18,9 +18,20 @@ const createInitialState = () => {
       resource: {},
       entities: {
         resourceTemplates: {},
+        resourceTemplateSummaries: {
+          'resourceTemplate:bf2:WorkTitle': {
+            key: 'resourceTemplate:bf2:WorkTitle',
+            name: 'Work Title',
+            id: 'resourceTemplate:bf2:WorkTitle',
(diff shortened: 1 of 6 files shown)


justinlittman:
To test, load RDF like the following on the Load RDF page:
```
<> <http://id.loc.gov/ontologies/bibframe/mainTitle> "foo"@en .
<> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://id.loc.gov/ontologies/bibframe/Title> .
```

You will be prompted to select a resource template.

jcoyne:
Please resolve conflicts

PR #1443, opened 2019-10-04
Discogs search should use the endpoint configured by authorityConfig.json

### LINKED ISSUE #1420: Error searching discogs on search tab
1480F572-063F-41FD-A0B9-ECC36FAD3415

jcoyne:
Fixes #1420

This switches the store from using "authority" to using "uri" which is a unique reference in the `authorityConfig.json`.  Using this `authorityConfig.json` allows it to know that "discogs" is a non-LD source.

### DIFF
--- a/__tests__/actionCreators/search.test.js
+++ b/__tests__/actionCreators/search.test.js
@@ -25,7 +25,7 @@ describe('fetchSinopiaSearchResults', () => {
     expect(dispatch).toBeCalledWith({
       type: 'SET_SEARCH_RESULTS',
       payload: {
-        authority: 'sinopia',
+        uri: 'sinopia',
         query: '*',
         searchResults: mockSearchResults.results,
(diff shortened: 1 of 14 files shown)


src/utilities/searchConfig.js
```diff
@@ -0,0 +1,7 @@
+// Copyright 2019 Stanford University see LICENSE for license
+
+import lookupConfig from '../../static/lookupConfig.json'
+
+export const findSearchConfig = searchUri => lookupConfig.find(configItem => configItem.uri === searchUri)
```
justinlittman: Confusing to call this method findSearchConfig when actually finding in lookupConfig.

jcoyne: Do you mind if I rename the whole thing to `authorityConfig`?   It is used for both lookup and searching.

justinlittman: That works for me.

PR #1510, opened 2019-10-09
Display environment in header (from config)

peetucket:
beginning work on LD4P/sinopia#268

Todo:

- [x] how do I fetch the env variable (doesn't seem to work as expected now)
- [x] how is the env variable set?
- [x] selectively display only if not blank?
- [x] include a test


.circleci/config.yml
```diff
@@ -59,7 +59,7 @@ jobs:
           name: Build & Register Images
           command: |
             docker build -t ld4p/sinopia_editor:latest .
-            docker build -t ld4p/sinopia_editor:dev --build-arg TRELLIS_BASE_URL=$TRELLIS_BASE_URL_DEV --build-arg SINOPIA_URI=$SINOPIA_URI_DEV --build-arg AWS_COGNITO_DOMAIN=$AWS_COGNITO_DOMAIN_DEV --build-arg COGNITO_CLIENT_ID=$COGNITO_CLIENT_ID_DEV --build-arg COGNITO_USER_POOL_ID=$COGNITO_USER_POOL_ID_DEV --build-arg INDEX_URL=$INDEX_URL_DEV --build-arg SEARCH_HOST=$INDEX_URL_DEV .
+            docker build -t ld4p/sinopia_editor:dev --build-arg TRELLIS_BASE_URL=$TRELLIS_BASE_URL_DEV --build-arg SINOPIA_ENV=DEV SINOPIA_URI=$SINOPIA_URI_DEV --build-arg AWS_COGNITO_DOMAIN=$AWS_COGNITO_DOMAIN_DEV --build-arg COGNITO_CLIENT_ID=$COGNITO_CLIENT_ID_DEV --build-arg COGNITO_USER_POOL_ID=$COGNITO_USER_POOL_ID_DEV --build-arg INDEX_URL=$INDEX_URL_DEV --build-arg SEARCH_HOST=$INDEX_URL_DEV .
```
mjgiarlo: You're missing a `--build-arg` in `--build-arg SINOPIA_ENV=DEV SINOPIA_URI=$SINOPIA_URI_DEV/` (before `SINOPIA_URI`)

.circleci/config.yml
```diff
@@ -75,7 +75,7 @@ jobs:
       - run:
           name: Build & Register Images
           command: |
-            docker build -t ld4p/sinopia_editor:$CIRCLE_TAG-stage --build-arg TRELLIS_BASE_URL=$TRELLIS_BASE_URL_STAGE --build-arg SINOPIA_URI=$SINOPIA_URI_STAGE --build-arg AWS_COGNITO_DOMAIN=$AWS_COGNITO_DOMAIN_STAGE --build-arg COGNITO_CLIENT_ID=$COGNITO_CLIENT_ID_STAGE --build-arg COGNITO_USER_POOL_ID=$COGNITO_USER_POOL_ID_STAGE --build-arg INDEX_URL=$INDEX_URL_STAGE --build-arg SEARCH_HOST=$INDEX_URL_STAGE .
+            docker build -t ld4p/sinopia_editor:$CIRCLE_TAG-stage --build-arg TRELLIS_BASE_URL=$TRELLIS_BASE_URL_STAGE --build-arg SINOPIA_ENV=STAGE SINOPIA_URI=$SINOPIA_URI_STAGE --build-arg AWS_COGNITO_DOMAIN=$AWS_COGNITO_DOMAIN_STAGE --build-arg COGNITO_CLIENT_ID=$COGNITO_CLIENT_ID_STAGE --build-arg COGNITO_USER_POOL_ID=$COGNITO_USER_POOL_ID_STAGE --build-arg INDEX_URL=$INDEX_URL_STAGE --build-arg SEARCH_HOST=$INDEX_URL_STAGE .
```
mjgiarlo: You're missing a `--build-arg` in `--build-arg SINOPIA_ENV=STAGE SINOPIA_URI=$SINOPIA_URI_STAGE/` (before `SINOPIA_URI`)

src/Config.js
```diff
@@ -41,7 +41,13 @@ class Config {
   }
 
   static get sinopiaEnv() {
-    return process.env.SINOPIA_ENV || '' // default to production, which requires no additional display
+    if (process.env.SINOPIA_ENV) {
```
mjgiarlo: What's up above should be equivalent to this

.circleci/config.yml
peetucket: thanks, fixed

src/Config.js
peetucket: yeah, it is now, bit I'm going to add in a ' - ' to the set value for display purposes (instead of doing that conditional in the view component code)

mjgiarlo: oh dang, need my :eyeglasses:

src/Config.js
```diff
@@ -40,6 +40,16 @@ class Config {
...
   }
 
+  static get sinopiaEnv() {
+    if (process.env.SINOPIA_ENV) {
+      return `- ${process.env.SINOPIA_ENV}`
+    }
+    else
+    {
+      return ''
+    }
```
mjgiarlo: Purely aesthetic, but I have a preference for the more guard-clause-y incantation if you are open to it:

```javascript
if (process.env.SINOPIA_ENV) {
  return `- ${process.env.SINOPIA_ENV}`
}

// We do not set this value in production, and don't want to see the env label in production
return ''
```

__tests__/Config.test.js
```diff
@@ -97,6 +98,10 @@ describe('Config', () => {
       expect(Config.sinopiaUrl).toEqual('https://sinopia.foo')
     })
 
+    it('sinopia env overrides static value for display', () => {
```
mjgiarlo: To be consistent with the other values in here, you should add a static test above too for when the env var is undefined.

mjgiarlo:
@peetucket if you add one more test to test the default env value in config.js (the empty string when the env var is undefined), I think this is about ready to go

__tests__/components/Header.test.js
```diff
@@ -3,16 +3,29 @@
...
+      SINOPIA_ENV: 'TEST',
+    }
+  })
+
+  it('sinopia env overrides static value for display', () => {
+    expect(Config.sinopiaEnv).toEqual(' - TEST')
+  })
+
+  it('displays the Sinopia text with the environment name', () => {
+    expect(wrapper.find('h1.editor-logo').text()).toBe('LINKED DATA EDITOR - TEST')
```
peetucket: There is one failing test, but I'm not sure why.  The `Config.sinopiaEnv` variable is getting setup correctly (as confirmed by the previous passing test) but is not being picked up the component correctly.  It works fine in the browser.

PR #1523, opened 2019-10-11
Adds copy uri button.

### LINKED ISSUE #1411: Copy URI button appears after Resource is Saved
## Describe the feature you'd like
After a resource is saved and published in Sinopia's Editor tab, display a  **Copy URI** button that copies the newly minted URI to the user's computer buffer. They can then paste the new URI in a Sinopia lookup instead of searching for the newly created entity in a different resource template.

## Give an example
Screen Shot 2019-10-02 at 4 07 11 PM

## Describe alternatives you've considered
The user could select the new URI and manually copy it into their computer's buffer.

justinlittman:
closes #1411


justinlittman:
Sinopia_Linked_Data_Editor

PR #1533, opened 2019-10-14
Fixes display of validation errors and availability of save button.

### LINKED ISSUE #1459: Resources are not validated on save
Steps to reproduce:
1. Create a valid resource.
2. Save it.
3. Change it so that it is invalid (e.g., delete the value for a required field).
4. Save it.

Expected result:
Resource is not saved and user is notified of errors.

Actual result:
Resource is saved.

justinlittman:
closes #1459

### DIFF
--- a/__tests__/components/editor/SaveAndPublishButton.test.js
+++ b/__tests__/components/editor/SaveAndPublishButton.test.js
@@ -1,29 +1,125 @@
 // Copyright 2019 Stanford University see LICENSE for license
-
 import React from 'react'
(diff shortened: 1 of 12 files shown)


justinlittman:
For the save button to be enabled:
* A resource must have been changed.
* If validation errors are being displayed, there are no errors.

When the save button is clicked:
* If there are validation errors, they are displayed.
* Otherwise, the resource is saved or the group choice modal is displayed.

PR #1548, opened 2019-10-14
Changed findResourceTemplate selector to take state

justinlittman:
To be consistent with other selectors that take state as param, rather than selectorReducer.


justinlittman:
Closing in favor of #1557.

PR #1550, opened 2019-10-14
Refactor Editor to use useSelector

jcoyne:


__tests__/components/editor/Editor.test.js
```diff
@@ -1,55 +1,138 @@
...
 
 describe('<Editor />', () => {
   describe('any user', () => {
-    const wrapper = shallow(<Editor.WrappedComponent {...props}/>)
+    const store = createReduxStore(createInitialState())
 
-    it('has div with id "editor"', () => {
-      expect(wrapper.find('div#editor').length).toBe(1)
-    })
     it('renders <ResourceTemplate /> component', () => {
```
justinlittman: Since the test setup is exactly the same for all of these tests, I'd recommend collapsing into a single `it`.

__tests__/components/editor/Editor.test.js
```diff
@@ -1,55 +1,138 @@
...
-      expect(wrapper.find(ResourceTemplate).length).toBe(1)
+      const { queryByText } = renderWithRedux(
```
justinlittman: Courtesy of @jgreben, there is now a renderWithReduxAndRouter method: https://github.com/LD4P/sinopia_editor/blob/master/__tests__/test_utilities/testUtils.js#L17

__tests__/components/editor/Editor.test.js
jcoyne: Cool. I wasn't sure if that would be too big of a change, but I'm on board with that.

PR #1564, opened 2019-10-16
Fix to InputLang test.

justinlittman:
This is failing for me locally.

### DIFF
--- a/__tests__/components/editor/property/InputLang.test.js
+++ b/__tests__/components/editor/property/InputLang.test.js
@@ -52,22 +52,13 @@ describe('<InputLang />', () => {
     expect(queryByText(expected)).toBeInTheDocument()
   })
 
-  it('change to match text changes input text', async () => {
-    global.document.createRange = () => ({
-      setStart: () => {},
-      setEnd: () => {},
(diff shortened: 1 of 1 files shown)


__tests__/components/editor/property/InputLang.test.js
```diff
@@ -52,22 +52,13 @@ describe('<InputLang />', () => {
...
 
-  it('change to match text changes input text', async () => {
-    global.document.createRange = () => ({
-      setStart: () => {},
-      setEnd: () => {},
-      commonAncestorContainer: {
-        nodeName: 'BODY',
-        ownerDocument: document,
-      },
-    })
```
jmartin-sul: possibly stupid questions:  what was the intent behind overriding `createRange` in the `document`?  do the other changes obviate the need for that, or did it turn out it was always superfluous?  (asking for a friend)

justinlittman: Dunno. @jermnelson?

jermnelson: I think this may have been leftover cruft from early testing on my part.  Glad to see it removed.

PR #1597, opened 2019-10-22
Validates that value template refs must have unique property URIs.

### LINKED ISSUE #1529: Validate that value template refs have unique resource URIs
When loading a resource, if a property has multiple value template refs need to determine which to use. This is done by trying to match the property against the resource URI of each of the value template refs. If there is more than one, an error is thrown: https://github.com/LD4P/sinopia_editor/blob/master/src/ResourceStateBuilder.js#L151

To prevent users from creating resources that can't be loaded, a check should be added to ResourceTemplateValidator to make sure value template refs have unique resource URIs. This will prevent the user for opening the editor with a bad resource template.

justinlittman:
closes #1529

### DIFF
--- a/src/ResourceTemplateValidator.js
+++ b/src/ResourceTemplateValidator.js
@@ -1,7 +1,8 @@
 // Copyright 2019 Stanford University see LICENSE for license
 
-import { foundResourceTemplate } from 'sinopiaServer'
+import { foundResourceTemplate, getResourceTemplate } from 'sinopiaServer'
 import { getTagNameForPropertyTemplate } from 'utilities/propertyTemplates'
+
 
 /**
  * Validates a resource template.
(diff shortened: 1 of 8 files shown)


justinlittman:
This can be tested with `test:RT:bf2:RareMat:Instance`, which is available as a fixture.

src/ResourceTemplateValidator.js
```diff
@@ -129,5 +131,40 @@ const validateTemplateRefsExist = async (resourceTemplate) => {
   return []
 }
 
+/**
+ * Validates that all value template refs have unique property URIs.
```
jermnelson: Should this be *all  value template refs have unique resource URIs*?

justinlittman: Yup. Fixed.

PR #1600, opened 2019-10-22
Added direct LOC Special Playback Characteristics list

sfolsom:


jermnelson:
I think we can merge this PR but will wait until @hudajkhan reviews as requested by @sfolsom

PR #1603, opened 2019-10-22
Allow user to upload rt or profile

### LINKED ISSUE #1602: A user should be able to upload a single resource template in additional to a profile in the resource template list page

Currently you can only upload a profile containing multiple resource templates. It would be useful to also allow a user to upload a single resource template file.

peetucket:
fixes #1602 

Allows a user to upload either a single resource template or a profile. Also, changed some labels to match the new reality.

Note: Some of the work was already done (validation code works fine in both scenarios).

To do:
- [x] test


justinlittman:
This is awesome, @peetucket. Looks like all it needs is one additional test in ImportResourceTemplate.

peetucket:
added a unit test and started adding a cypress integration test based on the existing one for a profile upload, but unclear if that is useful (I don't see any expectations in the current one)

PR #1604, opened 2019-10-23
Bookmarkable URL for resource templates

justinlittman:


justinlittman:
To test:
1. Open a resource template from resource template list page.
2. Observe the URL.

PR #1605, opened 2019-10-23
Adds searching text field and weights fields.

### LINKED ISSUE #1517: Sinopia search default behavior
* Should search title field and all text field, with title field more heavily weighed.
* Results should be relevancy ordered by default.

justinlittman:
closes #1517


justinlittman:
Note that testing this will require a pipeline running https://github.com/LD4P/sinopia_indexing_pipeline/pull/85

justinlittman:
You would expect that a resource with a term in the title will appear before resources with that term in a different field.

PR #1607, opened 2019-10-23
Scroll to top to when resource template error.

### LINKED ISSUE #1522: Make resource template warning messages visible even when you are not at top of Resource Templates list
See https://github.com/LD4P/sinopia_editor/pull/1521#pullrequestreview-300734031 regarding the placement of the resource template warning message at the top of the screen, where it may not be visible.

justinlittman:
closes #1522


justinlittman:
To test locally using fixture:
1. Go to the template list page.
2. Scroll down to "test invalid resource with refs and defaults" and click.

Expected result:
The error message is visible.

PR #1614, opened 2019-10-24
override bootstrap CSS to ensure text is black against white background

### LINKED ISSUE #1576: using arrow key to move through list of matching values in QA lookup field that has extended context, don't turn the text to white when focus is on that row
for a field that is a lookup to a QA source with extended context (DEV, template "_Monograph Work (BIBFRAME)", field "Subject of the Work")

search for something like coffee

get result list
Screen Shot 2019-10-17 at 10 24 33 PM

use down arrow to move down the result list
Screen Shot 2019-10-17 at 10 24 42 PM

for the rows with white background, the text changes to white and then you can't read it at all

for the tan background, the text changes from black to white, and is hard to read

need another way to show focus of arrow-ing, instead of whiting out the text.

peetucket:
fixes #1576 

when rendering drop-down menus in the LookupComponent, we should be sure the text is black (since the background is white), so we can read it

### DIFF
--- a/src/styles/main.css
+++ b/src/styles/main.css
@@ -340,6 +340,10 @@ input.rbt-input-main {
   padding-left: 8px;
 }
 
+.dropdown-item.active {
+  color: #000000;
+}
+
 th.search-header {
   background-color: #F8F6EF;
 }


jermnelson:
LGTM; will approve once merge conflict is resolved.

peetucket:
rebased and force pushed

PR #1615, opened 2019-10-24
left justify search result button text

### LINKED ISSUE #1587: on Search results tab, use left justify for text that wraps (rather than centered)
on Search tab, results with long labels are wrapping, but are being centered. should be left-justified instead
Screen Shot 2019-10-18 at 11 18 28 AM

peetucket:
fixes #1587 

ass new class to allow the button style to left justify text (and add necessary class to search result rendering component)


peetucket:
this no longer appears to be needed based on changes in #1612

PR #1620, opened 2019-10-24
Refactor errors

justinlittman:


justinlittman:
Errors should:
* Be displayed to user.
* Multiple errors should be displayed in multiple alerts.
* Alerts should be displayed on and pertain to the appropriate page.

Previously:
* Not all relevant errors were displayed to users.
* Multiple errors would overwrite each other.
* Errors might display on incorrect pages. (So, for example, a resource template error might appear on the Search page and the Load by RDF page.)

src/ResourceTemplateValidator.js
```diff
@@ -36,7 +36,7 @@ const validateRepeatedPropertyTemplates = (resourceTemplate) => {
     }
   })
   if (dupes.length > 0) {
-    return [`Repeated property templates with same property URI (${dupes}) are not allowed.`]
+    return [`Validation error for ${resourceTemplate.resourceURI}: Repeated property templates with same property URI (${dupes}) are not allowed.`]
```
jermnelson: The string `Validation error for ${resourceTemplate.resourceURI}` is repeated multiple times in the `ResourceTemplateValidator` module and could be a variable that is defined once

src/actionCreators/resourceTemplates.js
```diff
@@ -13,53 +12,50 @@ import { findResourceTemplate } from 'selectors/entitySelectors'
...
+      if (_.isEmpty(errors)) {
         dispatch(setResourceTemplate(resourceTemplate))
         return resourceTemplate
       }
-      dispatch(setRetrieveResourceTemplateError(resourceTemplateId, reason))
+      errors.forEach(error => dispatch(appendError(errorKey, error)))
     }).catch((err) => {
       console.error(err)
-      dispatch(setRetrieveResourceTemplateError(resourceTemplateId, err.toString()))
+      dispatch(appendError(errorKey, `Error retrieving ${resourceTemplateId}: ${err.toString()}`))
```
jermnelson: Instead of dispatching this error and writing it to the console, could you raise this as an error and the outer catch (starting on line 30) would handle the dispatch and console logging?

justinlittman: Fixed.

src/ResourceTemplateValidator.js
justinlittman: Fixed.

PR #1642, opened 2019-10-28
New useModalCss hook for Modals

### LINKED ISSUE #1553: Refactor css Modal classes as a hook
In the modals React component functions, the css classes `['modal', 'fade'] are hardcoded a number of times. These could be moved into a hook that is called by the wrapped modal instead

jermnelson:
Fixes #1553

### DIFF
--- a/src/components/ResourceTemplateChoiceModal.jsx
+++ b/src/components/ResourceTemplateChoiceModal.jsx
@@ -2,7 +2,7 @@
 
 import React, { useState, useMemo } from 'react'
 import { useDispatch, useSelector } from 'react-redux'
-import ModalWrapper from 'components/ModalWrapper'
+import ModalWrapper, { useModalCss } from 'components/ModalWrapper'
 import PropTypes from 'prop-types'
 import { hideModal } from 'actions/index'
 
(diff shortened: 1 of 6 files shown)


src/components/ResourceTemplateChoiceModal.jsx
```diff
@@ -2,19 +2,17 @@
...
+import ModalWrapper, { useModalCss } from 'components/ModalWrapper'
 import PropTypes from 'prop-types'
 import { hideModal } from 'actions/index'
 
 const ResourceTemplateChoiceModal = (props) => {
   const dispatch = useDispatch()
   const show = useSelector(state => state.selectorReducer.editor.modal === 'ResourceTemplateChoiceModal')
 
-  const classes = ['modal', 'fade']
   let display = 'none'
```
justinlittman: `const display = show ? 'block' : 'none'` would be more concise and clearer.

src/components/editor/GroupChoiceModal.jsx
```diff
@@ -34,16 +34,14 @@ const GroupChoiceModal = (props) => {
     event.preventDefault()
   }
 
-  const classes = ['modal', 'fade']
   let display = 'none'
```
justinlittman: Same as above and elsewhere.

justinlittman: And perhaps this should be made a function in ModalCss as well?

jermnelson: Done, using new `useDisplayStyle` hook

PR #1643, opened 2019-10-29
Fixed LoadByRDF so that only tries to use resource on submit. Changed…

### LINKED ISSUE #1608: trying to open a particular existing SInopia resource gives error
on stage, search for existing resource by its URI https://trellis.stage.sinopia.io/repository/stanford/d8452c5a-60c3-466a-903e-8c3e0aeb5497 or by word (right now the search string gefen finds only this resource)

try to open the found resource

error message:
```There was a problem retrieving https://trellis.stage.sinopia.io/repository/stanford/d8452c5a-60c3-466a-903e-8c3e0aeb5497: TypeError: this.state.entities.resourceTemplates[e] is undefined```

@justinlittman investigated and says: there seems to be a problem with referencing a RT (sinopia:resourceTemplate:bf2:Item:Enumeration) that exists but is returning a 406.

I attempted to re-import that RT and did so successfully. Sinopia didn't prompt me about overwriting (so it seems like that RT did NOT exist?) but even after I imported it, I still get the same error above when trying to open the Resource in question.

what is going wrong?

justinlittman:
… resource thunks so that template loading errors are properly handled.

closes #1608

### DIFF
--- a/__tests__/actionCreators/resources.test.js
+++ b/__tests__/actionCreators/resources.test.js
@@ -82,8 +82,8 @@ describe('retrieveResource', () => {
     const mockStore = configureMockStore([thunk])
     store = mockStore(blankState)
   })
-  // sinopiaServer.getResourceTemplate.mockImplementation(getFixtureResourceTemplate)
-  describe('when dispatch to existing resource returns undefined', () => {
+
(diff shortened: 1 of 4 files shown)


justinlittman:
Sinopia

PR #1647, opened 2019-10-30
Always set multiple prop to true and use function to set disabled prop

jgreben:
One way to leverage the Typeahead component to do what we want here is to use the component's multiple prop in a different way: it should always be set to true, and we can use the component's disabled prop to prevent subsequent multiple entries by passing it the boolean result of an expression that checks both the resource template's repeatable prop along with the length of the selections.

- Removing the disabled prop of the `<Token/>` component allows items to always be removed.
- Use a component ref to force a blur of the input.

intuitive

### DIFF
--- a/__tests__/components/editor/property/InputLookupQA.test.js
+++ b/__tests__/components/editor/property/InputLookupQA.test.js
@@ -1,92 +1,277 @@
 // Copyright 2019 Stanford University see LICENSE for license
 
 import React from 'react'
-import { shallow } from 'enzyme'
 import InputLookupQA from 'components/editor/property/InputLookupQA'
+import { fireEvent, wait, waitForElement } from '@testing-library/react'
+import { showValidationErrors, validateResource } from 'actions/index'
+import {
+  renderWithRedux, assertRDF, createReduxStore, setupModal,
+} from 'testUtils'
 
-const plProps = {
-  id: 'lookupComponent',
-  propertyTemplate: {
-    mandatory: 'false',
-    repeatable: 'false',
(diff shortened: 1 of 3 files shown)


jermnelson:
PR works as expected but I noticed that after a value is entered into a lookup field with `repeatable` set to *false*, the field is highlight like it has the focus even when the focus is on another field, see
Screen Shot 2019-10-30 at 1 03 54 PM

__tests__/components/editor/property/InputLookupQA.test.js
```diff
@@ -67,8 +67,30 @@ describe('<InputLookupQA />', () => {
     })
   })
 
-  it('sets the typeahead component multiple attribute according to the repeatable property from the template', () => {
-    expect(wrapper.find('#lookupComponent').props().multiple).toBeFalsy()
+  it('sets the typeahead component multiple attribute to true so user inputs are always tokenized', () => {
+    expect(wrapper.find('#lookupComponent').props().multiple).toBeTruthy()
+  })
+
+  // TODO: re-write tests using react-testing-library or remove this test
```
jgreben: Do these tests need to be re-written using react-testing-library? There does not seem to be an easy way to mock a react hook using enzyme, or I need help with it.

justinlittman: I think any time we find ourselves needing to mock a hook, we're testing the wrong thing or wrong way.

jgreben: 👌

PR #1658, opened 2019-11-01
Template search

justinlittman:


justinlittman:
Sinopia

justinlittman:
Sinopia

justinlittman:
Note that this depends on the latest indexing pipeline container and will require a reindex for any existing resource templates to be searchable.

PR #1669, opened 2019-11-04
Add a Cypress spec for the help and resources tab

### LINKED ISSUE #1666: Write Cypress test for Help & Resources
To test that Help and Resources can be opened and closed from the home page and each of the Sinopia tabs.

This is to prevent a regression of the bug that prompted https://github.com/LD4P/sinopia_editor/pull/1660.

I tried unsuccessfully to write this with RTL but was problematic because the offset is handled in the root component. This is more suited for a cypress test.

jcoyne:
Fixes #1666

### DIFF
--- a/cypress/integration/helpAndResources.spec.js
+++ b/cypress/integration/helpAndResources.spec.js
@@ -1,0 +1,103 @@
+describe('Looking at the Help and Resources gutter', () => {
+  it('Opens the app', () => {
+    cy.visit(Cypress.env('EDITOR_URL') || 'http://localhost:8000/')
+    cy.contains('The underdrawing for the new world of linked data in libraries')
+  })
+
+  it('Opens help and resources', () => {
+    cy.contains('a', 'Help and Resources').click()
+
+    cy.contains('External Identifier Sources and Vocabularies')
+  })
+
+  it('Closes help and resources', () => {
+    cy.contains('a', 'Help and Resources').click()
+
+    cy.contains('External Identifier Sources and Vocabularies')
+  })
+
+  it('Logs in', () => {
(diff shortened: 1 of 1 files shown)


cypress/integration/helpAndResources.spec.js
```diff
@@ -0,0 +1,49 @@
...
+      .type(Cypress.env('COGNITO_TEST_USER_NAME'))
+      .should('have.value', Cypress.env('COGNITO_TEST_USER_NAME'))
+    cy.get('#password')
+      .type(Cypress.env('COGNITO_TEST_USER_PASS'))
+      .should('have.value', Cypress.env('COGNITO_TEST_USER_PASS'))
+    cy.contains('Login').click()
+    cy.contains('current cognito user: sinopia-devs_client-tester')
+  })
+
+  it('Opens Linked Data Editor', () => {
```
jgreben: I noticed that the issue specified that the Help and Resources menu could be opened from each of the Sinopia tabs, but this test only tests the Resource Templates tabs. @justinlittman, is this sufficient?

justinlittman: Each of tabs need to be tested separately. (The bug that initiated this was only on some tabs.)

jcoyne: Covering every permutation of features (thinking broader than the issue at hand) in cypress is going to create a very slow test suite.  Can we just cover the cases that caused the trouble?

justinlittman: Philosophically I agree with you. However, in this case we want to make sure that a dev doesn't break the menu for any tab (as there are 2 places in code each tab can be broken, and it is not immediately obvious). So this isn't just a case of "test everything" but rather test a specific vulnerability due to the structure of the code that actually occurred (I broke some of the tabs, not all.)

PR #1675, opened 2019-11-04
[WIP] dismiss popovers when user clicks elsewhere

### LINKED ISSUE #1659: clear popover when user leaves page/tab
when user is in Edit tab and clicks the "i" icon to open a popover, the popover is currently staying on the screen even when you leave the Edit tab

on DEV open Sinopia:RT:Demo:SampleResourceFuller

click the "i" icon on fields 20 and 22 (you can just do one of them but it will be easier to see the bug if you open both)
Screen Shot 2019-11-01 at 9 04 25 AM

then leave and go to another tab like Search. the popovers are at the upper left of the screen
Screen Shot 2019-11-01 at 9 04 41 AM

peetucket:
SEE ALTERNATE APPROACH IN #1678 

fixes #1659 by simply dismissing the popover on next click anywhere

using documentation from https://getbootstrap.com/docs/4.0/components/popovers/ that indicates we need to switch this to an `<a>` element to produce the desired behavior

### DIFF
--- a/__tests__/components/editor/property/PropertyLabelInfoTooltip.test.js
+++ b/__tests__/components/editor/property/PropertyLabelInfoTooltip.test.js
@@ -16,7 +16,7 @@ describe('<PropertyLabelInfoTooltip />', () => {
   const wrapper = shallow(<PropertyLabelInfoTooltip {...props} />)
 
   it('displays a tooltip from the label', () => {
-    expect(wrapper.find('span[data-toggle="popover"]').length).toEqual(1)
+    expect(wrapper.find('a[data-toggle="popover"]').length).toEqual(1)
   })
 
   it('renders an info icon', () => {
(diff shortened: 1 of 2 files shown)


michelleif:
hmmm, I told users they could keep multiple popovers open and it was awesome!

peetucket:
> hmmm, I told users they could keep multiple popovers open and it was awesome!

ok, let me see what else i can dig up then - this popover behavior is all coming from bootstrap, so we are probably dependent on what they provide.

peetucket:
How about this: https://github.com/LD4P/sinopia_editor/pull/1678

It just hides any popovers when you click one of the navigation tab links.  One potential downside is that it will do this even if you get the "unsaved work" confirmation dialog box but then cancel it out.  But it has the advantage of being pretty simple to implement otherwise.

michelleif:
sounds good thanks for revisiting

peetucket:
closed in favor of #1678

PR #1678, opened 2019-11-04
hide popovers when a nav link is clicked

### LINKED ISSUE #1659: clear popover when user leaves page/tab

peetucket:
alternate approach to fix #1659 (as compared to work in #1675)

popover documentation at https://getbootstrap.com/docs/4.0/components/popovers/

this simply hides all popovers when a navigation/tab is clicked

### DIFF
--- a/src/components/Header.jsx
+++ b/src/components/Header.jsx
@@ -8,6 +8,12 @@ import { connect } from 'react-redux'
 import _ from 'lodash'
 
 class Header extends Component {
+  hidePopovers() {
+    if (window.$('.popover').popover) {
+      window.$('.popover').popover('hide')
+    }
+  }
+
   render() {
     return (
       <div className="editor-navbar">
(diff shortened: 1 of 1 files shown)


src/components/Header.jsx
```diff
@@ -8,6 +8,10 @@ import { connect } from 'react-redux'
 import _ from 'lodash'
 
 class Header extends Component {
+  hidePopovers() {
+    window.$('.popover').popover('hide')
```
peetucket: this works fine in browser, but makes the tests unhappy :(

jgreben: Maybe this?
```
if(window.$('.popover').popover) {
      window.$('.popover').popover('hide')
    }
```

peetucket: that worked, thanks

PR #1684, opened 2019-11-05
Move the modal messages into redux state

jcoyne:
This helps us move toward making ImportResourceTemplate a functional component.


__tests__/components/editor/actions/PreviewButton.test.js
```diff
@@ -17,6 +17,9 @@ describe('When the button is clicked', () => {
         modal: {
           name: undefined,
         },
+        modal: {
```
justinlittman: Why is modal in twice?

jcoyne: I think that was introduced in a merge conflict.  I'll remove it.

PR #1689, opened 2019-11-06
 Move flash messages to redux state 

jcoyne:
This will enable ImportResourceTemplate to be made functional.

### DIFF
--- a/src/components/templates/CreateResourceMessages.jsx
+++ b/src/components/templates/CreateResourceMessages.jsx
@@ -1,15 +1,18 @@
 // Copyright 2019 Stanford University see LICENSE for license
 
 import React from 'react'
-import PropTypes from 'prop-types'
+import { useSelector } from 'react-redux'
+import { flashMessages } from 'selectors/flashSelectors'
 
 /**
  * An info message that is populated when retrieving a resource template
  */
-const CreateResourceMessages = (props) => {
-  if (props.messages.length === 0) return null
+const CreateResourceMessages = () => {
+  const messages = useSelector(state => flashMessages(state))
 
-  const messageItems = props.messages.map(message => <li key={message}>{message}</li>)
+  if (messages.length === 0) return null
+
+  const messageItems = messages.map(message => <li key={message}>{message}</li>)
 
   return (
     <div className="alert alert-info">
@@ -18,10 +21,6 @@ const CreateResourceMessages = (props) => {
       </ul>
     </div>
   )
-}
-
-CreateResourceMessages.propTypes = {
-  messages: PropTypes.array,
 }
 
 export default CreateResourceMessages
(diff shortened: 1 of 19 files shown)


jcoyne:
@justinlittman so is `s/flash:/create_resource_messages:/` a sufficient change?

src/components/templates/CreateResourceMessages.jsx
```diff
@@ -1,15 +1,18 @@
...
+import { useSelector } from 'react-redux'
+import { createTemplateMessages } from 'selectors/flashSelectors'
 
 /**
  * An info message that is populated when retrieving a resource template
  */
-const CreateResourceMessages = (props) => {
-  if (props.messages.length === 0) return null
+const CreateResourceMessages = () => {
+  const messages = useSelector(state => createTemplateMessages(state))
```
justinlittman: Recommend calling the selector `findCreateTemplateMessages` because `createTemplateMessages` sounds like an action.

src/components/templates/ImportResourceTemplate.jsx
```diff
@@ -12,16 +12,10 @@ import CreateResourceMessages from './CreateResourceMessages'
 import { createResourceTemplate, updateResourceTemplate } from 'sinopiaServer'
 import { getCurrentUser } from 'authSelectors'
 import { clearModalMessages, addModalMessage, showModal } from 'actions/modals'
+import { clearTemplateMessages, setTemplateMessages } from 'actions/flash'
```
justinlittman: So there are other sorts of template messages (e.g., error messages). Would be clearer if consistently called these create template messages, like "clearCreateTemplateMessages".

src/components/templates/ImportResourceTemplate.jsx
```diff
@@ -131,11 +125,10 @@ class ImportResourceTemplate extends Component {
   render() {
     return (
       <div id="importResourceTemplate">
-        <UpdateResourceModal messages={this.props.modalMessages}
-                             update={this.handleUpdateResource} />
+        <UpdateResourceModal update={this.handleUpdateResource} />
         <Header triggerEditorMenu={this.props.triggerHandleOffsetMenu}/>
         <ImportFileZone setResourceTemplateCallback={this.setResourceTemplates} />
-        <CreateResourceMessages messages={this.state.flashMessages} />
+        <CreateResourceMessages />
```
justinlittman: Should this be "CreateResourceTemplateMessages"?

src/components/templates/ImportResourceTemplate.jsx
jcoyne: These are sometimes "update" messages too.

src/components/templates/ImportResourceTemplate.jsx
jcoyne: Possibly.  This wasn't changed or created in this PR though.

src/components/templates/CreateResourceMessages.jsx
jcoyne: Yes, I've made this change.

PR #1692, opened 2019-11-06
Lookup URI labels now persist and also load from RDF

### LINKED ISSUE #1641: Store labels for URIs
when a field offers a lookup (QA or Sinopia or LOC-direct), we show labels for the user to pick from; once user selects a label, we just store the URI

so that means no one can ever search for the label when trying to find something.

we want to store the label as well as the URI, and index the label, so it can be searched (this is for all kinds of lookup fields incl Sinopia)

here's a use case where you'd want to search by label, from a Stanford cataloger:

Resource title is Annual Report. 
Resource agent is Bank of Uganda; its URI is http://id.loc.gov/authorities/names/n85142007 
I want to see if something like this is already in Sinopia
I will not be able to search for the Agent "Bank of Uganda" because that is its label and Sinopia has not saved/indexed the label. 
Assuming titles are saved as text strings: I could search "Annual Report" and find thousands of existing descriptions. 
(For context: In Stanford's SearchWorks, a title search for Annual Report yields over 54000 results.) 
How do I figure out if one of these is for my Bank of Uganda one? 
I would need to memorize the URI and search for it?? Or look for it in some display? 
Let's say I get 54000 results under this title. Would I need to examine each one to see if any of them contained the URI for Bank of Uganda?

jermnelson:
When labels for URIs are present in a component's items, the label is used to generate a triple using `rdfs:label` in the Graph builder. When loading RDF back into the editor, the items label is set if to the `<uri> rdfs:label <label> ` is present.

Fixes #1641

### DIFF
--- a/__tests__/GraphBuilder.test.js
+++ b/__tests__/GraphBuilder.test.js
@@ -267,4 +267,39 @@ describe('GraphBuilder', () => {
       expect(result.length).toEqual(0)
     })
   })
+
+  describe('when the state has item labels', () => {
+    const state = createBlankState()
+    state.selectorReducer.entities.resourceTemplates = {
+      'ld4p:RT:bf2:Agents:RWO:Country': {
+        resourceURI: 'http://id.loc.gov/ontologies/bibframe/Work',
+      },
+    }
+    state.selectorReducer.resource = {
+      'ld4p:RT:bf2:Agents:RWO:Country': {
+        'http://www.loc.gov/mads/rdf/v1#associatedLocale': {
+          items: [
+            {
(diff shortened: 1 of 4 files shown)


jcoyne:
What if the upstream service returns an erroneous result?  How will we be able to identify or correct that?

jermnelson:
> What if the upstream service returns an erroneous result? How will we be able to identify or correct that?

Do you mean QA result or what is returned from Trellis?

jermnelson:
@justinlittman, added test for GraphBuilder.

Issue #838, opened 2019-06-26
better tests for SinopiaResourceTemplates component


jmartin-sul:
in particular, we don't have unit tests for confirming that:
* the error message displays when there's a problem connecting to trellis
* the table contents display correctly based on the results that come back from trellis

this ticket was spawned by #822, where i ran into some trouble testing as thoroughly as i'd have liked (though i did add a test in that PR to make sure the expected column headers display, and @jcoyne pointed out that there are integration tests that confirm that templates display appropriately in the resource template list table).

the two tests i was going to add were along the lines of:
```js
    describe('when there is an error connecting to server', () => {
      it('renders an error message if there is no connection to the server', () => {
        expect.assertions(1)
        listResourcesInGroupContainer.mockImplementation(async () => {
          throw 'uh oh!'
        })
        // listResourcesInGroupContainer.mockReturnValue(new Promise((_, reject) => { reject(Error('uh oh!')) }))

        const component = renderRoutes()
        console.info(`component.text()=${component.text()}`)
        const errMsg = 'No connection to the Sinopia Server is available, or there are no resources for any group.'
        expect(component.find('div.alert').text()).toEqual(errMsg)
      })
    })
```

```js
    it('renders the table rows with the expected data', () => {
      expect.assertions(3)

      listResourcesInGroupContainer.mockImplementation(async group => listFixtureResourcesInGroupContainer(group))
      getResourceTemplate.mockImplementation(async templateId => getFixtureResourceTemplate(templateId))

      const component = renderRoutes()
      const tableRowData = component.find('table#resource-template-list tbody tr')
        .map(trWrapper => trWrapper.find('td').map(tdWrapper => tdWrapper.text()))
      console.info(`component.text()=${component.text()}`)
      console.info(`tableRowData=${tableRowData}`)

      expect(tableRowData).toContain(['Barcode', 'resourceTemplate:bf2:Identifiers:Barcode', '', ''])
      expect(tableRowData).toContain(['RT RDA Manifestation for monographs', 'rt:rda:manifestation:monograph', 'Crystal Clements (cec23@uw.edu)', ''])
      expect(tableRowData).toContain(['Instance Title', 'resourceTemplate:bf2:Title', '', 'Title information relating to a resource: work title, preferred title, instance title, transcribed title, translated title, variant form of title, etc.'])
    })
```

[...]

PR #1713, opened 2019-11-08
Returns better error message if client cannot connect to backend

jermnelson:
Towards fixing #838

### DIFF
--- a/src/sinopiaSearch.js
+++ b/src/sinopiaSearch.js
@@ -22,7 +22,16 @@ export const getSearchResults = async (query, queryFrom = 0, size = Config.searc
   const url = `${Config.searchHost}${Config.searchPath}`
 
   return fetch(url, { method: 'POST', headers: { 'Content-Type': 'application/json' }, body: JSON.stringify(body) })
-    .then(resp => resp.json())
+    .then((resp) => {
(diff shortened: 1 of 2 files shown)


src/sinopiaSearch.js
```diff
@@ -70,9 +70,17 @@ export const getTemplateSearchResults = async (query) => {
   }
 
   const url = `${Config.searchHost}${Config.templateSearchPath}`
-
   return fetch(url, { method: 'POST', headers: { 'Content-Type': 'application/json' }, body: JSON.stringify(body) })
-    .then(resp => resp.json())
+    .then((resp) => {
+      if (resp.status > 300) {
```
mjgiarlo: Can you say more about what the significance of 300 is here? If Trellis does return a 300, do you not want this to catch? I might have expected `>= 300` or `> 299` instead to flag all non-3xx, 4xx, and 5xx responses.

jermnelson: Thanks for catching this! No particular significance of using 300 code for handling, changing to use `>=` operators.

mjgiarlo: @jermnelson super! thanks :smile:

PR #1719, opened 2019-11-08
Support multiple resources

justinlittman:


justinlittman:
image

astridu:
I prefer tabs, but if that's challenging, pills are ok too.

justinlittman:
Switched to tabs and `state.selectorReducer.entities`.

jcoyne:
Just one test needs a fixup.

PR #1738, opened 2019-11-13
Refactors InputLookupSinopia and InputLookupQA

### LINKED ISSUE #1153: allow profile to limit inputLookupSInopia to Works or to Instances
ORIGINAL TICKET: Initially we want to support selected typeaheads that add three types of searches, BIBFRAME Works, Instances, and Items. Once [30](ttps://github.com/LD4P/sinopia_indexing_pipeline/issues/30) is fixed, we can support this functionality a couple of different ways:
-  Create specific Sinopia URNs for each of three classes (ie.e BIBFRAME Work urn:ld4p:sinopia:work)
- Use the same `urn:ld4p:sinopia` urn but add a subauthority value for the specific RDF class.

Pros and cons to both approaches.

Blocked by https://github.com/LD4P/sinopia_indexing_pipeline/issues/30

CHALLENGE:
Not all Works will be bf:Work, see notes below. It's also possible that some resources will have type that's a subclass of bf:Work. Would need to take those two possibilities into account.

justinlittman:
Refactors InputLookupSinopia and InputLookupQA to use common component. Adds limiting InputLookupSinopia searches by type.

closes #1153


justinlittman:
image

src/components/editor/property/InputLookupQA.jsx
```diff
@@ -85,107 +32,14 @@ const InputLookupQA = (props) => {
...
-                      onSearch={search}
-                      renderToken={(option, props, idx) => renderTokenFunc(option, props, idx)}
-                      {...typeaheadProps}
-                      filterBy={() => true}
-      />
-      {error && <span className="text-danger">{error}</span>}
-    </div>
-  )
-}
+    <React.Fragment>
```
jcoyne: why do we need the `Fragment` here?

src/components/editor/property/InputLookupSinopia.jsx
```diff
@@ -1,123 +1,47 @@
...
-                      placeholder={props.propertyTemplate.propertyLabel}
-                      minLength={1}
-                      filterBy={() => true }
-                      allowNew={() => true }
-                      id="sinopia-lookup" />
-
-      <span className="text-muted">Use a * to wildcard your search.</span>
-      {error && <span className="text-danger">{error}</span>}
-    </div>
+    <React.Fragment>
```
jcoyne: Do we need Fragment here?

src/components/editor/property/InputLookupQA.jsx
justinlittman: No. Removed.

src/components/editor/property/InputLookupSinopia.jsx
justinlittman: No. Removed.

Issue #1730, opened 2019-11-11
Errors in one template are stopping me from opening others


sfolsom:
**Describe the bug**
What went wrong? What did you expect to happen?

Errors in one template are stopping me from opening others

- What is the impact on your work?

I have to close out and start over before opening a different Template.

- What environment are you in? (stage or production)

Stage

- What OS and browser are you using? (Mac/Windows; Safari, Firefox, etc.)

Mac/Chrome

- If you're cataloging, what Resource Template were you using (name or ID)

sinopia:resourceTemplate:bf2:Identifiers:LCCN from within CornellSinatra45:Instance.

**Please include a screenshot**
(Mac: shift-command-4; Windows: ; then drag the image file into the issue)

![Screen Shot 2019-11-11 at 10 25 00 AM](https://user-images.githubusercontent.com/8257258/68602996-7511b180-0475-11ea-8336-efb4ed2d9916.png)

justinlittman:
@sfolsom Can you please provide exact steps to reproduce? I'm must be missing something.

justinlittman:
In stage, :
1. Open CornellSinatra45:Work.
2. Add "Has Bibframe instance".
3. Within that, click Add Identifiers.
An error occurs.
4. Now go back to Resource Templates tab.
5. Try to open an resource template (that should open).

Expected result:
Resource template opens.

Actual result:
Same error as above.

justinlittman:
So I'm going to declare this a "fingers crossed", meaning that I think the problem is already fixed in current code. We can re-test after next release.

I also fixed some related defects that testing this revealed.

michelleif:
seems fixed on DEV. will keep this open until original reporter can verify on stage

justinlittman:
@sfolsom Can you verify if this is fixed?

sfolsom:
I fixed the error in the profile that was was causing other profiles not to open, so it was hard to recreate the problem. I did find another error in the same set of profiles that doesn't cause other profiles open, so maybe it's fixed. Off to fix the other errors in my profiles...

justinlittman:
Feel free to re-open if necessary.

PR #1743, opened 2019-11-14
Renders resource even when error. Fix search related reducers so that…

justinlittman:
… trigger renders due to state change.

refs #1730

### DIFF
--- a/__tests__/components/editor/ResourceTemplate.test.js
+++ b/__tests__/components/editor/ResourceTemplate.test.js
@@ -13,7 +13,8 @@ describe('<ResourceTemplate />', () => {
   const wrapper = shallow(<ResourceTemplate.WrappedComponent resourceTemplateId="resourceTemplate:bf2:Note"
                                                              resourceTemplate={resourceTemplate}
                                                              retrieveResourceTemplate={jest.fn()}
-                                                             unusedRDF="foo" />)
+                                                             unusedRDF="foo"
+                                                             errors={['Oooops']} />)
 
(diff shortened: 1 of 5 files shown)


justinlittman:
This addresses 2 problems related to #1730:
* After creating a resource from a resource template and returning to the resource template page, resource template search wasn't working.
* User request to not "lose" an entire resource when there was an error from a nested resource template.

justinlittman:
Also, fixed an html id that was set on a table that could have multiple instances (so the id was not unique).

__tests__/components/editor/ResourceTemplate.test.js
```diff
@@ -37,7 +38,17 @@ describe('<ResourceTemplate />', () => {
...
   it('displays Alert', () => {
+    expect(wrapper.find(Alerts).length).toEqual(1)
+  })
+
+  describe('When there is an error and no resource template', () => {
     const alertWrapper = shallow(<ResourceTemplate.WrappedComponent errors={['Oooops']} />)
-    expect(alertWrapper.find(Alerts).length).toEqual(1)
+    it('displays Alert', () => {
+      expect(alertWrapper.find(Alerts).length).toEqual(1)
+    })
```
mjgiarlo: This looks like the same stuff on lines 17 + 40-42. Intended to test it in two places, or cruft?

justinlittman: Different tests. In the first case, there is a resource template; in the second there is not. The code handles differently.

PR #1746, opened 2019-11-14
Removes special handling of non-LD QA results.

justinlittman:
QA is switching to use the same context format in non-LD result as LD results.

### DIFF
--- a/__tests__/__fixtures__/DiscogsLookup.json
+++ b/__tests__/__fixtures__/DiscogsLookup.json
@@ -1,0 +1,25 @@
+{
+  "id": "test:resource:DiscogsLookup",
+  "resourceLabel": "Testing discogs lookup",
+  "remark": "This hits discogs",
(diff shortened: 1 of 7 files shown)


justinlittman:
Pointing at QA staging (which has the context changes):
image

justinlittman:
To test this, you can change https://lookup.ld4l.org/authorities to https://lookup-stg.ld4l.org/authorities in `apidoc.json`.

The release of this to production will be coordinated with the release of QA server containing the accompanying changes.

src/lib/apidoc.json
```diff
@@ -10,7 +10,7 @@
   },
   "servers": [
     {
-      "url": "https://lookup.ld4l.org/authorities",
+      "url": "https://lookup-stg.ld4l.org/authorities",
```
jermnelson: Should this be reverted back to the production QA URI?

justinlittman: Umm, definitely.

PR #1751, opened 2019-11-15
Switches to only add template to history when opened by user.

### LINKED ISSUE #1750: in "Most recently used templates", show ONLY templates explicitly opened by user, NOT templates nested within those
if the user opens a template which itself refers to other templates, those referred templates then become part of the "Most recently used templates" list

In the recently used list, I think users expect to see ONLY the templates that they explicitly opened.

example on DEV:
first, before I open any Resource Templates in my session
Screen Shot 2019-11-15 at 9 14 35 AM

then I open just one template (id = ld4p:RT:bf2:Cartographic:Instance) and go back to recently used, and there's a list not just of ld4p:RT:bf2:Cartographic:Instance but of other templates
Screen Shot 2019-11-15 at 9 14 51 AM

justinlittman:
closes #1750


justinlittman:
To test:
1. Open a resource template that has nested resources.
2. Return to Resource Templates tab.
3. Toggle open recently used resource templates.

Expected result:
Only the resource template that you opened is listed. (Or rather, only resource templates that you opened from the resource template tab are listed.)

PR #1760, opened 2019-11-16
Switch to using iso-639-2

### LINKED ISSUE #1230: change language choices for literals to ISO 639-2
**Describe the bug**
I am using the sinopia user feedback template in Windows/Firefox and I don't understand where the language choices are coming from and what appears to be a random order.  Not all the languages in the MARC list of languages appear to be there.  For example, when I enter Turkish I would expect to see Turkish as well as Ottoman Turkish, but I don't.

jcoyne:
Still need to see about how this handles old data.

Fixes #1230

### DIFF
--- a/__tests__/GraphBuilder.test.js
+++ b/__tests__/GraphBuilder.test.js
@@ -44,7 +44,7 @@ describe('GraphBuilder', () => {
                 items: {
                   '3TzRpgv65': {
                     content: 'Very colorful',
-                    lang: 'en',
+                    lang: 'eng',
                   },
                   '5TzRpgv72': {
                     content: 'Sparkly',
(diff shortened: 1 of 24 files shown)


justinlittman:
Nice. A 1 character change required modifying 24 files ...

justinlittman:
Sees to hang on ...
image

jcoyne:
@justinlittman 60-70% is tests, which says something about the coupling of our tests and maybe about Redux (global state).

jcoyne:
@justinlittman Ha! the iso639-2.json response uses iso-639-1 languages in the response:

Screen Shot 2019-11-18 at 8 06 16 AM

justinlittman:
I thought tight coupling was the whole point of react/redux ...

PR #1772, opened 2019-11-18
Adds context to QA search results.

### LINKED ISSUE #1693: in Search tab, show extended context when searching QA resources
details coming

justinlittman:
closes #1693


justinlittman:
It's because this assumes the new non-LD context, which is only on stage. If you set `apidoc.json` to use https://lookup-stg.ld4l.org/authorities you will get Discogs context.

However, I'm not handling album covers because I don't want to write more Discogs special handling. (If users disagree, we can add in a separate ticket.)

PR #1803, opened 2019-11-19
scroll to top on edit template error

### LINKED ISSUE #1758: Fix scroll to top of search page
https://github.com/LD4P/sinopia_editor/pull/1686 scrolls when there is a search error; scrolling to top is needed when there is an error opening or copying a resource from the results list.

Thus:
* Remove scroll to top on search error. (When searching, you will always be at the top.)
* Add scrolling to top when there is an error opening or copying a resource from the results list for Sinopia search results.
* Add scrolling to top when there is an error opening a resource from the results list for QA search results.

peetucket:
fixes #1758 

- added scroll to top for opening a template from QA search results
- added new scroll to top code for the error case of opening a resource that triggers an error (like a validation error on the resource template)

### DIFF
--- a/src/components/search/SinopiaSearchResults.jsx
+++ b/src/components/search/SinopiaSearchResults.jsx
@@ -1,7 +1,7 @@
 // Copyright 2019 Stanford University see LICENSE for license
 /* eslint max-params: ["error", 4] */
 
-import React, { useState, useEffect } from 'react'
+import React, { useState, useEffect, useRef } from 'react'
 import { connect } from 'react-redux'
 import { bindActionCreators } from 'redux'
 import PropTypes from 'prop-types'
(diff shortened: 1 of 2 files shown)


src/components/search/SinopiaSearchResults.jsx
```diff
@@ -1,7 +1,7 @@
 // Copyright 2019 Stanford University see LICENSE for license
 /* eslint max-params: ["error", 4] */
 
-import React, { useState, useEffect } from 'react'
+import React, { useState, useEffect, useRef } from 'react'
```
justinlittman: Same change is needed for QASearchResults.jsx as well.

peetucket: done, thanks for the help

PR #1806, opened 2019-11-19
Normalize Class Column Names for Sinopia and QA search results

### LINKED ISSUE #1774: in Sinopia search, rename "Type" column
on Search tab, when searching Sinopia, change title of "Type" column to "Class"

andfor QA (external) searches, change the column name to "Class" (currently is "Classes") to be consistent

jermnelson:
Closes #1774

### DIFF
--- a/src/components/search/QASearchResults.jsx
+++ b/src/components/search/QASearchResults.jsx
@@ -134,7 +134,7 @@ const QASearchResults = (props) => {
                 Label
               </th>
               <th className="search-header" style={{ width: '25%' }}>
-                Classes
+                Class
               </th>
               <th className="search-header" style={{ width: '25%' }}>
                 Context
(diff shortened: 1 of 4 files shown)


src/components/search/QASearchResults.jsx
```diff
@@ -133,8 +133,8 @@ const QASearchResults = (props) => {
               <th className="search-header" style={{ width: '40%' }}>
                 Label
               </th>
-              <th className="search-header" style={{ width: '25%' }}>
-                Classes
+              <th className="search-header" style={{ width: '40%' }}>
```
justinlittman: 40 + 40 + 25 + 10 > 100

jermnelson: Yikes! Restored previous width.

PR #1811, opened 2019-11-19
best attempt to get to design of new import button

### LINKED ISSUE #1792: Change Import resource template / profile button
To match:
image

If it cannot easily be placed adjacent to search box, placing it slightly above is acceptable.

peetucket:
(sort of) fixes #1792 

I was able to more or less get the search bar and import button on the same line, but if you shrink your screen down, the search box overruns the import button (due to a fixed pixel setting for the width of the search box).  If you do the right thing and make the search box width = 100%, then it comes out too small to show the placeholder text.  I can't figure out how to get the search bar and the import button to make better use of the space available otherwise.

### DIFF
--- a/src/components/templates/ImportFileZone.jsx
+++ b/src/components/templates/ImportFileZone.jsx
@@ -4,6 +4,8 @@ import React, { Component } from 'react'
 import { connect } from 'react-redux'
 import DropZone from './DropZone'
 import PropTypes from 'prop-types'
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import { faUpload } from '@fortawesome/free-solid-svg-icons'
 import Ajv from 'ajv' // JSON schema validation
 import Config from 'Config'
 import { setResourceTemplates } from 'actionCreators/resourceTemplates'
@@ -189,7 +191,6 @@ class ImportFileZone extends Component {
     const importFileZone = {
       display: 'flex',
       justifyContent: 'right',
-      padding: '10px',
     }
     const dropzoneContainer = {
       display: 'flex',
@@ -209,9 +210,11 @@ class ImportFileZone extends Component {
     return (
       <section>
         <div className="ImportFileZone" style={importFileZone}>
+          <FontAwesomeIcon style={ { marginTop: '9px' } } icon={faUpload} />
           <button id="ImportProfile"
-                  className="btn btn-primary btn-lg"
-                  onClick={this.handleClick}>Import a Profile or Resource Template</button>
+                  className="btn btn-link"
+                  onClick={this.handleClick}>
+                  Import Profile / Resource Template</button>
         </div>
         <div className="dropzoneContainer" style={dropzoneContainer}>
           { this.state.showDropZone ? this.renderDropZone() : null }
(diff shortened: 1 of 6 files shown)


src/components/templates/ImportFileZone.jsx
```diff
@@ -209,9 +210,11 @@ class ImportFileZone extends Component {
     return (
       <section>
         <div className="ImportFileZone" style={importFileZone}>
+          <FontAwesomeIcon icon={uploadIcon} />
           <button id="ImportProfile"
-                  className="btn btn-primary btn-lg"
-                  onClick={this.handleClick}>Import a Profile or Resource Template</button>
+                  className="btn-link"
```
justinlittman: Should be `"btn btn-link"`.

src/components/templates/ImportResourceTemplate.jsx
```diff
@@ -13,9 +13,17 @@ const ImportResourceTemplate = props => (
   <section id="importResourceTemplate">
     <UpdateResourceModal />
     <Header triggerEditorMenu={props.triggerHandleOffsetMenu}/>
-    <ImportFileZone />
+    <div className="container">
```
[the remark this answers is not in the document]
peetucket: reverted

src/components/templates/TemplateSearch.jsx
```diff
@@ -41,8 +41,8 @@ const TemplateSearch = () => {
         <div className="col">
           <form className="form-inline" onSubmit={event => event.preventDefault()}>
             <div className="form-group" style={{ paddingBottom: '10px', paddingTop: '10px' }}>
-              <label htmlFor="searchInput">Find a resource template</label>&nbsp;
-              <div className="input-group" style={{ width: '750px', paddingLeft: '5px' }}>
+              <label className="font-weight-bold" htmlFor="searchInput">Find a resource template</label>&nbsp;
+              <div className="input-group" style={{ width: '500px', paddingLeft: '5px' }}>
```
justinlittman: Can you set this back to 750px?

src/components/templates/ImportFileZone.jsx
```diff
@@ -209,9 +210,11 @@ class ImportFileZone extends Component {
     return (
       <section>
         <div className="ImportFileZone" style={importFileZone}>
+          <FontAwesomeIcon icon={faUpload} />
```
justinlittman: Sorry to be fiddly, but can you adjust the icon alignment?
image

src/components/templates/TemplateSearch.jsx
peetucket: yes, though i think its better to not force it that wide (when you make the viewport smaller, it looks worse)

src/components/templates/ImportFileZone.jsx
peetucket: yeah, that bugged me too, but i had a hard time doing it in CSS...easy fix is to make it part of the link, but that makes the icon blue and underlines it like this:

Screen Shot 2019-11-20 at 10 24 02 AM

peetucket: never mind - aligned correctly

PR #1813, opened 2019-11-19
Change button style

### LINKED ISSUE #1783: Change button style
image

On the Editor page:
* Save should be primary
* Close should be secondary
* Copy URI should be secondary

On the Load by RDF page:
* Submit should be primary

jcoyne:
Fixes #1783


jcoyne:
@justinlittman this is the recommended way to customize bootstrap https://getbootstrap.com/docs/4.3/getting-started/theming/ and it's what we do on all of our Rails projects in DLSS.

justinlittman:
Approve pending fix conflict.

justinlittman:
Environment in header does not wrap.
image

PR #1824, opened 2019-11-21
Styles resource tabs.

### LINKED ISSUE #1789: Resource template tab styling
To match (within reason):
image

Note that this ticket does not include close buttons, open another, or navigate multiple shown above.

Also note that the current navs are using a style that is being overridden, causing the tabs not to render properly.

justinlittman:
closes #1789


justinlittman:
image

justinlittman:
Made some tweaks to be consistent with the header and to emphasize the selected resource.

PR #1826, opened 2019-11-21
Use devDependencies appropriately

jcoyne:


justinlittman:
Can you say more about this change?

PR #1828, opened 2019-11-21
Allow user to select lookups

### LINKED ISSUE #1776: improve navigation of lots of lookup results in fields that are lookups to multiple sources
a field of type lookup can have multiple external sources (useValuesFrom). the display can be overwhelming for the cataloger especially if the first listed source isn't the one they want to see results from, in which case they have to do a lot of scrolling to get down to the source they want to see

suggestion from user is to have the result sets collapsed and let the user open them

@astridu to design

an example of a field with multiple sources is in this template:
https://development.sinopia.io/editor/ld4p:RT:bf2:Monograph:Work:Un-nested
field = "Subject of the Work"

justinlittman:
closes #1776

### DIFF
--- a/src/components/editor/property/InputListLOC.jsx
+++ b/src/components/editor/property/InputListLOC.jsx
@@ -1,6 +1,8 @@
 // Copyright 2019 Stanford University see LICENSE for license
 
-import React, { useState, useEffect, useMemo } from 'react'
+import React, {
+  useState, useEffect, useMemo, useRef,
+} from 'react'
 import { Typeahead } from 'react-bootstrap-typeahead'
 import defaultFilterBy from 'react-bootstrap-typeahead/lib/utils/defaultFilterBy'
 import PropTypes from 'prop-types'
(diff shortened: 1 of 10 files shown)


justinlittman:
image

src/components/editor/property/InputLookup.jsx
```diff
@@ -47,18 +66,19 @@ const InputLookup = (props) => {
     tokens.current.push(token)
 
     // resultPromises is an array of Promise<result>
-    const resultPromises = props.getLookupResults(query, propertyTemplate)
+    const resultPromises = props.getLookupResults(query, Object.values(selectedLookupConfigs))
     resultPromises.forEach((resultPromise) => {
       resultPromise.then((resultSet) => {
         // Only use these results if not cancelled.
         if (!token.cancel) {
+          // console.log('resultSet', resultSet)
```
jermnelson: Cruft?

src/components/editor/property/InputListLOC.jsx
```diff
@@ -82,6 +91,20 @@ const InputListLOC = (props) => {
     changeSelections(payload)
   }
 
+  const lookupCheckboxes = Object.values(allLookupConfigs).map((lookupConfig) => {
```
jgreben: I wonder if it would be worth making a component of this (despite the rule of threes)?

justinlittman: Yeah, we do that with SearchResultRows. Not a huge fan, as it makes the code less clear.

PR #1830, opened 2019-11-21
UI tweaks

justinlittman:


justinlittman:
In Sinopia search results, adjusted widths and renamed "Modified Date" to "Modified".
image

justinlittman:
In resource template results, combined label and id into single column.
image

PR #1834, opened 2019-11-21
Add the modal backdrop

jcoyne:
Previously we were not activating the bootstrap modal component, just using it's styles, so the backdrop was not activating.

Screen Shot 2019-11-21 at 2 42 33 PM

### DIFF
--- a/cypress/integration/end2end.spec.js
+++ b/cypress/integration/end2end.spec.js
@@ -43,6 +43,7 @@ describe('End-to-end test', () => {
           const fileContent = JSON.stringify(fileJson)
           cy.get('input[type="file"]').upload({ fileContent, fileName, mimeType: 'application/json' })
         })
+        cy.get('button[data-dismiss=modal]').contains('Overwrite').click({ force: true }) // allow overwrite of templates if they exist aleady for some reason
         // Waiting for indexing. If this proves problematic, can try a different approach.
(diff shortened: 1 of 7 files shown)


justinlittman:
I've taken a look at the test and cannot figure out why it isn't working. (Basically, the Save button is clicked but nothing happens.) It does work correctly if only the css changes are made. I'd rather lose the curtain effect then lose more time, so let's go with css only.

peetucket:
This looks like a similar problem that I was having with my recently merged branch that moved the login panel to the home page.  I am rebasing from master and fiddling around to see if I can get it to pass here too.  If so, I could do a force push with my local rebased branch

peetucket:
I was able to get the cypress tests to work by rebasing from master and rebuilding my local docker containers

peetucket:
Damn - still busted even though it works locally. One of the issues I noticed running interactively locally was that if the resource templates it tries to upload are already in Trellis, it pops up the "overwrite?" modal and then this blocks the test from continuing, and they fail in this exact way.  This cleared locally by just restarting the docker containers, I presumed this would happen automatically in CI (assuming that is what is happening here).

justinlittman:
Oh wait, closing this in favor of https://github.com/LD4P/sinopia_editor/pull/1862

PR #1844, opened 2019-11-22
Paginate the template results

jcoyne:


src/components/templates/TemplateSearch.jsx
```diff
@@ -1,58 +1,81 @@
...
-            </div>
-          </form>
+            </form>
+          </div>
         </div>
       </div>
-    </div>
+
+      <SinopiaResourceTemplates history={props.history} />
+      { hasNextButton && (<button className="btn btn-outline-secondary" onClick={incrementPage}>Next Page</button>)}
```
justinlittman: As noted in the ticket, please add pagination similar to Sinopia search results (and re-using the Search Results Paging component).

jcoyne: That componetn is not reusable as it is tightly bound to the redux state of the sinopia search.

justinlittman: If I'm not mistaken, the structure of `state.selectorReducer.search` and `state.selectorReducer.templateSearch` are similar. It would just need to be unbound such that the redux path of the search result to use needs to be passed in.

jcoyne: I think it's a lot more than that.  I'm getting:
Screen Shot 2019-11-25 at 12 01 25 PM

justinlittman: Also encountered this in some refactoring of search. Will fix.

__tests__/components/templates/TemplateSearch.test.js
```diff
@@ -81,38 +82,45 @@ describe('<TemplateSearch />', () => {
...
-      const { getByPlaceholderText } = renderWithRedux(
+      const { getByPlaceholderText, unmount } = renderWithReduxAndRouter(
         <TemplateSearch />, store,
       )
 
       const input = getByPlaceholderText(/Enter id, label/)
-      await assertTemplate(store, result1.id)
-      await assertTemplate(store, result2.id)
-      await assertTemplate(store, result3.id)
+      // await assertTemplate(store, result1.id)
```
jcoyne: Cruft?

justinlittman:
Adding some additional `unmount()`.

PR #1859, opened 2019-11-26
Refactor of search to use options param in place of many params.

justinlittman:


jcoyne:
What is the motivation for this change?

PR #1861, opened 2019-11-26
Adds filtering to Sinopia search

justinlittman:


justinlittman:
image

justinlittman:
Just a note that I'm going to do class filterating separate from group filtering.

justinlittman:
image

src/components/search/TypeFilter.jsx
```diff
@@ -0,0 +1,96 @@
...
+  const clearFilter = () => {
+    dispatch(fetchSinopiaSearchResults(query, { ...searchOptions, startOfRange: 0, typeFilter: undefined }))
+    setTypeFilterShowDropdown(false)
+  }
+
+  const handleOnly = (type) => {
+    setSelectedTypeFilters([type])
+  }
+
+  const [typeFilterShowDropdown, setTypeFilterShowDropdown] = useState(false)
```
jermnelson: Not a blocker but should these state functions be moved to line 11 to make it more consistent with other components?

PR #1865, opened 2019-11-26
Change link list to dropdown list

jgreben:
Changes link list to dropdown list. 
 
dropdown

### DIFF
--- a/src/components/editor/property/ResourceList.jsx
+++ b/src/components/editor/property/ResourceList.jsx
@@ -3,7 +3,6 @@
 import React, { useEffect, useRef, useState } from 'react'
 import { useDispatch, useSelector } from 'react-redux'
 import PropTypes from 'prop-types'
-import { Link } from 'react-router-dom'
 import { getTemplateSearchResults } from 'sinopiaSearch'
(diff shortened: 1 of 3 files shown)


justinlittman:
The select is really ugly in my browser (Firefox):
image

justinlittman:
I know I said to use a select, but I changed this to a dropdown.

image

src/components/editor/property/ResourceList.jsx
```diff
@@ -33,13 +32,16 @@ const ResourceList = (props) => {
...
-            listItems.push(<li key={shortid.generate()}>
-              <Link to={{ pathname: '/editor', state: { } }}
-                    onClick={e => handleClick(hit.id, e)}>
-                {hit.resourceLabel}
-              </Link>
-              <span className="left-space">({hit.resourceURI})</span>
-            </li>)
+            listItems.push(
+              <button className="dropdown-item" href="#" data-resource-id={hit.id} key={shortid.generate()} onClick={async (e) => {
+                e.persist()
```
justinlittman: You should be able to avoid all use of the event.

PR #1868, opened 2019-11-26
Persist RDF to Trellis using Turtle

### LINKED ISSUE #1753: Reconsider how the editor loads data into Trellis
While working on [https://github.com/LD4P/sinopia_exporter/issues/53](https://github.com/LD4P/sinopia_exporter/issues/53)&mdash;a ticket to add subject URIs to JSON-LD representations of resources exported from Trellis&mdash;I learned that these subject URIs are blank in Trellis because of how the editor is loading data. Namely, the N-Triples specification does not allow for relative (or empty) subject URIs and that is what we currently send. 

It is also true that when we load new data from the editor, we do not yet have a subject URI. Thus, to resolve this, we should consider sending RDF data from the editor to Trellis using a serialization of something other than N-Triples.

This also means all of the data we've persisted thus far in every environment is non-conformant and may ultimately need to be migrated.

mjgiarlo:
Fixes #1753

To be discussed/tested before merging and going with this approach:

1. With the changes in this branch, how does/will the editor handle the currently broken triples coming back from Trellis? (I *think* they will load since we have code in place for handling null subjects in N-Triples data, but I could be mistaken.)
2. Assuming all is copacetic with (1) above, will the code in this branch automagically handle migration of data that needs fixing, or do we need a more intentional and systematic migration strategy for all the broken data?
3. When this is deployed, what effect will it have on open sessions (particularly w/ regard to the changes in this branch related to checking to see if a resource has changed, which used to do this check by hashing the N-Triples representation but now does it by hashing the Turtle representation)?
4. The `sinopiaServer` module is currently sending data as Turtle and receiving it is N-Triples. Do we want to change this behavior, or should we keep retrieving data from Trellis as N-Triples? (Because ... reasons.)

Big thanks to @justinlittman for sharing the writable stream technique; that was the key to convincing the N3Writer to operate in synchronous mode (which wasn't obvious to me from the N3 documentation, which is otherwise quite good).

### DIFF
--- a/__tests__/actionCreators/resources.test.js
+++ b/__tests__/actionCreators/resources.test.js
@@ -48,7 +48,7 @@ describe('update', () => {
     await store.dispatch(update('jkl012', currentUser, 'testerrorkey'))
 
     expect(store.getActions()).toEqual([
-      { type: 'SAVE_RESOURCE_FINISHED', payload: { resourceKey: 'jkl012', checksum: '5e30bd59d0186c5307065436240ba108' } },
+      { type: 'SAVE_RESOURCE_FINISHED', payload: { resourceKey: 'jkl012', checksum: 'e37c563187b12275acf955128f14f3f3' } },
     ])
   })
   it('dispatches actions when error occurs', async () => {
@@ -128,7 +128,7 @@ describe('retrieveResource', () => {
         { type: 'TOGGLE_COLLAPSE', payload: { reduxPath } },
         { type: 'RESOURCE_LOADED', payload: { resourceKey: 'abc123', resource: expectedResource, resourceTemplates: { [resourceTemplateId]: resourceTemplate } } },
         { type: 'SET_LAST_SAVE_CHECKSUM', payload: { resourceKey: 'abc123', checksum: undefined } },
-        { type: 'SET_LAST_SAVE_CHECKSUM', payload: { resourceKey: 'abc123', checksum: 'a4c091070fd59aeed47e608ad2194092' } },
+        { type: 'SET_LAST_SAVE_CHECKSUM', payload: { resourceKey: 'abc123', checksum: '547880ad2b8b81bfac4d15e0f856f248' } },
         { type: 'SET_UNUSED_RDF', payload: { resourceKey: 'abc123', rdf: '' } },
         { type: 'SET_CURRENT_RESOURCE', payload: 'abc123' },
       ])
@@ -236,7 +236,7 @@ describe('publishResource', () => {
     await store.dispatch(publishResource('jkl012', currentUser, group, 'testerrorkey'))
     expect(store.getActions()).toEqual([
       { type: 'SET_BASE_URL', payload: { resourceKey: 'jkl012', resourceURI: 'http://sinopia.io/repository/myGroup/myResource' } },
-      { type: 'SAVE_RESOURCE_FINISHED', payload: { resourceKey: 'jkl012', checksum: '5e30bd59d0186c5307065436240ba108' } },
+      { type: 'SAVE_RESOURCE_FINISHED', payload: { resourceKey: 'jkl012', checksum: 'e37c563187b12275acf955128f14f3f3' } },
     ])
   })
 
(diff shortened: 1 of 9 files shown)


__tests__/actionCreators/resources.test.js
```diff
@@ -48,7 +48,7 @@ describe('update', () => {
     await store.dispatch(update('jkl012', currentUser, 'testerrorkey'))
 
     expect(store.getActions()).toEqual([
-      { type: 'SAVE_RESOURCE_FINISHED', payload: { resourceKey: 'jkl012', checksum: '5e30bd59d0186c5307065436240ba108' } },
+      { type: 'SAVE_RESOURCE_FINISHED', payload: { resourceKey: 'jkl012', checksum: 'e37c563187b12275acf955128f14f3f3' } },
```
mjgiarlo: The checksum values change because N-Triples and Turtle representations are different.

src/GraphBuilder.js
```diff
@@ -35,6 +36,22 @@ export default class GraphBuilder {
     return this.dataset
   }
 
+  /**
+   * @return {string} a Turtle representation of a graph
+   */
+  async toTurtle() {
```
mjgiarlo: If anyone can find a way to take a graph and serialize it as Turtle or JSON-LD *without* needing to add promises up the stack... I'm all ears. I've had no luck, perusing the `jsonld` library, the `rdf-ext` libraries, and the `rdfjs` libraries, and random others I've dug up via searching around desperately.

src/actionCreators/resources.js
```diff
@@ -22,10 +22,10 @@ import ResourceStateBuilder from 'ResourceStateBuilder'
 import _ from 'lodash'
 
 // A thunk that updates (saves) an existing resource in Trellis
-export const update = (resourceKey, currentUser, errorKey) => (dispatch, getState) => {
+export const update = (resourceKey, currentUser, errorKey) => async (dispatch, getState) => {
```
mjgiarlo: This may or may not work, and may have side effects because I didn't bother to chase the promises up any farther in the stack.

src/actionCreators/resources.js
```diff
@@ -78,10 +78,10 @@ export const retrieveResource = (currentUser, uri, errorKey, asNewResource) => (
 }
 
 // A thunk that publishes (saves) a new resource in Trellis
-export const publishResource = (resourceKey, currentUser, group, errorKey) => (dispatch, getState) => {
+export const publishResource = (resourceKey, currentUser, group, errorKey) => async (dispatch, getState) => {
```
mjgiarlo: This may or may not work, and may have side effects because I didn't bother to chase the promises up any farther in the stack.

src/components/editor/actions/CloseButton.jsx
```diff
@@ -14,7 +14,7 @@ const CloseButton = (props) => {
   const history = useHistory()
 
   const resourceKey = useSelector(state => currentResourceKey(state))
-  const resourceHasChanged = useSelector(state => resourceHasChangesSinceLastSave(state))
+  const resourceHasChanged = useSelector(async state => await resourceHasChangesSinceLastSave(state))
```
mjgiarlo: Looking at the related test failure, this is likely not doing what I had hoped and I'm not inclined to go down this rabbit hole.

src/components/editor/actions/SaveAndPublishButton.jsx
```diff
@@ -27,7 +27,7 @@ const SaveAndPublishButton = (props) => {
   const showValidationErrors = () => dispatch(showValidationErrorsAction(resourceKey))
   const hideValidationErrors = () => dispatch(hideValidationErrorsAction(resourceKey))
 
-  const resourceHasChanged = useSelector(state => resourceHasChangesSinceLastSave(state))
+  const resourceHasChanged = useSelector(async state => await resourceHasChangesSinceLastSave(state))
```
mjgiarlo: Looking at the related test failure, this is likely not doing what I had hoped and I'm not inclined to go down this rabbit hole.

jcoyne:
Please see https://github.com/LD4P/sinopia_editor/issues/1753#issuecomment-560488682

justinlittman:
Need to make sure that Sinopia Export can handle this change.

justinlittman:
Also, looks like this breaks indexing pipeline.

justinlittman:
@mjgiarlo I added a fix that allows loading with or without relative URI. This should allow loading existing resources.

mjgiarlo:
>
> Need to make sure that Sinopia Export can handle this change.

I *believe* we may have accounted for this already: https://github.com/LD4P/sinopia_exporter/pull/70/files#diff-dbc2ddfc6efc5c676aa000e8abc942baR78-R80

PR #1871, opened 2019-11-27
Handle when no QA results without erroring.

### LINKED ISSUE #1860: using asterisk or quotation mark in search term for QA search kicks you to blank screen
Search tab

choose a Share-VDE source

enter a search term with an asterisk like `his*` or with quotation marks like `"index to the"` and hit enter

you get a blank screen

I don't know if QA is supposed to support * and "", if not, we should show a message that says so, or ignore those chars

justinlittman:
closes #1860


justinlittman:
Very true.

PR #1876, opened 2019-11-28
Paginate qa

justinlittman:


justinlittman:
Please review, but do not merge.

For testing, you'll want to change `https://lookup.ld4l.org/authorities` in `apidoc.json` to `https://lookup-stg.ld4l.org/authorities`.

justinlittman:
The necessary QA changes are now live, so this can be tested.

PR #1896, opened 2019-12-02
Widens rt choice modal.

### LINKED ISSUE #1875: Unable to see full resource template name when copying SHARE-VDE into Sinopia
When importing a SHARE resource and choosing a template, it would be nice if we could either expand the window to see the full template name, or have it show up when I hover over it.

**Give an example**
image

Otherwise we'll have to make sure our template names are much shorter.

justinlittman:
closes #1875


justinlittman:
image

PR #1897, opened 2019-12-02
Switched sinopia sort to a right-alinged dropdown.

### LINKED ISSUE #1870: Sinopia sort drops off side of screen
image

Please change it's drop direction (see BS docs).

justinlittman:
closes #1870


justinlittman:
image

PR #1937, opened 2019-12-04
Removed radio button scaling.

justinlittman:


jmartin-sul:
explanation for the indecision on this:  they looked weird in chrome and safari (but fine in firefox)

PR #1946, opened 2019-12-04
Adds URI to Sinopia and QA results.

### LINKED ISSUE #1941: Include resource URI in search results
Note:
* This is for both Sinopia search and QA search results.
* The column header should be changed.
* The URI should be added below the label when the label is not the URI (i.e., both are the same).

This is similar to resource template results:
image

justinlittman:
closes #1941


justinlittman:
image

justinlittman:
image

PR #1952, opened 2019-12-05
Do not trigger blur when click diacritic keyboard button.

justinlittman:


jermnelson:
The existing onBlur functionality that adds the item when leaving the `InputLiteral` does not seem to work for me. Is that intentional?

PR #1957, opened 2019-12-06
Adds HB configuration for source maps.

justinlittman:


justinlittman:
Taking the liberty of merging myself as now way to test without merge.

PR #1966, opened 2019-12-06
Fix InputLookQA for new QA result format.

### LINKED ISSUE #1965: QA lookup on DEV kicks to screen with "An Error Occurred"
in DEV, on fields that are lookups to QA sources, after I type a string to start the lookup, Sinopia kicks me to a screen that just says "An Error Occurred"

Tried with different templates and different browsers, same thing happens.

example to reproduce
https://development.sinopia.io/editor/Sinopia:RT:Demo:SampleResourceFuller
go to field "Example 9"
type something

justinlittman:
closes #1965

### DIFF
--- a/__tests__/components/editor/property/InputLookupQA.test.js
+++ b/__tests__/components/editor/property/InputLookupQA.test.js
@@ -1,0 +1,172 @@
+// Copyright 2019 Stanford University see LICENSE for license
+
+import React from 'react'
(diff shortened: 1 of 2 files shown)


jcoyne:
@justinlittman does this mean we  have to synchronize our deploys with when QA is deployed to stage & prod?

justinlittman:
Yes, we currently do that.

PR #1977, opened 2019-12-09
Makes language selection buttons consistent with other modals.

### LINKED ISSUE #1976: buttons on dialog to change literal language are not styled
Screen Shot 2019-12-09 at 8 41 12 AM

buttons don't look like buttons

justinlittman:
closes #1976


justinlittman:
image

PR #2066, opened 2020-01-29
Close Button sets resourceKey based on props or currentResource 

### LINKED ISSUE #2052: Incorrect message "Resource has unsaved changes. Are you sure you want to close?" Still happening with new wrinkle
I am still getting the message about Resource has unsaved changes, etc.  It no longer gives the message if you make no changes, but if you make a change and save, it still tells you you have unsaved changes.  In addition, if you go ahead and close, it closes the record in the tab farthest to the right instead of whatever tab you are looking at.  For example, search for "crossword" and open first the work, then the two instances.  Add a change to each (like add or delete a note or something) and save each.  Then click the tab for the work so the work record is on the screen.  Try to close the record and you will get the "unsaved changes" note.  Click close on the note popup and one of the other tabs will close, leaving the work tab.    This new bug has caused me to accidentally delete a record that actually had unsaved changes because i thought i was closing the record for which i had saved changes.

jermnelson:
Think this Fixes #2052 (but needs additional testing to confirm)

### DIFF
--- a/src/components/editor/actions/CloseButton.jsx
+++ b/src/components/editor/actions/CloseButton.jsx
@@ -13,8 +13,9 @@ const CloseButton = (props) => {
   const dispatch = useDispatch()
   const history = useHistory()
 
-  const resourceKey = useSelector((state) => currentResourceKey(state))
-  const resourceHasChanged = useSelector((state) => resourceHasChangesSinceLastSave(state))
+  let resourceKey = useSelector((state) => currentResourceKey(state))
+  if (props.resourceKey) { resourceKey = props.resourceKey }
+  const resourceHasChanged = useSelector((state) => resourceHasChangesSinceLastSave(state, resourceKey))
 
   const handleClick = (event) => {
     if (resourceHasChanged) {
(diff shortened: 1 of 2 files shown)


src/components/editor/actions/CloseButton.jsx
```diff
@@ -13,8 +13,9 @@ const CloseButton = (props) => {
   const dispatch = useDispatch()
   const history = useHistory()
 
-  const resourceKey = useSelector((state) => currentResourceKey(state))
-  const resourceHasChanged = useSelector((state) => resourceHasChangesSinceLastSave(state))
+  let resourceKey = useSelector((state) => currentResourceKey(state))
+  if (props.resourceKey) { resourceKey = props.resourceKey }
```
mjgiarlo: How would you feel about reworking this as follows:

```javascript
const resourceKey = props.resourceKey ? props.resourceKey : useSelector((state) => currentResourceKey(state))
```

jermnelson: Originally I was doing something similar but got an `eslint` error that `useSelector` needed to be a deterministic function call.

jermnelson:
> I ran a couple quick manual tests on my laptop and it seems to work for me (at least I could not reproduce the case of the bug that we saw before). I am not sure why this code would fix it though, at least in my reading it seems functionality equivalent to what is currently there...or am I missing something?

The problem was that when the *x* close button was being rendered, the `props.resourceKey` was getting overridden by the `useSelector` call to state for the resourceKey value.

src/components/editor/actions/CloseButton.jsx
mjgiarlo: OK, sounds good!

PR #2070, opened 2020-02-05
bump version to 2.0.5

jmartin-sul:
was unable to push to master per release issue template instructions, creating PR for version bump.

~~discussed w/ @jermnelson, who suggested just approving and merging myself, since i'd have pushed directly to master.~~
actually i am not allowed to self-approve (makes sense).


jmartin-sul:
spawned by #2069

PR #2086, opened 2020-02-24
add CERL, Ligatus, and subauths for MeSH

elrayle:
Add new authorities for CERL and Ligatus.

Update MeSH to include subauths.  There is not full search of MeSH, so it defaults to search the subject subauth.  This prevents breaking of any existing resource templates.

### DIFF
--- a/static/authorityConfig.json
+++ b/static/authorityConfig.json
@@ -25,6 +25,38 @@
     "component": "lookup"
   },
   {
+    "label": "CERL (QA)",
+    "uri": "urn:ld4p:qa:cerl",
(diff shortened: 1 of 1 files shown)


static/authorityConfig.json
```diff
@@ -24,6 +24,30 @@
     "language": "en",
     "component": "lookup"
   },
+  {
+    "label": "CERL_person (QA)",
```
justinlittman: Do you want the _ in the labels for CERL?

elrayle: Fixed this.  And fixed it for geonames too.

PR #2124, opened 2020-04-01
[WIP] relabel OCLC FAST authorities/subauthorities to clearly mark them as cache vs. direct

elrayle:
**WIP - Work In Progress -- DO NOT MERGE**

---

See details of exploration on how to proceed in issue #2125

### DIFF
--- a/static/authorityConfig.json
+++ b/static/authorityConfig.json
@@ -603,16 +603,16 @@
     "language": "en",
     "component": "lookup"
   },
+//  {
+//    "label": "OCLCFAST cache (QA)",
+//    "uri": "urn:ld4p:qa:fast",
+//    "authority": "oclcfast_ld4l_cache",
+//    "subauthority": "",
+//    "language": "en",
+//    "component": "lookup"
+//  },
   {
-    "label": "OCLCFAST (QA)",
-    "uri": "urn:ld4p:qa:fast",
-    "authority": "oclcfast_ld4l_cache",
-    "subauthority": "",
-    "language": "en",
-    "component": "lookup"
-  },
-  {
-    "label": "OCLCFAST concept (QA)",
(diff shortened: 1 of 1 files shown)


elrayle:
From discussion in meeting with QA/Sinopia team, @jermnelson and @justinlittman gave the following response to the resource template question.

### What happens to resource templates that reference the existing naming scheme?

* Labels can change.  The URI should stay the same.  The label is displayed to the resource template editor and users, so they will see something different which might be confusing.  But the template will continue to work properly as long as the URI doesn't change.
* Deleting an auth/subauth is not a good idea as it will break resource templates that use it.  If it is recommended that an auth/subauth pair no longer be used, then the label should include `DEPRECATED` at the end of the label.  This will give resource template editors an opportunity to remove the auth/subauth from use.

elrayle:
From discussion in meeting with QA/Sinopia team...

### Should the list of subauthorities match?

Short answer: They should as much as possible.

ACTION: @sfolsom will look at the list of potential SRU indices for direct search and compare them to the entities (i.e. subauths in QA) for the cache and make recommendations for...

* which direct SRU indices matches up to which cache entity the result of which becomes the list of potential subauths
* which subauths should be available for use in resource templates
* recommended label to show the user for each subauth (This will be the same for both direct access and cache access)

jermnelson:
@elrayle as this PR is over a year old; can we close this?

michelleif:
@elrayle what is status of this work? does #2125 cover the needed work on Sinopia side?

mjgiarlo:
@elrayle hi! are you thinking you'll be working on this again or might this no longer be necessary?

elrayle:
The work begun in this PR was addressed in PR #2127 and PR #3353.

PR #2293, opened 2020-07-15
Removed building graph to see if resource has changed.

justinlittman:
This changes the approach for determining if a resource has changed so that it does not rely on the GraphBuilder.

### DIFF
--- a/__tests__/selectors/resources.test.js
+++ b/__tests__/selectors/resources.test.js
@@ -80,22 +80,14 @@ describe('selectFullSubject()', () => {
 })
 
 describe('resourceHasChangesSinceLastSave', () => {
-  it('returns currentResource if key not provided', () => {
+  it('returns changed for currentResource if key not provided', () => {
     const state = createState({ hasResourceWithNestedResource: true })
+    state.selectorReducer.entities.subjects.ljAblGiBW.changed = true
     expect(resourceHasChangesSinceLastSave(state)).toBe(true)
   })
-  it('returns true if no lastSaveChecksum', () => {
+  it('returns changed for provided resource', () => {
     const state = createState({ hasResourceWithNestedResource: true })
-    expect(resourceHasChangesSinceLastSave(state, 'ljAblGiBW')).toBe(true)
-  })
-  it('returns false if lastSaveChecksum matches', () => {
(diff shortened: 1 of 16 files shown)


__tests__/selectors/resources.test.js
```diff
@@ -80,22 +80,14 @@ describe('selectFullSubject()', () => {
...
 
 describe('resourceHasChangesSinceLastSave', () => {
-  it('returns currentResource if key not provided', () => {
+  it('returns changed for currentResource if key not provided', () => {
     const state = createState({ hasResourceWithNestedResource: true })
+    state.selectorReducer.entities.subjects.ljAblGiBW.changed = true
     expect(resourceHasChangesSinceLastSave(state)).toBe(true)
   })
-  it('returns true if no lastSaveChecksum', () => {
+  it('returns cahnged for provided resource', () => {
```
jermnelson, line 88: typo `cahnged`

src/reducers/resources.js
```diff
@@ -62,17 +62,23 @@ export const setCurrentResource = (state, action) => {
...
   const newState = { ...state }
 
+  const subject = state.entities.subjects[action.payload.key]
   const newSubject = { ...action.payload }
   delete newSubject.properties
   delete newSubject.subjectTemplate
   newState.entities.subjects[action.payload.key] = newSubject
 
+  if (!_.isEqual(newSubject, subject)) {
+    newState.entities.subjects[newSubject.resourceKey].changed = true
```
jermnelson, line 72: I think setting this value to `true` is creating a regression because when a resource template is initially loaded and nothing has changed in the state, the **Save** button (`src/components/editor/actions/SaveAndPublishButton`) is enabled when it shouldn't be. Here is a screen shot just after the abbreviated title resource template is loaded with the save button enabled (should be disabled because nothing has changed):
Screen Shot 2020-07-15 at 4 49 40 PM

__tests__/selectors/resources.test.js
justinlittman: Fixed.

src/reducers/resources.js
justinlittman: Fixed by setting `changed` to false after resource is loaded.

Issue #2286, opened 2020-07-13
Write tests for actionCreators/templates


justinlittman:
* saveNewProfileOrResourceTemplate
* saveResourceTemplates

justinlittman:
Hoping to remove the need for these.

jermnelson:
Closing with merged PR #2326

Issue #2278, opened 2020-07-10
Write feature tests for downloading a resource template


justinlittman:

justinlittman:
Hoping to move this feature out of Sinopia.

jermnelson:
No longer applicable,  closing with merged PR #2326

Issue #2264, opened 2020-07-10
Write feature tests for loading a new profile


justinlittman:
Including:
* when valid and new
* when valid and already exists
* when not valid

justinlittman:
Hoping to move this feature out of Sinopia.

jermnelson:
No longer applicable,  closing with merged PR #2326

PR #2326, opened 2020-08-04
Resource templates as resources, switch to Mongo, more.

justinlittman:


justinlittman:
If this proceeds, close:
* #2286
* #2278 
* #2265
* #2264

justinlittman:
Additional work to be done or ticketed:
* Delete `__x_*_fixtures__`
* Fix `docker-compose.yml` to use images rather than build.
* Update README.
* Write test for actionCreators/lookups file: uri.
* Write sinopiaApi test.

PR #2331, opened 2020-08-06
Feature test for Preview RDF

### LINKED ISSUE #2262: Write feature tests for previewing RDF

jermnelson:
Feature test for `Preview RDF` modal, fixes #2262

### DIFF
--- a/__tests__/feature/previewRdf.test.js
+++ b/__tests__/feature/previewRdf.test.js
@@ -1,0 +1,48 @@
+// Copyright 2020 Stanford University see LICENSE for license
+
+import { renderApp, createHistory } from 'testUtils'
+import { fireEvent, screen } from '@testing-library/react'
+import * as sinopiaServer from 'sinopiaServer'
+import { getFixtureResourceTemplate } from 'fixtureLoaderHelper'
+import Config from 'Config'
+
+jest.mock('sinopiaServer')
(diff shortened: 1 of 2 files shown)


__tests__/feature/previewRdf.test.js
```diff
@@ -0,0 +1,47 @@
...
+    await screen.findByText(/Uber template1/)
+    fireEvent.click(screen.getAllByRole('button', { name: 'Preview RDF' })[0])
+
+    // Wait for RDF Preview Modal and selects turtle Format
+    await screen.findByText(/RDF Preview/)
+    fireEvent.change(screen.getByRole('combobox', { name: 'RDF Format Selection' }),
+      { target: { value: 'turtle' } })
+
+    // Tests for presence of turtle RDF in the model
+    expect(screen.findByText(rdf)).toBeTruthy()
```
justinlittman, line 45: This returns a promise, so you either need an `await` or change to `getByText`. Also, the `expect` is not necessary, because it will throw if no match. See https://testing-library.com/docs/react-testing-library/cheatsheet#queries

jermnelson: Good catch. I just updated the PR to use `getByTestId` to extract the `<pre>` and then compares the text content with the expected RDF string. Tried using `getByText` approach but was running into problems with selecting on the escaped HTML.

PR #2338, opened 2020-08-11
Stop InputLookup from re-querying when a prop changed.

justinlittman:
## Why was this change made?
InputLooks were querying once they had a value and any property changed. This means that changing other inputs was causing a re-query.

Locally.

No.

### DIFF
--- a/src/components/editor/property/InputLookup.jsx
+++ b/src/components/editor/property/InputLookup.jsx
@@ -41,9 +41,10 @@ const InputLookup = (props) => {
     setSelectedAuthorities(allAuthorities)
   }, [allAuthorities])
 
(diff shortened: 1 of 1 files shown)


src/components/editor/property/InputLookup.jsx
```diff
@@ -43,7 +43,7 @@ const InputLookup = (props) => {
 
   useEffect(() => {
     if (!query) return
-
+    console.log('querying', query)
```
jermnelson, line 46: Cruft?

justinlittman: Yup. Removed.

jermnelson:
Approved, merge when build passes.

PR #2345, opened 2020-08-19
Adds a test for the NewsItem component

### LINKED ISSUE #2275: Write feature tests for reading news

jgreben:
Fixes #2275 

 If there is a better way to narrow the rendered listitems to just the specific component I could more precisely test the length of the list.

### DIFF
--- a/__tests__/feature/readNews.test.js
+++ b/__tests__/feature/readNews.test.js
@@ -1,0 +1,15 @@
+// Copyright 2020 Stanford University see LICENSE for license
+
+import { renderApp } from 'testUtils'
+import { screen } from '@testing-library/react'
+
+describe('reading the latest Sinopia news items', () => {
+  it('displays a div containing the latest news, version of Sinopia, link to github site, and a static text list', async () => {
+    renderApp()
+    await screen.findByText('Latest news')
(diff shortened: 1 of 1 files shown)


__tests__/feature/readNews.test.js
```diff
@@ -0,0 +1,15 @@
...
+import { screen } from '@testing-library/react'
+
+describe('reading the latest Sinopia news items', () => {
+  it('displays a div containing the latest news, version of Sinopia, link to github site, and a static text list', async () => {
+    renderApp()
+    await screen.findByText('Latest news')
+    await screen.findByText(/Sinopia Version \d\.\d\.\d highlights/)
+    screen.getByRole('link', { name: 'Sinopia help site' })
+    const list = screen.getAllByRole('listitem')
+    expect(list.length).not.toBeLessThan(4)
```
jermnelson, line 13: I think the exact number of items depends on @michelleif for each major or minor version updates. I think we will always have at least one item.

jermnelson:
> Fixes #2275
> 
> If there is a better way to narrow the rendered listitems to just the specific component I could more precisely test the length of the list.

The number of items varies depending on PO, but we should always have at least have one item.

PR #2351, opened 2020-08-21
New tests for actionCreators/resources

### LINKED ISSUE #2239: Write tests for actionCreators/resources

jermnelson:
## Why was this change made?
Fixes #2239, adds missing action to `contractProperty`.

Run test suite

n/a

### DIFF
--- a/__tests__/actionCreators/resources.test.js
+++ b/__tests__/actionCreators/resources.test.js
@@ -1,7 +1,10 @@
 import {
   newResourceFromDataset, loadResource, newResource, newResourceCopy,
-  expandProperty, addSiblingValueSubject,
+  expandProperty, addSiblingValueSubject, saveNewResource, saveResource,
+  contractProperty,
 } from 'actionCreators/resources'
+import mockConsole from 'jest-mock-console'
+import * as sinopiaApi from 'sinopiaApi'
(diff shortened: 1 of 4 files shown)


__tests__/actionCreators/resources.test.js
```diff
@@ -15,12 +16,19 @@ import _ from 'lodash'
...
   sinopiaServer.foundResourceTemplate.mockResolvedValue(true)
   sinopiaServer.getResourceTemplate.mockImplementation(getFixtureResourceTemplate)
   sinopiaServer.loadRDFResource.mockImplementation(getFixtureResource)
   let counter = 0
   shortid.generate = jest.fn().mockImplementation(() => `abc${counter++}`)
+  console.error = jest.fn()
+})
+
+afterEach(() => {
+  console.error = orgConsoleError
```
jgreben, line 31: Could this be renamed to something like 'resetConsoleErrorCache' or add a short comment about why this is done?

jermnelson: Sure, I like `resetConsoleErrorCache` as a better descriptor. I'll also add a short comment to explain that is to avoid unnecessary console errors when we're testing raising errors.

jgreben (approved):

PR #2353, opened 2020-08-21
Adds a test for property info

### LINKED ISSUE #2272: Write feature tests for getting resource/property related info
For example, tool tips.

jgreben:
## Why was this change made?
Fixes #2272. Tests the rendering of the property remark and whether it becomes a tooltip or a link.

Updates the uber_template3 fixture to use a url in one of the property remarks, and updates another test that uses it. This change should not materially effect the nature of the dependent test, but I can instead create a new fixture for this test if desired.

### DIFF
--- a/__tests__/feature/editing/propInfo.test.js
+++ b/__tests__/feature/editing/propInfo.test.js
@@ -15,11 +15,11 @@ describe('getting property related info from a resorce', () => {
     const history = createHistory(['/editor/resourceTemplate:testing:uber3'])
     renderApp(null, history)
 
-    // // if the tooltip remark is text
+    // if the tooltip remark is text
     const infoIcon3 = await screen.findByTitle('Uber template3, property1')
     expect(infoIcon3).toHaveAttribute('data-content', 'A literal')
 
-    // // if the remark us a Url
+    // if the remark us a Url
     const infoLink = await screen.findByRole('link', { name: 'http://access.rdatoolkit.org/1.0.html' })
     expect(infoLink).toHaveClass('prop-remark')
   })


__tests__/feature/editing/propInfo.test.js
```diff
@@ -0,0 +1,42 @@
...
+
+describe('getting property related info from a resorce', () => {
+  sinopiaServer.foundResourceTemplate.mockResolvedValue(true)
+  sinopiaServer.getResourceTemplate.mockImplementation(getFixtureResourceTemplate)
+
+  it('has tooltip text info or a link based on the content of a top-level property remark', async () => {
+    const history = createHistory(['/editor/resourceTemplate:testing:uber3'])
+    renderApp(null, history)
+
+    // // if the tooltip remark is text
```
jermnelson, line 18: Any reason to have two forward slashes pairs (`//`) here?

__tests__/feature/editing/propInfo.test.js
```diff
@@ -0,0 +1,42 @@
...
+    const infoIcon3 = await screen.findByTitle('Uber template3, property1')
+    expect(infoIcon3).toHaveAttribute('data-content', 'A literal')
+
+    // // if the remark us a Url
```
jermnelson, line 22: Same question above about the two forward slashes pairs (`//`)

__tests__/feature/editing/propInfo.test.js
jgreben: Just to indicate that it is a kind of nested test. I was under the impression that we still could not use ruby-styled describe- and it- blocks.

__tests__/feature/editing/propInfo.test.js
jgreben: same answer as above ;)

PR #2382, opened 2020-08-27
Adds aria alert role to LoginPanel authentication error div

### LINKED ISSUE #2376: missing ARIA alert role on login error message div
**Describe the bug**
> What went wrong? What did you expect to happen?

when a user attempts to log in with incorrect credentials, the error message `div` does not have an [ARIA alert role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Alert_Role).

> - What is the impact on your work? (can't work / work stoppage; workaround exists; minor issue)

this could hinder interaction for users who use adaptive technology like screen readers.

> - What environment are you in? (stage or production)

discovered while writing feature tests, see https://github.com/LD4P/sinopia_editor/pull/2352#discussion_r474858506

> - What OS and browser are you using? (Mac/Windows; Safari, Firefox, etc.)

have not tested with an actual screen reader, just feature tests

> - If you're cataloging, what Resource Template were you using (name or ID)

n/a

Screen Shot 2020-08-26 at 12 10 54 PM

even after the above error message is displayed, no page elements have the `alert` role.

jgreben:
## Why was this change made?
Fixes #2376 by adding the role attribute to the LoginPanel authentication error div

Modified the existing test's CSS selector to look for the div's role attribute instead of the classname

### DIFF
--- a/__tests__/components/home/LoginPanel.test.js
+++ b/__tests__/components/home/LoginPanel.test.js
@@ -81,6 +81,6 @@ describe('<LoginPanel /> when there is an authentication failure', () => {
   const wrapper = shallow(<LoginPanel.WrappedComponent authenticationError={authenticationError}/>)
 
   it('renders the failure message', () => {
-    expect(wrapper.find('.alert').text()).toMatch('borked!')
+    expect(wrapper.find('div[role="alert"]').text()).toMatch('borked!')
   })
 })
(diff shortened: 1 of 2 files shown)


jmartin-sul:
turns out i should've been using [`description`](https://www.w3.org/TR/accname-1.1/#dfn-accessible-description) instead of name, as this worked:
```js
expect(screen.getByRole('alert', { description: /incorrect username or password/i })).toBeInTheDocument()
```

small follow-on PR coming shortly...

jmartin-sul:
https://github.com/LD4P/sinopia_editor/pull/2387

Issue #2276, opened 2020-07-10
Write feature tests for searching for Sinopia resources


justinlittman:
Including:
* Filtering
* Sorting
* Paging
* When no results

jmartin-sul:
all but filtering were tested in #2487, and the filtering test now has a separate ticket (https://github.com/LD4P/sinopia_editor/issues/2499), so closing this one...

PR #2387, opened 2020-08-27
figured out a working accessibility test for login form errors

jmartin-sul:
## Why was this change made?

test that there's an accessible `alert` element on the page when a user gets an error trying to log in.

after follow up conversation and fiddling, also attempts to make the login form more accessible (specifically the error message when login fails).  see also #2376 

this just augments an existing feature test

n/a

### DIFF
--- a/__tests__/feature/userAuthentication.test.js
+++ b/__tests__/feature/userAuthentication.test.js
@@ -71,12 +71,15 @@ describe('user authentication', () => {
 
     awsCognito.CognitoUser.mockImplementation(MockCognitoUserAuthenticationFailure)
 
+    // we don't expect any alerts to be present yet...
+    expect(screen.queryByRole('alert')).not.toBeInTheDocument()
+
     // try to login
     fireEvent.change(screen.getByLabelText('User name'), { target: { value: 'baz.le.quux@example.edu' } })
     fireEvent.change(screen.getByLabelText('Password'), { target: { value: 'password1' } })
     fireEvent.click(screen.getByRole('button', { name: 'Login' }))
 
-    // an error message should be presented
-    await screen.findByText(/Incorrect username or password./)
+    // ...and an accessible alert should be present now that a failed login attempt was made
+    await screen.findByRole('alert', { name: 'Incorrect username or password.' })
   })
 })
--- a/src/components/home/LoginPanel.jsx
+++ b/src/components/home/LoginPanel.jsx
@@ -96,9 +96,18 @@ class LoginPanel extends Component {
       </React.Fragment>
     )
 
+    const loginErrorMessageElement = (
+      <React.Fragment>
+        {
+          authenticationError
+            && <div className="alert alert-danger alert-dismissible" role="alert" aria-label={ authenticationError.message }>{ authenticationError.message }</div>
+        }
+      </React.Fragment>
+    )
+
     return (
       <React.Fragment>
-        { authenticationError && <div className="alert alert-danger alert-dismissible" role="alert" aria-label={ authenticationError.message }>{ authenticationError.message }</div> }
+        { loginErrorMessageElement }
         <form className="login-form" onSubmit={this.handleLoginSubmit}>
           { !currentSession && inlineLoginForm }
         </form>
--- a/src/components/home/LoginPanel.jsx
+++ b/src/components/home/LoginPanel.jsx
@@ -98,7 +98,7 @@ class LoginPanel extends Component {
 
     return (
       <React.Fragment>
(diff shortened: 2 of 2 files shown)


jmartin-sul:
tiny follow on to #2382

jmartin-sul:
er, actually, i realized i can make this a bit more concise, will fix and un-draft shortly...

jmartin-sul:
ugh, i got worried that my test wasn't doing what i expected, and i was right, because both of these expectations passed too:
```js
await screen.findByRole('alert', { description: /incorrect username or abcd/i })
await screen.findByRole('alert', { foo: /incorrect username or abcd/i })
```

which leads me to believe that despite `description` being part of the accessibility tree:  https://developer.mozilla.org/en-US/docs/Glossary/AOM

it doesn't seem to be a param for react testing library's `*ByRole` queries:  https://testing-library.com/docs/dom-testing-library/api-queries#byrole
```js
getByRole(
  container: HTMLElement, // if you're using `screen`, then skip this argument
  role: TextMatch,
  options?: {
    exact?: boolean = true,
    hidden?: boolean = false,
    name?: TextMatch,
    normalizer?: NormalizerFn,
    selected?: boolean,
    checked?: boolean,
    pressed?: boolean,
    queryFallbacks?: boolean,
  }): HTMLElement
```

amending back to the less terse but more precise version.

jmartin-sul:
3rd time's a charm?

__tests__/feature/userAuthentication.test.js
```diff
@@ -71,12 +71,17 @@ describe('user authentication', () => {
...
+
     // try to login
     fireEvent.change(screen.getByLabelText('User name'), { target: { value: 'baz.le.quux@example.edu' } })
     fireEvent.change(screen.getByLabelText('Password'), { target: { value: 'password1' } })
     fireEvent.click(screen.getByRole('button', { name: 'Login' }))
 
     // an error message should be presented
     await screen.findByText(/Incorrect username or password./)
+    // ...and an accessible alert should be present now that a failed login attempt was made (TODO: figure out querying on alert text)
+    screen.getByRole('alert')
```
justinlittman, line 85: I think you want `screen.getByRole('alert', { name: 'ALERT TEXT' })` to query on alert text.

jmartin-sul: so i'd tried a couple variations of that, and the expectation always failed when i tried to check the text in addition to the alert role.

while working on #2276 later yesterday evening, i discovered the accessibility debugger, and i went back and used it to look into this problem.  i noticed there wasn't actually a `name` property:
Screen Shot 2020-08-28 at 2 35 35 PM

later, when debugging my WIP on #2276, i noticed the use of `aria-label` in the search form, and thought i'd try applying it to the login form's error message div.  and that seemed to work:
Screen Shot 2020-08-28 at 2 36 28 PM

i just pushed up a commit with the failing test that i was writing/you're suggesting, about to push up a commit with the fix to the login form (hopefully).

__tests__/feature/userAuthentication.test.js
```diff
@@ -71,12 +71,15 @@ describe('user authentication', () => {
...
+
     // try to login
     fireEvent.change(screen.getByLabelText('User name'), { target: { value: 'baz.le.quux@example.edu' } })
     fireEvent.change(screen.getByLabelText('Password'), { target: { value: 'password1' } })
     fireEvent.click(screen.getByRole('button', { name: 'Login' }))
 
-    // an error message should be presented
-    await screen.findByText(/Incorrect username or password./)
+    // ...and an accessible alert should be present now that a failed login attempt was made
+    await screen.findByRole('alert', { name: 'Incorrect username or password.' })
```
jmartin-sul, line 83: fails prior to the second commit in this PR:
Screen Shot 2020-08-28 at 2 42 33 PM

jmartin-sul:
grr, line length...

jmartin-sul:
3rd commit on the fourth approach is a charm, as they say.

jmartin-sul:
couldn't resist poking at this further since i thought i'd figured out how to make the login error alert more accessible.  but the end to end test seems to be failing consistently (or at least failed twice in the same way).  unexpected, since the test makes it past login, and fails on opening a resource template and subsequent steps.  will investigate further monday.

jgreben (commented):
Looks good other than to get the build and end2end cypress test to pass and to rebase from master.

jmartin-sul:
i don't now that this PR makes sense anymore, given that its primary purpose was a `findByRole` test (and [we've stopped using `ByRole`](https://github.com/LD4P/sinopia_editor/pull/2470) selectors for performance reasons).

closing...

PR #2402, opened 2020-08-31
Left nav

justinlittman:
## Why was this change made?
To support left nav.

Cypress test

NA

### DIFF
--- a/src/components/editor/property/PanelResourceNavItem.jsx
+++ b/src/components/editor/property/PanelResourceNavItem.jsx
@@ -1,0 +1,39 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { setCurrentComponent } from 'actions/index'
+import { useDispatch, useSelector } from 'react-redux'
+import { displayResourceValidations } from 'selectors/errors'
+import { selectCurrentComponentKey } from 'selectors/index'
+
+const PanelResourceNavItem = (props) => {
+  const dispatch = useDispatch()
+
+  const classNames = ['list-group-item', 'list-group-item-action']
+  const currentComponentKey = useSelector((state) => selectCurrentComponentKey(state, props.resourceKey))
+  if (props.componentKey === currentComponentKey) classNames.push('active')
+  // Check or bullet
+  const prefixMark = props.hasValue ? '\u2713' : '\u2022'
+  const displayValidations = useSelector((state) => displayResourceValidations(state))
+  if (displayValidations && props.hasError) classNames.push('text-danger')
+
+  const prefix = '\u00A0'.repeat(props.level * 4)
+
+  return (
+    <button type="button"
+            className={classNames.join(' ')}
+            onClick={() => dispatch(setCurrentComponent(props.resourceKey, props.componentKey))}>
+      {prefix}{prefixMark} {props.label}
+    </button>
+  )
+}
+
+PanelResourceNavItem.propTypes = {
+  resourceKey: PropTypes.string.isRequired,
+  componentKey: PropTypes.string.isRequired,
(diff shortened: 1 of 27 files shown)


src/components/editor/property/PanelResourceNavItem.jsx
```diff
@@ -0,0 +1,38 @@
...
+import { selectCurrentComponentKey } from 'selectors/index'
+
+const PanelResourceNavItem = (props) => {
+  const dispatch = useDispatch()
+
+  const classNames = ['list-group-item', 'list-group-item-action']
+  const currentComponentKey = useSelector((state) => selectCurrentComponentKey(state, props.resourceKey))
+  if (props.componentKey === currentComponentKey) classNames.push('active')
+  // Check or bullet
+  const prefixMark = props.hasValue ? '\u2713' : '\u2022'
```
jermnelson, line 15: Could you add a comment about what these unicode values are? (i.e.: checkmark and bullet)

justinlittman: Is L14 adequate or are you looking for something different?

src/components/editor/property/PanelResourceTopNavItem.jsx
```diff
@@ -0,0 +1,38 @@
...
+
+  const classNames = ['list-group-item', 'list-group-item-action', 'list-group-top-item']
+  const currentComponentKey = useSelector((state) => selectCurrentComponentKey(state, props.resourceKey))
+  if (props.componentKey === currentComponentKey) classNames.push('active')
+  // Check or space
+  const prefixMark = props.hasValue ? '\u2713' : '\u00A0\u00A0'
+  const displayValidations = useSelector((state) => displayResourceValidations(state))
+  if (displayValidations && props.hasError) classNames.push('text-danger')
+
+  return (<button
```
jermnelson, line 19: Should completed nested properties propagate up to top level so that the top-level reflects the status of it's  children? 

For example when I load up the **Uber template 1** resource template and  expand the *Uber Template, Property 1* panel, I see the children in the left nav as expected:

Screen Shot 2020-09-01 at 2 02 20 PM

When I fill in text for the nested  **Uber Template 2** and **Uber Template 3** the children in the left nav are checked correctly but the top-level **Uber Template 1**  is not checked:

Screen Shot 2020-09-01 at 1 45 16 PM

I am fine if you want to just ticket this matter to resolve instead of implementing in this PR.

src/styles/main.scss
```diff
@@ -396,3 +380,26 @@ ul.resources-nav-tabs button.active {
 .icon-lg {
   font-size: 20px;
 }
+
```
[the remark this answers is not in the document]
justinlittman: Yes. I didn't want to change the background colors now because they would have been the same as the right panels. Seems like a single ticket to change all background colors makes sense.

src/components/editor/property/PanelResourceTopNavItem.jsx
justinlittman: I intentionally implemented the simplest possible way (no bubble up), figuring that if users felt this was not optimal then could ticket.

src/styles/main.scss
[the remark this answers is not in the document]
jermnelson: Done, ticket https://github.com/LD4P/sinopia_editor/issues/2408.

src/components/editor/property/PanelResourceTopNavItem.jsx
jermnelson: Just confirmed with @michelleif that we should expect the status to bubble up. Here is the LD4P/sinopia_editor/issues/2409

PR #2410, opened 2020-09-01
Do not restart mongo resources

aaron-collier:
## Why was this change made?

This may be something only I'm seeing, or there may be a better approach, but the `restart: always` configs here mean that the mongo resources are configured to always start when docker-desktop starts. This is not ideal for local resources where we don't want an extraneous service running when we're not working on this project.

In order to remove the restart directive, `docker-compose up -d` needs to run with this config (including the `restart: "no"`, we can likely eventually just remove the restart directive all together, but if you've already run the current config, you need to force it to get the new no-restart.

If we need the restart directive, perhaps `restart: unless-stopped` would be more appropriate so we can actually stop the thing.

Locally

N/A

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -13,7 +13,6 @@ services:
       - api
   pipeline:
     image: ld4p/sinopia_indexing_pipeline:latest
-    # Replace image with build to build from local code. Do not commit docker-compose.yml.
     # build:
     #   context: ../sinopia_indexing_pipeline
     environment:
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -4,7 +4,7 @@ services:
     build:
       context: .
       args:
-        - USE_FIXTURES='false'
+        - USE_FIXTURES='true'
```
justinlittman, line 7: Is there a reason that you are switching this?

docker-compose.yml
```diff
@@ -44,7 +44,7 @@ services:
       - elasticsearch
   mongo:
     image: mongo:4.4
-    restart: always
+    restart: "no"
```
justinlittman, line 47: Default is "no", so you can just remove all restart: "no" lines.

docker-compose.yml
aaron-collier: Cruft. Did not mean to commit that. I'll revert.

docker-compose.yml
aaron-collier: Done.

aaron-collier:
rebasing nightmare from changes since opening this PR... will open a new PR

PR #2413, opened 2020-09-02
allow for recent templates to be editable with button

### LINKED ISSUE #2362: (DEV): Error when trying to edit a template from "Most recently used list"
1) open any template to create a Resource (i.e., click the name of a template on the Resource Templates tab)
2) go back to Resource Templates tab, and the template you used in step 1 will now be in the “Most recently used resource templates” menu. Attempt to edit the **template itself** by clicking the pencil icon
3) ```Error retrieving undefined: Error parsing resource: JSON.parse: unexpected character at line 1 column 1 of the JSON data```

peetucket:
## Why was this change made?

To fix #2362 

Although the same component is being used to render resource template rows in both the recent and full list, the data structure passed to the component is not the same, and in the case of the recent templates, it didn't have the expected full template URI.  This adds it.

`props.row.uri` was undefined in the recent templates structure here: https://github.com/LD4P/sinopia_editor/blob/master/src/components/templates/ResourceTemplateRow.jsx#L37

localhost development environment

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,6 +193,7 @@ The following are only in the resource subject (that is, the base subject).
 {
   key: <resource template id, e.g., resourceTemplate:bf2:Monograph:Instance>,
   id: <resource template id, e.g., resourceTemplate:bf2:Monograph:Instance>,
+  uri: <resource template uri, e.g. http://datastore/repository/resourceTemplate:bf2:Monograph:Instance>,
   class: <resource URI, e.g., http://id.loc.gov/ontologies/bibframe/Instance>,
   label: <resource label, e.g., "BIBFRAME Instance">,
   author: <author>,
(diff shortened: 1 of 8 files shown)


justinlittman (changes requested):
This works, but I would prefer to see it added to the subject model in TemplateBuilder, around https://github.com/LD4P/sinopia_editor/blob/master/src/TemplatesBuilder.js#L24. (I believe this is the URI, which is passed into the TemplateBuilder.)

peetucket:
> This works, but I would prefer to see it added to the subject model in TemplateBuilder, around [`master`/src/TemplatesBuilder.js#L24](https://github.com/LD4P/sinopia_editor/blob/master/src/TemplatesBuilder.js#L24). (I believe this is the URI, which is passed into the TemplateBuilder.)

ok, fixed and squashed commits

justinlittman (approved):

PR #2418, opened 2020-09-03
Move lookups into modals per new design

### LINKED ISSUE #2397: Lookup Modal Stub
Create and style modal for QA/Lookup with close button in upper-right corner:

image

Blocks #2398

aaron-collier:
## Why was this change made?

Fixes #2397 by adding a stub modal window for lookups.

Marking this ready, though I think some styling should be done (perhaps in a follow up) but I do not currently have the style information for the new look.

Existing tests


aaron-collier:
@jermnelson / @justinlittman I think this is close but could use guidance on what might be the best way (if we care) to not display the selected value of a lookup before the modal is closed.

Also - the new styling will need to be applied, but otherwise this is now showing on the template page.

jermnelson:
> @jermnelson / @justinlittman I think this is close but could use guidance on what might be the best way (if we care) to not display the selected value of a lookup before the modal is closed.
> 

After testing this locally, I think it is fine for now to display the selected value before the modal is closed (if users have an issue we can always ticket and fix later). I did notice some weirdness when testing with the Uber template 1 fixture (`resourceTemplate:testing:uber1`). Clicking the button for *Uber template1, property10* opens up two modals and shifts to the *Uber template1, property17*, any values are displayed in that field and you're unable to enter any values for the original *Uber template1, property10*.

Screen Shot 2020-09-15 at 8 23 30 AM

aaron-collier:
@jermnelson confirm I see this locally... looking.

src/components/editor/property/InputListLOC.jsx
```diff
@@ -130,41 +136,123 @@ const InputListLOC = (props) => {
...
+        <span aria-hidden="true"><FontAwesomeIcon className="trash-icon" icon={faTrashAlt} /></span>
+        </button>
+        <a href={lookupValue.uri}>
+          <span aria-hidden="true"><FontAwesomeIcon className="globe-icon" icon={faGlobe} /></span>
+        </a>
+      </div>
+    )
+  })
+
+  const modal = (
```
jermnelson, line 182: Is there a way to create a Lookup modal component that both `InputListLOC` and `InputLookup` could instead of defining a local modal for each of these components?"

src/components/editor/property/InputLookup.jsx
```diff
@@ -142,28 +148,98 @@ const InputLookup = (props) => {
...
+        <span aria-hidden="true"><FontAwesomeIcon className="trash" icon={faTrashAlt} /></span>
+        </button>        
+        <a href={lookupValue.uri}>
+          <span aria-hidden="true"><FontAwesomeIcon className="globe-icon" icon={faGlobe} /></span>
+        </a>
+      </div>
+    )
+  })
+
+  const modal = (
```
jermnelson, line 197: See previous comment about possible extracting this modal into a separate component.

jermnelson (changes requested):
Question revolving around possible DRY to a general modal component.

src/components/editor/property/InputListLOC.jsx
aaron-collier: Looking into that now. Not sure as they are slightly different, but those differences may be inconsequential.

jermnelson (approved):
Thanks @aaron-collier for all of your efforts on this PR. Merge when eslint passes.

PR #2422, opened 2020-09-03
Auto-resize input literals to reduce scrolling

### LINKED ISSUE #2177: Text fields should be expandable and expand to fit length of text
1. Some text fields are too small to verify they are entered correctly. All text fields should have an option to expand the field so user can see entry in its entirety.
2. Also, if text entered is really long and exceeds the amount of space given in the text field, text field should expand to fit text once it is entered.

mjgiarlo:
Fixes #2177

## Why was this change made?

Note that a UI change here is switching from an input element to a textarea element for all input literals. Why? Because as far as I can tell, input elements are not auto-resizable in HTML.

I briefly looked into rolling my own auto-resize logic but it is over my head, even without all the React machinery in the way. I also looked into using contentEditable but I believe that approach is geared towards divs and spans rather than input elements (such as textarea) within a form.

I'm not sure if the small test change is sufficient to test this commit. Advice welcome! I suspect that the line I removed from the input literal test, which began failing, is due to moving from an input element to a textarea element.

Also includes:

* Clarify in README how to start Express server

In browser. See screencaps of adding a new resizable element and editing an existing one:

Peek 2020-09-03 11-45

Peek 2020-09-03 11-46

README.

### DIFF
--- a/__tests__/feature/editing/editingLiteral.test.js
+++ b/__tests__/feature/editing/editingLiteral.test.js
@@ -41,7 +41,7 @@ describe('editing a literal property', () => {
     fireEvent.click(editBtn)
     expect(input).not.toBeDisabled()
     expect(input).toHaveValue('foo')
-    expect(screen.queryAllByText('foo').length).toBeFalsy()
+    expect(screen.queryAllByText('foo').length).toBeLessThan(2)
 
     // Clicking remove
     fireEvent.change(input, { target: { value: 'foobar' } })
--- a/cypress/integration/leftNav.spec.js
+++ b/cypress/integration/leftNav.spec.js
@@ -69,7 +69,7 @@ describe('Left-nav test', () => {
   })
 
   it('Marks properties with values with a check', () => {
-    cy.get('input[placeholder="Uber template1, property4"]')
+    cy.get('textarea[placeholder="Uber template1, property4"]')
       .type('foo{enter}')
     cy.get('button.active > h5')
       .should('contain', 'Uber template1, property4')
@@ -86,7 +86,7 @@ describe('Left-nav test', () => {
     cy.contains('button', 'Item of').click()
     // Wait for scroll
     // eslint-disable-next-line cypress/no-unnecessary-waiting
-    cy.wait(1000)
+    cy.wait(1500)
     cy.isInViewport('div[data-label="Item of"]')
     cy.get('button.active > h5').should('contain', 'Item of')
   })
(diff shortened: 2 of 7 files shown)


__tests__/feature/editing/editingLiteral.test.js
```diff
@@ -41,7 +41,6 @@ describe('editing a literal property', () => {
     fireEvent.click(editBtn)
     expect(input).not.toBeDisabled()
     expect(input).toHaveValue('foo')
-    expect(screen.queryAllByText('foo').length).toBeFalsy()
```
mjgiarlo, line 44: Removed, apparently because this passes with `<input>`-based literal fields but fails with `<textarea>`-based ones.

cypress/integration/leftNav.spec.js
```diff
@@ -86,7 +86,7 @@ describe('Left-nav test', () => {
     cy.contains('button', 'Item of').click()
     // Wait for scroll
     // eslint-disable-next-line cypress/no-unnecessary-waiting
-    cy.wait(1000)
+    cy.wait(1500)
```
mjgiarlo, line 89: Sorry to have to increase this, but I couldn't get it to pass locally with just `1000`. :unamused:

jgreben (approved):

__tests__/feature/editing/editingLiteral.test.js
jgreben: Can you change it to: `expect(screen.queryAllByText('foo').length).toBeLessThan(2)`?

jgreben: Apparently the `textarea` is also counted as a text value. This is supposed to test that the little rbt-token text of the input value has dissappeared.

jgreben (changes requested):
What do you think about using `expect(screen.queryAllByText('foo').length).toBeLessThan(2)` or something similar?

__tests__/feature/editing/editingLiteral.test.js
mjgiarlo: @jgreben :speech_balloon: 
> Apparently the `textarea` is also counted as a text value. This is supposed to test that the little rbt-token text of the input value has dissappeared.

Done! Thanks for the suggestion.

mjgiarlo:
@jgreben :speech_balloon: 
> What do you think about using `expect(screen.queryAllByText('foo').length).toBeLessThan(2)` or something similar?

Done. Have another :eye: ?

jgreben (approved):

PR #2428, opened 2020-09-04
Use err when no err.message

justinlittman:
## Why was this change made?
Sometimes err.message is undefined, in which case use error.

I don't know why.

NA

NA


jermnelson (approved):
Weird that error can vary.

PR #2434, opened 2020-09-08
Upgrade jest to v26

jcoyne:
## Why was this change made?

The old version of jest depended on a version of minimist which had an advisory.


justinlittman:
I rebased off master.

Issue #2403, opened 2020-08-31
SPIKE: Investigate workflow for sending Bibframe entities URI to rdf2marc and download MARC21


jermnelson:
Investigate options (at most 1/2 day) for sending a BIBFRAME entity URI (Work or Instances) to  the `rdf2marc` service to generate one or more MARC records for download. For WC3 all we need is a mechanism for running this conversion with a button that is enabled for BF Works or BF Instances. This could be as simple as a command-line script that takes a CSV file of these URIs.

The outcome for this research should be one or more tickets outlining an approach for accomplishing this work.

justinlittman:
@jermnelson @michelleif This is deployed to Dev. OK to close this ticket?

jermnelson:
Thanks @justinlittman for all of your work on this ticket! Closing now,

PR #2444, opened 2020-09-09
Marc

justinlittman:
## Why was this change made?
Allow user to request conversion to MARC.

refs #2403.

Unit test. Deliberately not testing MarcButton (and related feature), as this is a temporary approach to exposing functionality in UI.

NA


justinlittman:
image

image

jcoyne (approved):
Nice! I'm glad the MARC transform is happening via the API and not here in the editor.

PR #2445, opened 2020-09-09
Expand current property navigation item only

### LINKED ISSUE #2419: Expand/Collapse Left Nav Depending on Active Panel
Currently the left navigation in the editor displays the top-level property and all of it's sub-properties and nested resource templates like this example:
Screen Shot 2020-09-03 at 10 35 59 AM

In the above screenshot, the active top-level property panel, *Title Information*, is expanded with *Instance Title*, *Variant TItle*, *Paralle Title*, and *Collective Title*  being display. However, the top-level *Contributors* and *Publication, Distribution, Manufacture, and Production* properties panels are also expanded as well with their sub-properties being displayed. 

Instead we want to only display the active property-panel's sub-properties and any nested templates in the left nav and collapse all other properties, so it should look like this (even if  *Contributors* and *Publication, Distribution, Manufacture, and Production*  have values in their properties. 

Screen Shot 2020-09-03 at 10 58 24 AM

mjgiarlo:
Fixes #2419

## Why was this change made?

To fix #2419.

This commit has a few different components:

* Subject nav items should inject the resource key property into its property nav items
* Property nav items should check their resource key (see prior bullet) to determine if they are the currently selected property nav item and, if and only if they are, they should render their child navigation items
* Top-level navigation items should include an aria-label, partly for a11y purposes and partly to facilitate integration testing

CI and local in-browser testing. See demo of changes here:

Peek 2020-09-09 10-16

None

### DIFF
--- a/__tests__/feature/editing/openAndCloseResource.test.js
+++ b/__tests__/feature/editing/openAndCloseResource.test.js
@@ -14,7 +14,8 @@ describe('switching between multiple resources', () => {
     renderApp(store, history)
 
     await screen.findAllByRole('heading', { name: 'Abbreviated Title' })
-    const abbreviatedTitleTab = await screen.findByRole('button', { name: 'Abbreviated Title' })
+    // NOTE: There are two buttons with this name, so retrieve element using a selector
+    const abbreviatedTitleTab = screen.getByText('Abbreviated Title', { selector: 'button.nav-link' })
     expect(abbreviatedTitleTab).toHaveClass('active')
   })
 
@@ -47,7 +48,8 @@ describe('closing the resources', () => {
   it('removes the navigation tabs and the resources from view', async () => {
     renderApp(store, history)
 
-    const abbreviatedTitleTab = await screen.findByRole('button', { name: 'Abbreviated Title' })
+    // NOTE: There are two buttons with this name, so retrieve element using a selector
+    const abbreviatedTitleTab = screen.getByText('Abbreviated Title', { selector: 'button.nav-link' })
     const noteTab = await screen.findByRole('button', { name: 'Note' })
 
     // Closing the active tab will reveal the inactive resource as the one shown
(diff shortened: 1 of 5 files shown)


jermnelson (approved):
Looks good; approve once tests are green.

mjgiarlo:
@jermnelson @justinlittman I'm about to push my revised branch, which ought to fix the test suite failures. Do you think the testing strategy here (using Cypress) fits this work, or should I look to replace the integration test with a feature test? Happy to revise if so.

jermnelson:
> @jermnelson @justinlittman I'm about to push my revised branch, which ought to fix the test suite failures. Do you think the testing strategy here (using Cypress) fits this work, or should I look to replace the integration test with a feature test? Happy to revise if so.

Adding a feature test would be helpful here. I find feature tests easier to tie back to specific user-facing  code.

mjgiarlo:
> > @jermnelson @justinlittman I'm about to push my revised branch, which ought to fix the test suite failures. Do you think the testing strategy here (using Cypress) fits this work, or should I look to replace the integration test with a feature test? Happy to revise if so.
> 
> Adding a feature test would be helpful here. I find feature tests easier to tie back to specific user-facing code.

Sounds good! I'll get right on that.

mjgiarlo:
@jermnelson @justinlittman is it OK to have this feature covered by both a feature test and an integration test? or should I dump the latter? Otherwise, I expect this will be :green_heart: and ready to go momentarily.

justinlittman:
If the feature test is adequate, I would get rid of the integration test as I want to maintain as few integration tests as possible.

mjgiarlo:
@justinlittman OK, this is ready to go.

justinlittman (approved):
Approve pending green. I ran locally, and all fine so probably just a Circle issue. I'd recommend increasing the timeout, since Circle seems to need longer timeouts.

mjgiarlo:
> Approve pending green. I ran locally, and all fine so probably just a Circle issue. I'd recommend increasing the timeout, since Circle seems to need longer timeouts.

Bumped the timeout value again.

PR #2460, opened 2020-09-11
Adds re-ordering of nested resources.

### LINKED ISSUE #2358: Editing templates: need a way to reorder Property Templates
the order of the Property Templates is significant as it determines how the "form" appears to the cataloger when editing a Resource

when creating or editing a Template, there is currently no way to re-order the properties if you find you need to change the order in which you want them to appear to catalogers

need something like the old Profile Editor's ability to drag the Property Templates to re-order

justinlittman:
closes #2358

## Why was this change made?
Support re-ordering of nested resources, in particular, re-ordering of property templates when editing a resource template.

image

This can be tested when creating a new resource template or using Uber template 1 > property 19.

Unit tests, feature test, locally.

NA

### DIFF
--- a/README.md
+++ b/README.md
@@ -286,6 +286,7 @@ The following are only in the resource subject (that is, the base subject).
   label: <label for uri>,
   valueSubjectKey: <key for subject for a nested resource>,
   -> valueSubject: {<subject>}
+  -> index: <1 based index of the value (relative to siblings)>
 }
 ```
 -> Added by selector, not stored in state.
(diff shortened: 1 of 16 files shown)


README.md
```diff
@@ -286,6 +286,7 @@ The following are only in the resource subject (that is, the base subject).
   label: <label for uri>,
   valueSubjectKey: <key for subject for a nested resource>,
   -> valueSubject: {<subject>}
+  -> index: <1 based index of the value (relative to siblings)>
```
jcoyne, line 289: I thought `->` means not stored in state, but this is putting it in state, isn't it?

jcoyne (changes requested):

README.md
justinlittman: No, it is being injected by the selector.

jcoyne (approved):

PR #2470, opened 2020-09-13
Purged ByRole selectors to improve performance.

justinlittman:
## Why was this change made?
*ByRole selectors were bogging down the tests (taking ~1 sec per query rather than ms) causing timeouts.

Feature tests.

NA


mjgiarlo (approved):
Thanks for helping the build run (and pass) more consistently.

PR #2475, opened 2020-09-14
Validate after making property changes

### LINKED ISSUE #2462: when you remove a selected value from a required lookup field, you are not prevented from saving the record. Sinopia thinks the field is still full.
to reproduce use https://development.sinopia.io/editor/ld4p:RT:bf2:Monograph:Work:Un-nested
1) Content field is required, so if you enter nothing and try to save, you are warned, as expected

2) select a value from dropdown. now the "Required" message and red styling disappear, also as expected

3) now remove the value you entered

4) try to save again, and save is successful. instead, you should be warned that the field is empty

NOTE: this is a bug only with lookups, and only when you select a value from the lookup (rather than entering a "new literal")

jcoyne:
## Why was this change made?

Fixes #2462

### DIFF
--- a/src/reducers/resources.js
+++ b/src/reducers/resources.js
@@ -120,9 +120,6 @@ const addPropertyToNewState = (newState, property) => {
   newProperty.propertyTemplateKey = newProperty.propertyTemplate.key
   delete newProperty.propertyTemplate
 
-  // Errors
-  newProperty.errors = errorsForProperty(newProperty, property.propertyTemplate)
-
   // Add property to state
   const oldProperty = newState.entities.properties[newProperty.key]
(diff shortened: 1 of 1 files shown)


justinlittman (changes requested):
Testing locally this did not fix the bug.

jermnelson (approved):
This changes the behavior of the editor for required properties. The save button is disabled until all of the required items have values. I like the change but @michelleif should weigh in before merging.

michelleif:
problem is, we rely on clicking the save button to show the user what required properties still need to be filled in. if we disable save until all required props are filled, how is the user to know which are missing?

jermnelson:
Closing as #2462 no longer a bug.

Issue #2439, opened 2020-09-08
PropertyPanel Display Changes


jermnelson:
The following display changes are needed for the `PropertyPanel` CSS.  Helpful reference [design document](https://docs.google.com/document/d/12IbU5_e0cb9k1lKTOVe0HMarYYA0dG10LzIcvDyX9fo/edit#). One of the biggest changes in the new design is that background color for the `PropertyPanel` card-header will be different depending if a resource or resource template is being editing. 

## card-header 
The current `PropertyPanel`'s card-header renders with a background color of #8C8A83 with a white heading text color:
![Screen Shot 2020-09-08 at 4 02 48 PM](https://user-images.githubusercontent.com/71847/92532091-c5b87f00-f1ec-11ea-838d-67eda7d8ea5d.png) 

### card-header for RDF Resource
For resources a background color #D0C1A8 with a black heading text color. 

### card-header for Resource Template
For resource templates, the uses a background color of #B26F16 with a white heading text color. 
![image](https://user-images.githubusercontent.com/71847/92531844-52167200-f1ec-11ea-803f-a9d69238c234.png)

## border
Remove the current border color #8C8A83

## Background color
Change background-color to #F7F4F1

# Nested Property and Nested Resource Display Changes
See [Right Panel Colors](https://github.com/LD4P/sinopia_editor/wiki/Sinopia-Design-Document#right-panel-colors) for display changes.

PR #2483, opened 2020-09-15
Set the background color for the active editor menu

### LINKED ISSUE #2436: Background Color for Active Editor Nav
Currently the active property in the left navigation has a blue background color:

Screen Shot 2020-09-08 at 11 47 10 AM

In the new design, the active property should either be a solid tan #D0C1A8 with black text if editing a RDF resource or a solid orange #B26F16 with white text if editing a resource template, 
image

jcoyne:
## Why was this change made?

Fixes #2436 
Screen Shot 2020-09-15 at 1 30 12 PM

### DIFF
--- a/src/components/editor/leftNav/PanelResourceNav.jsx
+++ b/src/components/editor/leftNav/PanelResourceNav.jsx
@@ -1,14 +1,19 @@
 // Copyright 2019 Stanford University see LICENSE for license
 
 import React from 'react'
+import { useSelector } from 'react-redux'
 import PropTypes from 'prop-types'
 import PanelPropertyNav from './PanelPropertyNav'
 import ActivePanelPropertyNav from './ActivePanelPropertyNav'
 import { selectCurrentPropertyKey } from 'selectors/index'
-import { useSelector } from 'react-redux'
 
 const PanelResourceNav = (props) => {
   const currentPropertyKey = useSelector((state) => selectCurrentPropertyKey(state, props.resource?.key))
+  const isTemplate = props.resource.subjectTemplateKey === 'sinopia:template:resource'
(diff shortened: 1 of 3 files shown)


jermnelson (changes requested):
This isn't quite correct. The orange color (#B26F16 ) should only be set when editing a template, the default for editing resources should be  a solid tan #D0C1A8.

jcoyne:
@jermnelson how do we know if it's a template or a resource?  Aren't they now in the same editor?

aaron-collier:
Same question for https://github.com/LD4P/sinopia_editor/issues/2439

jermnelson:
> @jermnelson how do we know if it's a template or a resource? Aren't they now in the same editor?

I think the easiest way would be to check the URL and see if it ends with `editor/sinopia:template:resource`, @justinlittman may know of a better option.

jcoyne:
@jermnelson I don't think `findRootResourceTemplateId` is going to work because that requires you to provide a dataset (which I beleive is for importing RDF).

jermnelson (approved):

jermnelson:
@jcoyne I think this can merged after a rebase.

justinlittman:
Editing a resource template:
image

Editing a resource:
image

justinlittman:
Pinging @jermnelson for a review at your convenience.

jcoyne:
@justinlittman This needs a rebase now.  Would you be able to put the styles in https://github.com/LD4P/sinopia_editor/blob/master/src/styles/resourceNav.scss ?

justinlittman:
Done @jcoyne

jermnelson (approved):

PR #2487, opened 2020-09-16
Search feature test

### LINKED ISSUE #2276: Write feature tests for searching for Sinopia resources

jmartin-sul:
## Why was this change made?

add feature test coverage for various basic search result display behaviors.

connects with #2276

PR only adds tests

n/a

### DIFF
--- a/__tests__/feature/search.test.js
+++ b/__tests__/feature/search.test.js
@@ -1,0 +1,232 @@
+import {
+  findByText, fireEvent, screen,
+} from '@testing-library/react'
+
+import { renderApp } from 'testUtils'
+import Config from 'Config'
+
+describe('sinopia resource search', () => {
+  const fooBarHit = {
+    _index: 'sinopia_resources',
+    _type: 'sinopia',
+    _id: 'resource/cornell/34ef053e-f558-4299-a8a7-c8b79a598d99',
+    _score: 0.2876821,
+    _source: {
+      title: ['foo bar'],
+      uri: 'http://platform:8080/resource/cornell/34ef053e-f558-4299-a8a7-c8b79a598d99',
+      label: 'foo bar',
+      type: ['http://id.loc.gov/ontologies/bibframe/AbbreviatedTitle'],
+      group: 'cornell',
+      created: '2019-11-27T19:05:50.496Z',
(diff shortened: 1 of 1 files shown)


__tests__/feature/search.test.js
```diff
@@ -0,0 +1,232 @@
...
+    renderApp()
+
+    fireEvent.click(screen.getByText(/Linked Data Editor/, { selector: 'a' }))
+    fireEvent.click(screen.getByText(/Search/, { selector: 'a' }))
+    fireEvent.change(screen.getByLabelText('Query', { selector: 'input#searchInput' }), { target: { value: '*' } })
+    fireEvent.click(screen.getByLabelText('Submit search', { selector: 'button' }))
+
+    await screen.findByText(/foo bar/)
+
+    // TODO: why don't filtering options show up in test UI?
```
jmartin-sul, line 157: using `screen.debug()`, afaict, the filter elements were not rendering in the page for the tests.  i haven't figured out why, though i suspect maybe my fixture search results are incomplete?  but i didn't see anything obviously related missing from them, compared to what i saw when i ran searches on my local instance and looked at the response traffic in the browser's network debugger.

jmartin-sul (commented):
i'm inclined to file a follow-on ticket for the TODO, and to get the rest merged, assuming the rest of the testing meets the other goals outlined in #2276.

__tests__/feature/search.test.js
jermnelson: Created issue https://github.com/LD4P/sinopia_editor/issues/2499

jermnelson (approved):
Looks good. I found it interesting the use of both `screen.findByText` and the `findByText` function from the React testing library.

jmartin-sul:
> I found it interesting the use of both `screen.findByText` and the `findByText` function from the React testing library.

ah, yeah, meant to comment on that specifically:  i was doing that just because it was the only reliable way i found to search for text within a given node that'd already been retrieved, and i wanted to retrieve the nodes in first column in one go using `screen.getAllByText`, so that i could check that the UI presented things  in the desired order.  but if there's a way to do that with `screen.findByText`, or some other more idiomatic approach, i'm happy to revisit.

PR #2488, opened 2020-09-16
Insert diacritics at the cursor's current position

### LINKED ISSUE #2097: DiacriticsSelection insert value at Input's cursor location
**Describe the bug**
When user enters values using the `DiacriticsSelection` component, the value is append at the end regardless of the user's cursor location.

When user enters a quick string of characters and expects to go back and insert a character, they expect that the selection would be inserted into the cursor location. Workaround is only the characters up to the point of insertion, select character, and then continue data entry.

both
Mac/Firefox.

(Mac: shift-command-4; Windows: ; then drag the image file into the issue)
sinopia-bug-02020-03-06

jcoyne:
## Why was this change made?
Fixes #2097

### DIFF
--- a/src/components/editor/property/InputValue.jsx
+++ b/src/components/editor/property/InputValue.jsx
@@ -32,7 +32,6 @@ const InputValue = (props) => {
       </button>
     </div>
     <button
-      id="editItem"
       onClick={handleEditClick}
       style={ { marginRight: '.25em' } }
       aria-label={`Edit ${label}`}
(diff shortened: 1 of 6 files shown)


src/components/editor/property/InputValue.jsx
```diff
@@ -32,7 +32,6 @@ const InputValue = (props) => {
       </button>
     </div>
     <button
-      id="editItem"
```
jcoyne, line 35: This ID is not unique on the page.

jermnelson (approved):

Issue #2473, opened 2020-09-14
Panel expansion not working as expected


mjgiarlo:
There was recently work done to make sure to expand only one property in the panel at a time. This introduced another bug where clicking on any child properties of the currently highlighted property causes the panel navigation to disappear. See:

![Peek 2020-09-14 08-53](https://user-images.githubusercontent.com/131982/93108665-f0c92580-f667-11ea-8a08-8dc0358604b4.gif)

Issue #2447, opened 2020-09-10
when field names wrap in left nav, second line should be indented


michelleif:
when field names are long and wrap in the left nav, the second line is aligned to the left of the first line. This makes it difficult to quickly scan the checkboxes and fieldnames
see https://development.sinopia.io/editor/Sinopia:RT:Demo:SampleResourceFuller
<img width="598" alt="Screen Shot 2020-09-10 at 1 04 04 PM" src="https://user-images.githubusercontent.com/23069165/92796622-32a35480-f366-11ea-82de-2cfcc4eeefda.png">
The second line should be aligned with the first, or possible indented further, if aligning with first would cause confusion.

Another solution could be to allow user to resize the left nav if they don't want to see the wrapping. Resizing would be nice also if you want to make left nav narrower while you are cataloging?

michelleif:
@astridu could you check above and comment on how you expect longer field names to wrap in the left nav? is it currently as expected?

astridu:
I agree. When field names wrap, they should left align with the line directly above it so it's easy to scan.

michelleif:
now there are bullets down the left side, which is nice for indenting, but now it's hard to see which things are checked. there's not as much contrast between a checkmark and a bullet as between a checkmark and blank space.

can you do the indenting without a bullet?

@astridu please advise as well on whether you think bullet vs checkmark is hard to discern

astridu:
I agree that the bullet makes it difficult to see the checkmark. Also, I don't see the correct indenting either.

Issue #2409, opened 2020-09-01
Top-level Navigation Item should be checked if any of its Children are checked-off


jermnelson:
When LD4P/sinopia_editor#2402 is merged, we will want the top-level navigation item to be checked-off if any of its children have content. 

Currently it displays the following:

![Left nav children](https://user-images.githubusercontent.com/71847/91900629-33e4cb00-ec5c-11ea-9029-9e56beba4883.png)

The *Uber template, property 1* top-level should have a checkmark to indicate that **any** of its children have content.

justinlittman:
Just the top level, or every parent? In the example above "Uber template2".

Have we tested this with users? I think it might actually be confusing, because it creates the impression that every property needs a "check" to be complete, which is not the case.

michelleif:
@astridu what do you think?

astridu:
What I described to users was that the top-level section will be checked off if there is anything filled out in its subsections. So even if only 1 of 5 subsections is filled out, the top-level section is checked off. The point of the check is to indicate it's been filled in.

mjgiarlo:
This *may* be blocked by #2473 but I am not sure.

PR #2502, opened 2020-09-18
Refactors left nav.

justinlittman:
## Why was this change made?
* Simplify the left nav components to meet current requirements.
* Provide marking when values present for property and ancestors.
* Provide marking when errors present for property and ancestors.
* Wrap left nav text.

Unit and integration test.

README updated for state model changes.


justinlittman:
refs #2473, #2447, #2409.

justinlittman:
image

image

image

image

jcoyne (approved):

Issue #2399, opened 2020-08-28
Authority Display with Context


jermnelson:
- Results are displayed by authority 
- Results display in a list with the search result bolded and other content below it in plain text. 
- Images appear to the left of the search result if available. 
- User can click on any of the items listed, and it will be populated in the field on the previous page. 
![image](https://user-images.githubusercontent.com/71847/91617648-aea0a400-e945-11ea-8ecd-4aa7041a3bd5.png)

Blocks #2400

PR #2505, opened 2020-09-18
Place the search results in tabs

### LINKED ISSUE #2398: Text field at top where user can search
Create text field for user to enter search term in Lookup Modal:

image

Blocks #2399
Blocks #2084

jcoyne:
## Why was this change made?

Ref #2399
Fixes #2398

Screen Shot 2020-09-18 at 4 38 21 PM
Screen Shot 2020-09-18 at 4 38 28 PM

### DIFF
--- a/__tests__/feature/editing/selectFromLookup.test.js
+++ b/__tests__/feature/editing/selectFromLookup.test.js
@@ -22,9 +22,9 @@ describe('selecting a value from the lookup modal', () => {
 
     fireEvent.click(addButton)
 
-    await waitFor(() => expect(screen.getByText('Lookup')).toHaveClass('modal-title'))
+    await waitFor(() => expect(screen.getByLabelText('Instance of (lookup)')))
 
-    const input = screen.getByPlaceholderText('Instance of (lookup)')
+    const input = screen.getByLabelText('Instance of (lookup)')
     fireEvent.change(input, { target: { value: 'test' } })
 
     expect(input).toHaveValue('test')
--- a/__tests__/components/editor/property/InputLookup.test.xjs
+++ b/__tests__/components/editor/property/InputLookup.test.xjs
@@ -1,7 +1,7 @@
 // Copyright 2019 Stanford University see LICENSE for license
 
 import React from 'react'
-import InputLookupModal from 'components/editor/property/InputLookupModal'
+import InputLookup from 'components/editor/property/InputLookup'
 import { fireEvent, wait, waitForElement } from '@testing-library/react'
 import { showValidationErrors, validateResource } from 'actions/index'
 import {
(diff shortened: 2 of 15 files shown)


package.json
```diff
@@ -83,6 +83,7 @@
     "query-string": "^6.13.1",
     "rdf-ext": "^1.3.0",
     "react": "^16.13.1",
+    "react-bootstrap": "^1.3.0",
```
justinlittman, line 86: We're intentionally not using react-bootstrap.

__tests__/feature/editing/selectFromLookup.test.js
```diff
@@ -22,9 +22,9 @@ describe('selecting a value from the lookup modal', () => {
 
     fireEvent.click(addButton)
 
-    await waitFor(() => expect(screen.getByText('Lookup')).toHaveClass('modal-title'))
+    await waitFor(() => expect(screen.getByLabelText('Instance of (lookup)')))
 
-    const input = screen.getByPlaceholderText('Instance of (lookup)')
+    const input = screen.getByLabelText('Instance of (lookup)')
```
justinlittman, line 27: Additional testing should be added to selectFromLookup to reflect the new functionality.

justinlittman (changes requested):

package.json
jcoyne: Why not? This allows us to build less logic, which means we can focus more on the users problem rather than figuring out which tab should show.

__tests__/feature/editing/selectFromLookup.test.js
jcoyne: There is no new functionality here. Both before and after the user is able to select a resource from a search. This only changes how the function appears.

justinlittman: Even if not testing the tabbing functionality, should be testing that tabs appear. (Does the test currently do this?)

package.json
jcoyne: I spent a good day working on a solution with the jquery tab stuff provided by bootstrap, but React completely interrupts its functionality (by re-rendering) so eventually I replaced that with react-bootstrap that works around this. I guess we could make our own tabs components, is that what you want?

__tests__/feature/editing/selectFromLookup.test.js
jcoyne: I don't have a good idea of how one writes this sort of test (that depends on an external service).  But the existing test should already cover selecting a resource, and with the testing already done by bootstrap-react, we should have sufficient coverage, right?

package.json
justinlittman: I'd suggest taking a look at how we do resource "tabbing" (i.e., when more than one resource is open in the editor).

jermnelson (approved):

PR #2518, opened 2020-09-22
[HOLD] Upgrade react-bootstrap-typeahead to version 5

jcoyne:
## Why was this change made?

This seems to fix the componentDidUpdate is deprecated warnings.

Screen Shot 2020-09-22 at 1 14 09 PM


justinlittman:
I did additional work, but came to the conclusion that version 5 is problematic.

Issue #2485, opened 2020-09-16
In reducers, correctly update nested objects


justinlittman:
See https://redux.js.org/recipes/structuring-reducers/immutable-update-patterns/

jermnelson:
@justinlittman can we close or is there remaining work for this ticket?

PR #2520, opened 2020-09-22
Moves search out of selectorReducer.

justinlittman:
refs #2485

## Why was this change made?
Refactor reducers.

Unit/feature.

NA


jcoyne (approved):

jcoyne:
"Refactor reducers." is what you are doing. You haven't really answered the why. It would be enlightening to your colleagues if you could help us understand why it's better to move search out of selector reducers

justinlittman:
I'm moving search out of selector reducers to make the state shallower, so it is easier to modify.

What I didn't say that I should have is that the primary goal of this ticket is to have our search reducers stop mutating state. The shallower state is just a convenience.

PR #2521, opened 2020-09-22
Move more out of selectorReducer

justinlittman:
## Why was this change made?
This is one of several PRs for making our reducers stop mutating state in accordance with Redux best practice. To assist with that, state is being made more shallow by eliminating `selectorReducer`.

Unit/feature.

NA

### DIFF
--- a/src/store.js
+++ b/src/store.js
@@ -26,6 +26,7 @@ export const initialState = {
       cursorOffset: null,
     },
     errors: {}, // {<error key>: [errors...]} or {<error key>: {<resourceKey>: [errors...]}}
+    historicalTemplates: [],
     lastSave: {}, // {<resourceKey>: date}
     modal: {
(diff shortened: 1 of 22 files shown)


src/store.js
```diff
@@ -57,16 +63,12 @@ export const initialState = {
   },
   selectorReducer: {
```
jermnelson, line 64: Is there a reason why we're keeping `selectorReducer` in the store?

jermnelson (changes requested):
Looks good, glad we're  getting rid of the `selectorReducer` from the Redux state. This is definitely a vestige of when we first were using Redux in the editor. Have one question on `store.js` otherwise I think this PR is ready to be merged.

src/store.js
justinlittman: Entities still need to be moved out. That will be the next PR ...

jermnelson (approved):

PR #2525, opened 2020-09-23
Publish Sinopia vocab

jermnelson:
## Why was this change made?
Fixes https://github.com/LD4P/sinopia/issues/277

Added unit tests for new component

n/a

### DIFF
--- a/__tests__/components/vocabulary/Vocab.test.js
+++ b/__tests__/components/vocabulary/Vocab.test.js
@@ -1,0 +1,51 @@
+// Copyright 2020 Stanford University see LICENSE for license
+
+import React from 'react'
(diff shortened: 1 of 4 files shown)


__tests__/components/vocabulary/Vocab.test.js
```diff
@@ -0,0 +1,51 @@
+// Copyright 2020 Stanford University see LICENSE for license
```
jermnelson, line 1: Initially tried to create a feature test but was running into a know [error](https://github.com/jsdom/jsdom/issues/2112) with `jsdom` and navigation so just testing with the component rendered with React Router variables.

justinlittman (approved):

PR #2527, opened 2020-09-23
Refactor entity reducers

justinlittman:
## Why was this change made?
This is the last of several PRs refactoring our reducers so that they don't mutate state.

Also, removed the editTemplateResource test since it was flapping and didn't actually test editing templates (since the root templates aren't available as fixtures).

Unit/feature

NA


jermnelson (approved):
Going through this PR helped me better understand where we were mutating state at deeper levels within reducer. @justinlittman thanks again for all your hard work in tracking down and refactoring the reducers.

PR #2529, opened 2020-09-23
Add view resource modal to search results rows

### LINKED ISSUE #2172: Allow user to view record before copying or editing it.
Currently user can only copy or edit record to view it. Allow user to view the record (maybe in an overlay) before copying/editing to see if it's the one they want.

mjgiarlo:
Fixes #2172

Peek 2020-09-24 10-45

**NOTE**: Since I made this screencast, I had to pull the change that blurred the background because we don't have a maintainable, testable approach for this in the editor.

## Why was this change made?

To allow catalogers to view resources among search results before editing or copying them.

CI, and in the browser, locally.

None

### DIFF
--- a/__tests__/feature/searchAndViewResource.test.js
+++ b/__tests__/feature/searchAndViewResource.test.js
@@ -1,0 +1,75 @@
+import { fireEvent, screen } from '@testing-library/react'
+import { renderApp } from 'testUtils'
+import * as sinopiaSearch from 'sinopiaSearch'
+import { resourceSearchResults } from 'fixtureLoaderHelper'
+import Config from 'Config'
+
+// This forces Sinopia server to use fixtures
+jest.spyOn(Config, 'useResourceTemplateFixtures', 'get').mockReturnValue(true)
+jest.mock('sinopiaSearch')
+// Mock jquery
+global.$ = jest.fn().mockReturnValue({ popover: jest.fn() })
+
+describe('searching and viewing a resource', () => {
+  sinopiaSearch.getTemplateSearchResults.mockResolvedValue({
+    totalHits: 0,
+    results: [],
+    error: undefined,
+  })
+
+  it('renders a modal without edit controls', async () => {
+    // Setup search component to return known resource
+    const uri = 'http://localhost:3000/resource/c7db5404-7d7d-40ac-b38e-c821d2c3ae3f'
+    sinopiaSearch.getSearchResultsWithFacets.mockResolvedValue(resourceSearchResults(uri))
+
+    renderApp()
+
+    fireEvent.click(screen.getByText('Linked Data Editor', { selector: 'a' }))
+    fireEvent.click(screen.getByText('Search', { selector: 'a' }))
+
+    fireEvent.change(screen.getByLabelText('Query'), { target: { value: uri } })
+    fireEvent.click(screen.getByTestId('Submit search'))
+
+    await screen.findByText(uri)
+    expect(screen.getByText('http://id.loc.gov/ontologies/bibframe/Fixture')).toBeInTheDocument()
+
+    // Modal hasn't rendered yet
+    expect(screen.queryByRole('dialog', { name: 'View Resource' })).not.toBeInTheDocument()
+    expect(screen.getByTestId('view-resource-modal').classList).not.toContain('show')
+
+    // Click the view icon next to the search result row
+    expect(screen.getByTestId(`View ${uri}`)).toBeInTheDocument()
+    fireEvent.click(screen.getByTestId(`View ${uri}`))
+
+    // Modal has now rendered
+    expect((await screen.findByTestId('view-resource-modal')).classList).toContain('show')
+    expect((await screen.findAllByText('Uber template1, property1', { selector: 'span' }))).toHaveLength(1)
+
+    // Edit controls are disabled in view modal
+    screen.getAllByTestId('Remove Uber template3, property1').forEach((removeButton) => {
+      expect(removeButton).toBeDisabled()
+    })
+    screen.getAllByText('ä').forEach((languageButton) => {
+      expect(languageButton).toBeDisabled()
+    })
+    expect(screen.getByPlaceholderText('Uber template3, property1')).toBeDisabled()
+    expect(screen.getByTestId('Edit Uber template3, property1')).toBeDisabled()
+    expect(screen.getByTestId('Change language for Uber template3, property1')).toBeDisabled()
+    expect(screen.getByTestId('Remove analog')).toBeDisabled()
+    screen.getAllByTestId('lookup').forEach((lookupValueControl) => {
+      expect(lookupValueControl).toBeDisabled()
+    })
+    screen.getAllByText('+').forEach((lookupControl) => {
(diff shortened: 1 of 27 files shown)


jermnelson:
@mjgiarlo, I agree with your message on Slack about the `props.readOnly`  being to spread through-out the components because they weren't initially designed for it. You've likely already considered this, but instead of passing down the readOnly boolean through props, could a boolean be set at the editor level in state (similar to `currentResource`) that the components could use in place of props?

mjgiarlo:
@jermnelson :speech_balloon: 
> @mjgiarlo, I agree with your message on Slack about the `props.readOnly` being spread through-out the components as they weren't initially designed for support readOnly. You've likely already considered this, but instead of passing down the readOnly boolean through props, could a boolean be set at the editor level in state (similar to `currentResource`) that the components could use in place of props?

I had a hack at that. Take another look and let me know how close I got?

jermnelson:
> I had a hack at that. Take another look and let me know how close I got?

@mjgiarlo this is exactly what I was thinking, looks good!

astridu:
This is exciting! Thank you! Overall looks good. I like the buttons on the bottom. A few suggestions (if it's possible):
- Reduce the width of the modal so there isn't so much empty space on the right.
- Can we white out the background so the modal pops more?
- Could you change the buttons to this blue: 00548F (not sure if you're looking for that level of feedback here?)
- Make the "x" in the upper right black and bold so it's more noticeable and easy to click.
- Is it possible to remove the horizontal scroll bar?

src/components/ViewResourceModal.jsx
```diff
@@ -0,0 +1,82 @@
...
+          <div className="modal-header">
+            <h4 className="modal-title">View Resource</h4>
+            <div className="view-resource-buttons">
+              <button type="button" className="close" onClick={ close } aria-label="Close">
+                <span aria-hidden="true">&times;</span>
+              </button>
+            </div>
+          </div>
+          <div className="modal-body view-resource-modal-content">
+            { currentResource && <ResourceComponent />}
```
justinlittman, line 55: I think that rather than `current resource &&` to handle when no resource, around L73 should be: `if(!resource) return null`.

src/components/editor/ResourceComponent.jsx
```diff
@@ -54,11 +55,12 @@ const ResourceComponent = () => {
           <ResourceURIMessage />
           <SaveAlert />
         </section>
-        {dataset
-         && <div className="alert alert-warning" role="alert">
-           <strong>Unable to load the entire resource.</strong> The unused triples are:
-           <RDFDisplay dataset={dataset} />
-         </div>
+        {
+          dataset && !readOnly
```
justinlittman, line 59: Totally agree with suppressing these for preview.

justinlittman (approved):
Looks good! Minor comments for your consideration.

justinlittman:
I meant to comment rather than approve, as this is only a draft.

src/components/ViewResourceModal.jsx
mjgiarlo: The complication is that this line will still error out even if we return null beneath it. And we can't add the check before this point, or the app will complain about using React Hooks with a conditional.

mjgiarlo:
@astridu :speech_balloon: 
> This is exciting! Thank you! Overall looks good. I like the buttons on the bottom. A few suggestions (if it's possible):
> 
> * Reduce the width of the modal so there isn't so much empty space on the right.
> * Can we white out the background so the modal pops more?
> * Could you change the buttons to this blue: 00548F (not sure if you're looking for that level of feedback here?)
> * Make the "x" in the upper right black and bold so it's more noticeable and easy to click.
> * Is it possible to remove the horizontal scroll bar?

I believe I've gotten all of these. Can you take a look at the new screencast above to verify and see if there are any other changes I should make? Thank you muchly!

astridu:
Looks excellent!! Thank you!

mjgiarlo:
@astridu Note that I had to back out the change that blurs the background because the method I used to achieve this is not compatible with React applications, and we do not have a good solution for this at this time (after consulting with both @justinlittman and @jermnelson who confirmed this).

Note to reviewers: the tests pass locally, but I get the dreaded warning about `act(() => { ... })`. And wrapping lines of the feature test in `act()` are not sufficient to make the warning go away. I'm not sure how much more time to sink into figuring this out. (I've read Kent Dodds' blog post about this phenomenon, and it seems like getting past this is a timesink.)

__tests__/feature/searchAndViewResource.test.js
```diff
@@ -0,0 +1,53 @@
...
+    expect(screen.getByTestId('view-resource-modal').classList).not.toContain('show')
+
+    // Click the view icon next to the search result row
+    expect(screen.getByTestId(`View ${uri}`)).toBeInTheDocument()
+    fireEvent.click(screen.getByTestId(`View ${uri}`))
+
+    // Modal has now rendered
+    expect((await screen.findByTestId('view-resource-modal')).classList).toContain('show')
+    expect((await screen.findAllByText('Uber template1, property1', { selector: 'span' }))).toHaveLength(1)
+
```
justinlittman, line 47: Should test that some of the form exists and is read only. (Just a sampling.)

justinlittman (changes requested):

__tests__/testUtilities/stateUtils.js
```diff
@@ -247,6 +248,92 @@ const buildResourceWithLiteral = (state, options) => {
   }
 }
 
+const buildReadOnlyResource = (state, options) => {
+  if (!options.hasReadOnlyResource) return
+
+  state.editor.currentResource = 't9zVwg2zO'
+  state.editor.currentResourceIsReadOnly = true
```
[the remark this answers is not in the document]
mjgiarlo: @justinlittman Ah, good point. Let me tweak this to use an existing resource, and perhaps add a flag to set this bit.

__tests__/feature/searchAndViewResource.test.js
mjgiarlo: @justinlittman yeah, this is as far as I could get before the jest output became incomprehensible. I wanted to write a fuller test but was eager to stop beating my head against the test suite. I'll give it another go today.

This is also where I have found Cypress tests easier to work with than feature tests, because I don't have a good grasp on what's going on in these feature tests (what data is there? what does the page look like when the failure occurs? etc.). `screen.debug()` helps a bit, but then I'm scrolling through thousands of lines of HTML in my terminal trying to suss out what I'm looking at. Anyway, preaching to the choir here, and I reckon these are all familiar pain points!

__tests__/testUtilities/stateUtils.js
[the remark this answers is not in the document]
mjgiarlo: On second thought, I think this was for an aborted earlier approach. Dumping it entirely. :clap:

__tests__/feature/searchAndViewResource.test.js
mjgiarlo: @justinlittman Done. Take another look?

justinlittman (approved):

jermnelson (approved):

PR #2533, opened 2020-09-24
prevent backspace for sending back button to browser when editing with diacritics

### LINKED ISSUE #2103: backspacing in diacritics box causes template to close
**Describe the bug**

When entering text in a literal field, I opened the diacritics module and chose some diacritics, then without clicking back to the literal field, I hit backspace, and the whole template closed.

When I opened the template again, my last template was still there, so I had two open templates. And in the new blank template when I tried to add diacritics to a literal field, it seems like nothing happened when I clicked on the diacritic in the diacritic module. 

I have to keep reopening the template to do my work.

stage

tested on mac and windows in firefox

doesn't matter which template, happens no matter which template

here's what it looked like before I hit backspace
Screen Shot 2020-03-11 at 1 53 20 PM

peetucket:
## Why was this change made?

Fixes #2103 

~~Also added the ability to hit the escape key and have it close the panel too (which follows other UI patterns).~~

Added a new test, but it still passes even without my change, so I don't think the test is working correctly.

### DIFF
--- a/__tests__/feature/editing/editingLiteral.test.js
+++ b/__tests__/feature/editing/editingLiteral.test.js
@@ -108,6 +108,10 @@ describe('editing a literal property', () => {
     fireEvent.click(await screen.findByText('ọ'))
     expect(input).toHaveValue('Foọ')
 
+    // press backspace while the focus is on the diacritic panel and make sure we are still on the edit page
+    fireEvent.keyDown(await screen.findByText('ọ'), { key: 'Backspace', code: 8, charCode: 8 })
+    expect(screen.queryAllByText('Latin Extended')).toHaveLength(1)
+
     // Close it
     fireEvent.click(diacriticBtn)
     expect(screen.queryAllByText('Latin Extended')).toHaveLength(0)
--- a/src/components/editor/diacritics/DiacriticsSelection.jsx
+++ b/src/components/editor/diacritics/DiacriticsSelection.jsx
@@ -32,7 +32,11 @@ const DiacriticsSelection = (props) => {
 
   if (!props.showDiacritics) return null
 
-  return (<div id={props.id} tabIndex="0" >
+  const keyPressHandler = (event) => {
+    if (event.which === 8) event.preventDefault() // backspace should not be passed to the browser as it can cause the page to go back
+  }
+
+  return (<div id={props.id} onKeyDown={keyPressHandler} role="presentation" tabIndex="0" >
     <div className="row">
       <section className="col-1 offset-11">
         <button className="btn btn-lg" onClick={closeHandler}>&times;</button>


src/components/editor/diacritics/DiacriticsSelection.jsx
```diff
@@ -36,7 +36,19 @@ const DiacriticsSelection = () => {
...
+        event.preventDefault() // backspace should not be passed to the browser as it causes the page to go back
+        break
+      case 27:
+        dispatch(hideDiacritics()) // esc should close the panel, all other keys can pass through
+        break
+      // no default
+    }
+  }
+
+  return (<div id="diacritics-selection" role="presentation" onKeyDown={keyPressHandler} className="sticky-top bg-white" tabIndex="0">
```
jcoyne, line 51: Shouldn't this be `role="dialog"`?

jcoyne (changes requested):

src/components/editor/diacritics/DiacriticsSelection.jsx
peetucket: Yeah, I had the same thought but it made the linter mad:

``` 51:11  error  Non-interactive elements should not be assigned mouse or keyboard event listeners  jsx-a11y/no-noninteractive-element-interactions```

But we really want a keyboard listener on this element because it is only way we could find to trap the backspace key.

File: __tests__/feature/editing/editingLiteral.test.js [hunk 2]
```diff
@@ -114,6 +114,11 @@ describe('editing a literal property', () => {
     fireEvent.click(screen.getByText('ọ'))
     expect(input).toHaveValue('Foọ')
 
+    // press backspace while the focus is on the diacritic panel and make sure we are still on the edit page
+    const diacriticsPanel = document.querySelector('#diacritics-selection')
+    fireEvent.keyDown(diacriticsPanel, { key: 'Backspace', code: 8, charCode: 8 })
```
jermnelson, line 119: The **code** parameter is incorrect, it should be `keyCode` with call looking like `fireEvent.keyDown(diacriticsPanel, { key: 'Backspace', keyCode: 8, charCode: 8 })`. But, here might be a case where jsdom is not replicating the browser functionality because even when I tested sending the correct backspace code (confirmed by doing a `console.log` in the `keyPressHandle` function) the Diacritics panel is still showing present and not closed.

src/components/editor/diacritics/DiacriticsSelection.jsx
jcoyne: Why is it on the modal in particular and not on the the whole page?  I'm seeing the backspace navigate back to the previous page regardless of whether the diacritics dialog is open.

peetucket: I was just responding to this particular ticket - not sure if that backspace key going back in the browser is desired in other contexts, so didn't want to unilaterally disable it everywhere.  The esc key was just a gratuitous addition that made sense to me.

jcoyne: @jermnelson @astridu ?

jermnelson: Doing some testing, the backspace issue exists only for Firefox and not Chrome (although you can download an [extension](https://chrome.google.com/webstore/detail/enable-backspace/ddijblnjiploanefjnbpimgbfjkpifia?hl=en#:~:text=Enable%20Backspace%20extension%20enables%20you,this%20functionality%20to%20your%20browser.) to restore this functionality). 

For example (all using Firefox):
*  Open  the editor with the four tabs **Resource Templates**, **Search**, **Load RDF**, and **Exports**. Click on any of **Search**, **Load RDF**, and **Exports** tabs, click backspace returns you to the **Resource Templates** tab.
*  Open a resource template from the **Resource Templates** and after the template is loaded into the **Editor** tab, hitting backspace returns you to the **Resource Templates** tab. However, we have the following wrinkles (I used the [ld4p:RT:bf2:Monograph:Work:Un-nested](https://development.sinopia.io/editor/ld4p:RT:bf2:Monograph:Work:Un-nested) template on development):
   *  Adding any text to a top-level property (used **Date of Work**) and hitting enter prevents the backspace from going back to the **Resource Templates** tab.
   *  Starting from newly loaded page with the Work template, adding text to a nested property (used **Title Information > Work Title > Preferred Title of Work**) and hitting enter to save the value but hitting the backspace goes back to the **Resource Templates** tab.

Working on this branch, when opening a template and opening the Diacritics panel but not selecting any vocabulary, hitting the backspace returns you to the **Resource Templates** tab. If you select any of the vocabularies and display the character map, the fix works. 

Given this behavior, I second @jcoyne suggestion to see if we can add this to the whole page to see if we can prevent the different examples above where hitting backspace returns to the previous page.

peetucket:
Let me see if I can get that (whole window block) to work this morning.

________________________________

src/components/editor/diacritics/DiacriticsSelection.jsx
peetucket: Just one thing to note, this behavior is not totally unexpected.  It turns out that Firefox and Chrome have different shortcut keys for going back a page.  Firefox uses delete, and Chrome removed that shortcut a couple years ago and uses a different key combination now instead.  The examples you site above seem to indicate that when the focus is in a text entry field, Firefox makes the delete key behave as expected (delete a key in the field being edited), and when the focus is elsewhere, it uses the default Firefox behavior of going back a page.

Just highlighting this, as I don't think this is a bug necessarily - it's acting the way the browser is intended to act.  I really like the idea of disabling for diacritics panel usage, because that is a mode where the user is entering text by clicking buttons (so they are in a sort of "text entry" mindset).  The other contexts are less clear.  Still experimenting with whole screen blocking if possible.

peetucket:
So I tried putting the keyPressDown event all the way the top id="app" level in `App.jsx` and it doesn't seem to work as expected.  It catches events in text box areas nested in the div but not otherwise.

As I was doing this, I realized two things:

1. If we did it at the top level using this approach, we would need to distinguish between being in a text entry area (where we want delete to be passed through) and not being in a text entry area (where we want to block it).

2. We would be constantly firing the javascript function each time a user pressed a key anywhere, and then doing nothing most of the time.  This seems like a lot of unnecessary overhead for Firefox.

I'm not sure this approach makes sense given these (unless I am misthinking it).

Maybe there are other approaches to disabling backspace globally as a navigation mechanism - I will have a quick look.

__tests__/feature/editing/editingLiteral.test.js
peetucket: i noted that right below there is keyPress event that uses the same syntax: https://github.com/LD4P/sinopia_editor/blob/2103-fix-backspace-bug/__tests__/feature/editing/editingLiteral.test.js#L48

Regardless, the test passes with either syntax

jermnelson:
@peetucket and @jcoyne is there anything remaining to do on this PR? (This is error still exists for the inline diacritics panel) Thanks!

peetucket:
In my local testing, this change continues to prevent the backspace key from sending the browser back in the inline diacritics panel (I rebased after the change to inline to verify).  It does not make this change globally across any other part of the page though, as was suggested as another possibility.  My comments above regarding why the global blocking may not be advisable still stand.

jermnelson:
I'm merging this PR now and created https://github.com/LD4P/sinopia_editor/issues/2641 for a future work-cycle to address.

PR #2543, opened 2020-09-25
Make property tooltips hide when elsewhere is clicked

### LINKED ISSUE #2539: Tooltip should close when user clicks away
**Describe the bug**

The **i** button doesn't disappear after it is clicked. I was expecting that it would dissapear after I navigate away from the **i** message.

image

If you don't click the **i** again to make it close, the little info box will persist forever!

stage

Mac Firefox
_Monograph Work (BF2) Un-Nested

Screen Shot 2020-09-25 at 12 45 27 PM

mjgiarlo:
Fixes #2539

## Why was this change made?

To make tooltips behave per user expectations.

CI and local browser.

**NOTE**: I fought with RTL/Jest for 1-2 hours and could not get the commented-out test line to pass. That line is the most important line to testing that the clicks work properly. Advice or code are most welcome.

None

### DIFF
--- a/src/components/editor/property/PropertyLabelInfoTooltip.jsx
+++ b/src/components/editor/property/PropertyLabelInfoTooltip.jsx
@@ -14,15 +14,20 @@ const PropertyLabelInfoTooltip = (props) => {
   }, [key])
 
   return (
-    <span data-toggle="popover"
-          data-placement="right"
-          data-container="body"
-          title={props.propertyTemplate.label}
-          data-content={props.propertyTemplate.remark}
-          key={key}
-          id={key}>
+    <a href="#"
+       className="tooltip-heading"
+       tabIndex="0"
+       data-toggle="popover"
+       data-trigger="focus"
+       data-placement="right"
+       data-container="body"
+       data-testid={props.propertyTemplate.label}
+       title={props.propertyTemplate.label}
+       data-content={props.propertyTemplate.remark}
+       key={key}
+       id={key}>
       <FontAwesomeIcon className="info-icon" icon={faInfoCircle} />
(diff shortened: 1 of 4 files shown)


src/components/editor/property/PropertyLabelInfoTooltip.jsx
```diff
@@ -14,15 +14,20 @@ const PropertyLabelInfoTooltip = (props) => {
...
-          data-container="body"
-          title={props.propertyTemplate.label}
-          data-content={props.propertyTemplate.remark}
-          key={key}
-          id={key}>
+    <a href="#"
+       className="tooltip-heading"
+       tabIndex="0"
+       data-toggle="popover"
+       data-trigger="focus"
```
mjgiarlo, line 21: This is the line that does the magic, but it only works inside an `<a>` element, not a `<span>` or `<button>` element.

src/styles/main.scss
```diff
@@ -92,6 +93,11 @@ $primary-disabled: lighten( map-get($theme-colors,"secondary"), 20% );
   border-style: dashed;
 }
 
+a.tooltip-heading {
```
mjgiarlo, line 96: Retain look & feel after changing `<span>` to `<a>`

justinlittman (approved):
For what it is worth, this is the sort of case where a small Cypress test might be appropriate. However, no need to hold up this fix.

mjgiarlo:
> For what it is worth, this is the sort of case where a small Cypress test might be appropriate. However, no need to hold up this fix.

Great idea. Here you go: https://github.com/LD4P/sinopia_editor/pull/2547

PR #2546, opened 2020-09-28
Move side navigation components into their own directory

jcoyne:
Refactor only (no functional changes)

## Why was this change made?
The property directory is for property editing components... whereas these components are for navigation


mjgiarlo (approved):
LGTM, but would like @jermnelson or @justinlittman to sign off first.

PR #2548, opened 2020-09-28
Remove bullets from the left side navigation

### LINKED ISSUE #2447: when field names wrap in left nav, second line should be indented

jcoyne:
## Why was this change made?

This makes the checkmarks more prominent

Fixes #2447 
Screen Shot 2020-09-28 at 11 23 20 AM

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -129,7 +129,7 @@ a.tooltip-heading {
   box-shadow: -3px 3px 10px 0px #999999;
 }
 
-.desc-panel{
+.desc-panel {
   padding: 0 10%;
   margin-bottom: 10%;
 }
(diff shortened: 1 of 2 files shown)


mjgiarlo (approved):
Nice, small change.

PR #2557, opened 2020-09-29
When the globe icon is clicked go to the source uri

jcoyne:

### DIFF
--- a/src/components/editor/property/InputLookup.jsx
+++ b/src/components/editor/property/InputLookup.jsx
@@ -46,7 +46,7 @@ const InputLookup = (props) => {
   const lookupSelection = props.lookupValues.map((lookupValue) => (
     <div key={lookupValue.key} className="lookup-value">
       <span key={lookupValue.key}>{lookupValue.label || lookupValue.literal}</span>
-      <a href={lookupValue.uri}>
+      <a href={lookupValue.uri} aria-label={`Link to value ${lookupValue.uri}`} target="_new">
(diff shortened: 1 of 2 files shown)


src/components/editor/property/InputLookup.jsx
```diff
@@ -46,7 +46,7 @@ const InputLookup = (props) => {
   const lookupSelection = props.lookupValues.map((lookupValue) => (
     <div key={lookupValue.key} className="lookup-value">
       <span key={lookupValue.key}>{lookupValue.label || lookupValue.literal}</span>
-      <a href={lookupValue.uri}>
+      <a href={lookupValue.uri} aria-label="Link to value URI" target="_new">
```
justinlittman, line 49: I would make the aria-label `Link to ${lookupValue.uri` to be consistent with our other aria labels.

src/hooks/useNavigableComponent.js
```diff
@@ -34,7 +34,8 @@ const useNavigableComponent = (rootSubjectKey, rootPropertyKey, componentKey) =>
   const dispatch = useDispatch()
   const navClickHandler = (event) => {
     dispatch(setCurrentComponent(rootSubjectKey, rootPropertyKey, componentKey))
-    event.preventDefault()
+    // Preventing default keeps the links from opening.  Do we need it?
+    // event.preventDefault()
```
justinlittman, line 38: Unsure. If it doesn't seem to break anything then maybe not.

justinlittman (approved):

Issue #2516, opened 2020-08-24
Users can see the last 5-20 records they modified or top 5-10 resource templates they use


astridu:
On their personalized dashboards, users can see the last 5-20 records or top 5-10 resource templates. They can click on any of them to edit/use them instead of having to search for them each time they log in.

![image.png](https://images.zenhubusercontent.com/5e13b724f29cd505881d7257/f4e89e06-4b65-465b-8dc7-0b65329bf2b3)

jermnelson:
@astridu should we store references to all of the records a cataloger modified and only display the latest 20 records? If so, should we also provide a pagination component to allow the cataloger to go back through the record history? (Same questions applies to the resource templates  as well)

astridu:
I think we should only display 10. In the user testing, the catalogers said it was helpful to see the last 5-10 they worked on. It would be nice to see a pagination component, but maybe we could cut it off at 50 records/resource templates so there aren't more than 5 pages to go through. The idea is to get to their item faster, and if they have to go through 5 clicks, it might have been easier to just search. Same applies for both records and resource templates. Thank you!

jermnelson:
Closing with PR #2578 being merged.

PR #2560, opened 2020-09-30
Updates API when user uses template, edits resource, and searches. Re…

justinlittman:
…trieves user data from API. Updates historical templates from user data.

refs #2516

## Why was this change made?
This is initial work on personalization.

Unit, feature, integration

NA


justinlittman:
Now persists across sessions:
image

__tests__/feature/editing/openAndCloseResource.test.js
```diff
@@ -1,9 +1,12 @@
...
 import { fireEvent, screen } from '@testing-library/react'
 import { createState } from 'stateUtils'
+import { featureSetup } from 'featureUtils'
 
-// Mock jquery
-global.$ = jest.fn().mockReturnValue({ popover: jest.fn() })
+featureSetup()
+//
+// // Mock jquery
+// global.$ = jest.fn().mockReturnValue({ popover: jest.fn() })
```
jermnelson, line 9: Can these comments be removed with the new `featureSetup` utility function>

__tests__/feature/editing/saveAndCopyResource.test.js
```diff
@@ -1,19 +1,18 @@
 import { renderApp, createHistory } from 'testUtils'
-import { fireEvent, wait, screen } from '@testing-library/react'
+import { fireEvent, waitFor, screen } from '@testing-library/react'
 import * as sinopiaApi from 'sinopiaApi'
-import Config from 'Config'
+// import Config from 'Config'
+//
+// // This forces Sinopia server to use fixtures
+// jest.spyOn(Config, 'useResourceTemplateFixtures', 'get').mockReturnValue(true)
```
jermnelson, line 7: Cruft?

jermnelson (changes requested):
A couple of minor fixes to remove commented out lines. Looks good but I have a question if a user's is in working in multiple environments (stage, production) will the saved templates, searches and resources be restricted to just that environment? I saw that in most cases we are capturing the URI but wasn't clear we're actually filtering out the results by environment. (This may be a non-issue if the user  is being saved by environment)

justinlittman:
@jermnelson Sorry for the cruft -- removed and ready for re-review.

Yes, the saved templates, searches and resources be restricted to just that environment. The history is being stored in the environment specific API.

jermnelson (approved):

PR #2561, opened 2020-09-30
Organize stylesheets

jcoyne:
## Why was this change made?
The styles are becoming disorganized and there isn't a clear place for different functions (e.g. defining the palate, set bootstrap variables, override bootstrap styles, set local variables, components, etc.)


justinlittman (approved):
Thanks for cleaning this up.

PR #2562, opened 2020-09-30
Remove inoperative class btn-default

jcoyne:
## Why was this change made?
This is a holdover from bootstrap 3


mjgiarlo (approved):
Goodbye, sweet prince...

PR #2570, opened 2020-09-30
Remember resource and template search state

### LINKED ISSUE #2227: Stickiness of "Resource Template" and "Resources" search
Our catalogers would find it useful, if when searching and selecting a Resource Templates in the Linked Data Editor, if one uses the back button, the last Resource Template search would remain instead of resetting to search all Resource Templates again.

e.g. in Resource Templates search: "Harvard Rare". 3 templates in results.

select Work, decide you need Instance instead hit backspace

"Harvard Rare" search is lost.  Have to research.  Would be better if the search is sticky.

mjgiarlo:
Fixes #2227

Also adds clear search button to both search screens which resets state and input elements.

See screencast to watch feature in action:

Peek 2020-10-01 11-26

## Why was this change made?

To prevent catalogers from having to re-type searches for resources and templates over and over again

CI and in browser locally

None

### DIFF
--- a/cypress/integration/end2end.spec.js
+++ b/cypress/integration/end2end.spec.js
@@ -67,7 +67,6 @@ describe('End-to-end test', () => {
           cy.url().should('include', '/templates')
 
           cy.get('#searchInput')
-            .type('resourceTemplate:bf2:WorkTitle')
             .should('have.value', 'resourceTemplate:bf2:WorkTitle')
 
           // eslint-disable-next-line cypress/no-unnecessary-waiting
@@ -114,6 +113,9 @@ describe('End-to-end test', () => {
 
     cy.get('a').contains('Search').click()
     cy.url().should('include', '/search')
+
+    // Test search clear button
+    cy.get('button[title="Clear query string"]').click()
 
     // Indexing latency is possible problem here.
     // Force is necessary because reflow of search inputs is suboptimal.
--- a/cypress/integration/leftNav.spec.js
+++ b/cypress/integration/leftNav.spec.js
@@ -50,7 +50,6 @@ describe('Left-nav test', () => {
         cy.url().should('include', '/templates')
 
         cy.get('#searchInput')
-          .type('resourceTemplate:testing:uber1')
           .should('have.value', 'resourceTemplate:testing:uber1')
 
         // eslint-disable-next-line cypress/no-unnecessary-waiting
--- a/src/components/search/Search.jsx
+++ b/src/components/search/Search.jsx
@@ -10,45 +10,57 @@ import {
   fetchSinopiaSearchResults as fetchSinopiaSearchResultsCreator,
   fetchQASearchResults as fetchQASearchResultsCreator,
 } from 'actionCreators/search'
-import { clearSearchResults as clearSearchResultsAction } from 'actions/search'
+import { clearSearchResults as clearSearchResultsAction, setSearchResults } from 'actions/search'
 import SinopiaSearchResults from './SinopiaSearchResults'
 import QASearchResults from './QASearchResults'
 import SearchResultsPaging from './SearchResultsPaging'
 import SearchResultsMessage from './SearchResultsMessage'
 import Alert from '../Alert'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
-import { faSearch } from '@fortawesome/free-solid-svg-icons'
+import { faTrashAlt, faSearch } from '@fortawesome/free-solid-svg-icons'
 import searchConfig from '../../../static/searchConfig.json'
-import { selectSearchError, selectSearchUri, selectSearchOptions } from 'selectors/search'
+import {
+  selectSearchError, selectSearchQuery, selectSearchUri, selectSearchOptions,
+} from 'selectors/search'
 
 const Search = (props) => {
   const dispatch = useDispatch()
-  const fetchQASearchResults = (queryString, uri, startOfRange) => dispatch(fetchQASearchResultsCreator(queryString, uri, { ...searchOptions, startOfRange }))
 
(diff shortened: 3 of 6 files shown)


justinlittman (changes requested):
Seems like the authority should also be sticky.

mjgiarlo:
> Seems like the authority should also be sticky.
done!

cypress/integration/end2end.spec.js
```diff
@@ -67,7 +67,6 @@ describe('End-to-end test', () => {
           cy.url().should('include', '/templates')
 
           cy.get('#searchInput')
-            .type('resourceTemplate:bf2:WorkTitle')
```
mjgiarlo, line 70: No need to type it in again because it's still there.

cypress/integration/leftNav.spec.js
```diff
@@ -50,7 +50,6 @@ describe('Left-nav test', () => {
         cy.url().should('include', '/templates')
 
         cy.get('#searchInput')
-          .type('resourceTemplate:testing:uber1')
```
mjgiarlo, line 53: Don't type it in because it's already there.

src/components/search/Search.jsx
```diff
@@ -10,45 +10,57 @@ import {
...
+import { faTrashAlt, faSearch } from '@fortawesome/free-solid-svg-icons'
 import searchConfig from '../../../static/searchConfig.json'
-import { selectSearchError, selectSearchUri, selectSearchOptions } from 'selectors/search'
+import {
+  selectSearchError, selectSearchQuery, selectSearchUri, selectSearchOptions,
+} from 'selectors/search'
 
 const Search = (props) => {
   const dispatch = useDispatch()
-  const fetchQASearchResults = (queryString, uri, startOfRange) => dispatch(fetchQASearchResultsCreator(queryString, uri, { ...searchOptions, startOfRange }))
```
mjgiarlo, line 26: I rejiggered the order of this function to concentrate the `use*` functions towards the top.

src/components/search/Search.jsx
```diff
@@ -10,45 +10,57 @@ import {
...
-    queryString, { ...searchOptions, startOfRange },
-  ))
+  const clearSearchResults = useCallback(() => dispatch(clearSearchResultsAction('resource')), [dispatch])
 
-  const fetchNewSinopiaSearchResults = (queryString) => dispatch(fetchSinopiaSearchResultsCreator(
-    queryString,
-  ))
+  const topRef = useRef(null)
 
+  const defaultUri = 'sinopia'
```
mjgiarlo, line 38: To name a thing is to imbue it with identity and power.

src/components/search/Search.jsx
```diff
@@ -127,14 +139,29 @@ const Search = (props) => {
               <div className="input-group" style={{ width: '100%' }}>
                 <input id="searchInput" type="text" className="form-control"
                        onChange={ (event) => setQueryString(event.target.value) }
-                       onKeyPress={handleKeyPress} />
+                       onKeyPress={handleKeyPress}
+                       placeholder="Enter query string"
```
mjgiarlo, line 143: Placeholder added for consistency w/ template search component.

src/components/search/Search.jsx
```diff
@@ -127,14 +139,29 @@ const Search = (props) => {
...
+                       value={ queryString } />
                 <span className="input-group-btn">
                   <button className="btn btn-default"
                           type="submit"
+                          title="Submit search"
```
mjgiarlo, line 148: Give it a title so we get hover text that explains the purpose of the icon.

src/components/search/Search.jsx
```diff
@@ -127,14 +139,29 @@ const Search = (props) => {
...
                           aria-label="Submit search"
                           data-testid="Submit search">
                     <FontAwesomeIcon className="fa-search" icon={faSearch} />
                   </button>
+                  <button className="btn btn-default"
+                          type="button"
+                          aria-label="Clear query string"
+                          title="Clear query string"
```
mjgiarlo, line 156: Give it a title so we get hover text that explains the purpose of the icon.

src/components/search/Search.jsx
```diff
@@ -127,14 +139,29 @@ const Search = (props) => {
...
+                          data-testid="Clear query string"
+                          onClick={() => {
+                            setQueryString('')
```
mjgiarlo, line 159: Blank the query string in the input field, reset the URI in the dropdown, and ditch the search results.

src/components/search/Search.jsx
```diff
@@ -127,14 +139,29 @@ const Search = (props) => {
...
+                            setUri(defaultUri)
+                            clearSearchResults()
+                          } }>
+                    <FontAwesomeIcon className="trash-icon" icon={faTrashAlt} />
```
mjgiarlo, line 163: I assume trash icon is OK in this context but happy to use a different FA icon. Cc: @astridu

src/components/templates/TemplateSearch.jsx
```diff
@@ -10,25 +10,28 @@ import Alert from '../Alert'
...
   const dispatch = useDispatch()
   // Tokens allow us to cancel an existing search. Does not actually stop the
   // search, but causes result to be ignored.
   const tokens = useRef([])
 
-  const [query, setQueryText] = useState('')
+  const error = useSelector((state) => selectSearchError(state, 'template'))
+  const lastQueryString = useSelector((state) => selectSearchQuery(state, 'template'))
+
+  const [queryString, setQueryString] = useState(lastQueryString || '')
```
mjgiarlo, line 27: Aligning variable names across like components.

justinlittman (approved):

PR #2577, opened 2020-10-01
Adds nonexistent relative href to prevent jumping to top of screen

jermnelson:
## Why was this change made?
Fixes problem when user clicks on a tooltip and the page jumps to the top of the window.

Testing suite

n/a


mjgiarlo (commented):
I hate to suggest it but might this be a good candidate for a cypress test? We test other nav-related stuff there, I think?

mjgiarlo (approved):

PR #2582, opened 2020-10-02
Add recent searches to dashboard

justinlittman:
## Why was this change made?
* Add recent search history to the dashboard.
* Refactors search functions into hook.

Unit, feature, integration

NA


justinlittman:
image

src/components/dashboard/SearchList.jsx
```diff
@@ -0,0 +1,70 @@
...
+        <table className="table table-bordered search-list">
+          <thead>
+            <tr>
+              <th style={{ backgroundColor: '#F8F6EF', width: '30%' }}>
+                Authority
+              </th>
+              <th style={{ backgroundColor: '#F8F6EF', width: '66%' }}>
+                Query
+              </th>
+              <th style={{ backgroundColor: '#F8F6EF', width: '4%' }}
```
jermnelson, line 52: Since the `backgroundColor` is common to all there table-headers, maybe move to a CSS class?

jermnelson, line 52: Since the `backgroundColor` is common to all three of these table headers, maybe move them to a CSS class?

jermnelson (changes requested):
A couple of minor comments, I think this PR is a good user-facing improvement.

docker-compose.yml
```diff
@@ -62,13 +62,13 @@ services:
...
-  #   environment:
-  #     ME_CONFIG_MONGODB_ADMINUSERNAME: root
-  #     ME_CONFIG_MONGODB_ADMINPASSWORD: sekret
+  mongo-express:
+    image: mongo-express
+    ports:
+      - 8082:8081
+    environment:
+      ME_CONFIG_MONGODB_ADMINUSERNAME: root
+      ME_CONFIG_MONGODB_ADMINPASSWORD: sekret
```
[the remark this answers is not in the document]
justinlittman: Oops. Fixed.

src/components/dashboard/SearchList.jsx
justinlittman: Good idea. Made change elsewhere as well.

jermnelson (approved):

PR #2567, opened 2020-09-30
add search spinner while waiting for search to run

peetucket:
## Why was this change made?

Addresses part of #2182 (search spinner)
Timeouts are in #2583

- adds search spinner when searches are executing (sinopia or QA)
- clears any current search results as soon as a new search is started (and shows the spinner when running the new search of course)

This is how the loading spinner looks: (Note: the "results loading" text is in a span that only shows for screen readers.)

spinner

## How was this change tested?

- Localhost browser

## Which documentation and/or configurations were updated?

None

Issue #2182, opened 2020-05-07
Wait state/timeouts needed for searches


astridu:
Note: search spinner is now done and merged with #2567 ... this ticket remains open for now for the 30 second timeout in #2583

Spinner or other wait state to indicate that lookup promise is working and not frozen.

astridu:
A couple options (1. bootstrap spinner, 2. skeleton)
![image](https://user-images.githubusercontent.com/42216982/82008886-b8b2c700-9622-11ea-992d-e9653d767396.png)
![image](https://user-images.githubusercontent.com/42216982/82008903-c49e8900-9622-11ea-9734-369938909592.png)

jermnelson:
Add a timeout (30 seconds but confirm with @astridu) and display an error message.

astridu:
Yes, 30 seconds sounds like a good amount of time, and then send an error message if nothing comes back in that time.

peetucket:
use bootstrap spinner: https://getbootstrap.com/docs/4.5/components/spinners/

peetucket:
Note: search spinner is now done and merged with #2567 ... this ticket remains open for now for the 30 second timeout in #2583

jgreben:
We decided not to pursue this any further unless it becomes an future bug report.

PR #2583, opened 2020-10-02
add timeouts to QA search

peetucket:
## Why was this change made?

Addresses part of #2182 (search timeouts)
Search spinner is in #2567

This is the timeout error message (not great, but matches the error message style you get for other search messages, and a custom message will potentially be substantially more work to implement):

Screen Shot 2020-09-30 at 4 40 45 PM

Updated tests


peetucket:
Using this technique for timeing out the fetch calls: whatwg/fetch#951 (comment)

jgreben:
We decided not to pursue this any further unless it becomes a bug report.

PR #2585, opened 2020-10-02
Add resource history to dashboard

justinlittman:
## Why was this change made?
So that a user can quickly find resources that she has been editing.

Unit, feature, integration

NA


justinlittman:
image

jermnelson (approved):

PR #2590, opened 2020-10-04
Fix broken test.

justinlittman:
## Why was this change made?
I prefer green.

NA

NA


mjgiarlo (approved):
Feel free to merge when eslint is happy.

PR #2612, opened 2020-10-06
Display nested resources from dataset and template

justinlittman:
## Why was this change made?
To fix a bug whereby when loading a resource that has a property that has multiple nested resource templates, only the resource templates with data were being displayed.

Unit

NA


justinlittman:
Before:
image

After:
image

jermnelson (approved):

PR #2625, opened 2020-10-07
Set maxwidth for MARC preview.

### LINKED ISSUE #2624: MARC preview falls off screen
When the MARC is too large, the drop left drops right which requires additional scrolling.

image

justinlittman:
closes #2624

## Why was this change made?
The MARC preview was dropping right, which made it difficult to use.

Locally

NA

### DIFF
--- a/src/components/editor/actions/MarcButton.jsx
+++ b/src/components/editor/actions/MarcButton.jsx
@@ -67,7 +67,7 @@ const MarcButton = () => {
   }
 
   return (
-    <div className="dropleft">
+    <div className="btn-group dropleft">
       <button type="button"
               id="marcBtn"
               className={btnClasses.join(' ')}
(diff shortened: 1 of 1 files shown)


justinlittman:
Before:
image

After:
image

src/components/editor/actions/MarcButton.jsx
```diff
@@ -83,7 +83,7 @@ const MarcButton = () => {
             <button className="btn btn-link dropdown-item" onClick={(event) => handleDownloadTxt(event)}>Download text</button>
             <button className="btn btn-link dropdown-item" onClick={(event) => handleDownloadMarc(event)}>Download MARC</button>
             <pre style={{
-              marginLeft: '10px', marginRight: '10px', paddingLeft: '10px', paddingRight: '10px',
+              marginLeft: '10px', marginRight: '10px', paddingLeft: '10px', paddingRight: '10px', maxWidth: '750px',
```
jermnelson, line 86: This may be better as a CSS class

jermnelson (changes requested):

src/components/editor/actions/MarcButton.jsx
justinlittman: In principle, I agree. However, I've been treating the current MARC integration as a proof-of-concept and trying to minimize the entanglement with the rest of the code base, figuring that it will be replaced once we better understand the use case. (This is why MarcButton is a single component.) Thus, in this case I'd recommend that we keep the CSS local rather than adding to the stylesheet.

jermnelson (approved):

PR #2626, opened 2020-10-07
Adds lookup pagination.

### LINKED ISSUE #2538: QA/Lookup Pagination
The Lookup Panel only displays the top 8 results from QA. The panel should support pagination so the user can load more results from QA.

justinlittman:
closes #2538

## Why was this change made?
Yeah, need more than 8 results.

Unit, feature

NA


justinlittman:
image

justinlittman:
image

src/components/search/SearchResultsPaging.jsx
```diff
@@ -69,7 +62,9 @@ const SearchResultsPaging = (props) => {
 
 SearchResultsPaging.propTypes = {
   changePage: PropTypes.func.isRequired,
-  searchType: PropTypes.oneOf(['resource', 'template']).isRequired,
+  totalResults: PropTypes.number.isRequired,
```
jermnelson, line 65: I'm encountering an error related to a couple of the authorities in the `ld4p:RT:bf2:Monograph:Work:Un-nested` template (from production) in the Subject of Work. When doing a search the `OCLCFAST geographic (QA) - direct` and `OCLCFAST topic (QA) - direct` are causing an error and display a **Nan** for the total records:

Screen Shot 2020-10-07 at 5 41 10 PM

From this error in the web console: 
Screen Shot 2020-10-07 at 5 41 24 PM

I think it is because the returned value from QA for the OCLC FAST direct has a string value in the *total_records* property in header::
```javascript
response_header: {
 start_record: 0, 
 requested_records: "DEFAULT", 
 retrieved_records: 8, 
total_records: "NOT REPORTED"}
```

jermnelson (changes requested):
Looks good; I discovered an edge case regarding the returned value from QA call that we should at least account for in the `SearchResultsPaging.jsx`

src/components/search/SearchResultsPaging.jsx
justinlittman: Great catch. I added handling for this.

jermnelson (approved):

PR #2638, opened 2020-10-08
Aligns tabs for InputLookup.

### LINKED ISSUE #2575: in QA Modal with more than one authority, the "tabs" don't look like tabs, it's hard to distinguish among the authorities
@astridu can we make it more obvious that these are "tabs"? like a real "tab" visual? or a border between the authorities or? 

example:
https://development.sinopia.io/editor/ld4p:RT:bf2:Monograph:Work:Un-nested
field "Subject of the work"

justinlittman:
closes #2575

## Why was this change made?
It was a mess.

Locally.

NA

### DIFF
--- a/src/components/editor/Tabs.jsx
+++ b/src/components/editor/Tabs.jsx
@@ -2,6 +2,7 @@
 
 import React from 'react'
 import PropTypes from 'prop-types'
+import _ from 'lodash'
 
 const Tabs = (props) => {
   const children = props.children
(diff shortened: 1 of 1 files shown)


justinlittman:
image

jcoyne (changes requested):
This seems a drastic departure from the recommended markup in https://getbootstrap.com/docs/4.5/components/navs/#tabs  Can you explain more about why this is needed?

~EDIT~: ah, two column layout.

jcoyne (approved):

Issue #2494, opened 2020-09-17
PropertyLabel component should be an HTML <label>


jcoyne:
This is an accessiblity concern. All fields should have visible labels.

jermnelson:
All of the componets depend on the `PropertLabel` to display the label. Currently we are using a `<span>`, this should be changed to a `<label>` with a **for** attribute.

jmartin-sul:
implementation for this ticket partially working, closing in favor of #2683, which captures what's not yet working.

PR #2647, opened 2020-10-08
turn PropertyLabel into an HTML <label>

jmartin-sul:
## Why was this change made?

to make usage of the `PropertyLabel` component more accessible, per #2494

_note_, per #2683, this does not always make the `<label>` association work, despite the right pieces seeming to be in place.

manual testing in browser, unit tests

n/a

### DIFF
--- a/src/components/editor/property/InputList.jsx
+++ b/src/components/editor/property/InputList.jsx
@@ -94,6 +94,7 @@ const InputList = (props) => {
           onBlur={selectionChanged}
           aria-label={`Select ${props.propertyTemplate.label}`}
           data-testid="list"
+          id={props.propertyLabelId}
         >{options}
         </select>
         {error && <span className="invalid-feedback">{error}</span>}
@@ -109,6 +110,7 @@ InputList.propTypes = {
   propertyTemplate: PropTypes.object.isRequired,
   values: PropTypes.array,
   addValue: PropTypes.func,
+  propertyLabelId: PropTypes.string.isRequired,
 }
 
 const mapStateToProps = (state, ownProps) => ({
--- a/src/components/editor/property/InputLiteral.jsx
+++ b/src/components/editor/property/InputLiteral.jsx
@@ -75,6 +75,7 @@ const InputLiteral = (props) => {
   }
 
   // TextareaAutosize does not disable well, so using a disabled input.
+
   return (
     <div className="form-group">
       <div className="input-group" onBlur={handleBlur} id={id}>
@@ -93,7 +94,8 @@ const InputLiteral = (props) => {
               onKeyDown={handleKeyDown}
               onClick={handleClickDiacritics}
               value={currentContent}
-              ref={inputLiteralRef} />
+              ref={inputLiteralRef}
+              id={props.propertyLabelId} />
         }
         <div className="input-group-append" tabIndex="0">
           <button className="btn btn-outline-primary"
@@ -120,6 +122,7 @@ InputLiteral.propTypes = {
   propertyTemplate: PropTypes.object.isRequired,
   addValue: PropTypes.func,
   content: PropTypes.string,
+  propertyLabelId: PropTypes.string.isRequired,
 }
 
 const mapStateToProps = (state, ownProps) => ({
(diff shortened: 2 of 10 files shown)


jmartin-sul:
This does seem to push the ball forward in terms of what #2494 is asking for.  But since I'm not the most adept with accessibility stuff, maybe someone more experienced will see an obvious problem with this implementation that makes scrapping this PR and trying a different approach preferable.  But if this PR does help, I've filed https://github.com/LD4P/sinopia_editor/issues/2683 to address the seeming shortcomings here (in which case maybe we close the original issue this addresses, in favor of the follow-on I just filed?).

src/components/editor/property/InputList.jsx
```diff
@@ -109,6 +110,7 @@ InputList.propTypes = {
   propertyTemplate: PropTypes.object.isRequired,
   values: PropTypes.array,
   addValue: PropTypes.func,
+  propertyLabelId: PropTypes.string.isRequired,
```
jmartin-sul, line 113: i made this a required prop in all of the places where it seemed like the value would always be provided (by `grep`ing for usage of the component).  happy to loosen if this seems overly strict.

jermnelson (approved):
Looks good; since I'm pushing for doing a complete accessibility audit as part of the next Sinopia work-cycle, I think #2683 would be actionable then.

src/components/editor/property/InputLiteral.jsx
```diff
@@ -75,6 +75,7 @@ const InputLiteral = (props) => {
   }
 
   // TextareaAutosize does not disable well, so using a disabled input.
+
```
jmartin-sul, line 78:
```suggestion
```

d'oh, this extra line break was editing cruft on my part, feel free to squash in this suggestion for deleting the new line if its bothersome.

src/components/editor/property/InputList.jsx
jermnelson: That is fine, I prefer having props to be strict as it usually points out an error or edge case we are not handling properly.

src/components/editor/property/InputLiteral.jsx
jermnelson: If eslint doesn't complain than I'm fine with the line break :-).

src/components/editor/property/NestedProperty.jsx
```diff
@@ -6,21 +6,23 @@ import NestedPropertyHeader from './NestedPropertyHeader'
 import PropertyComponent from './PropertyComponent'
 import { selectNormProperty } from 'selectors/resources'
 import { connect } from 'react-redux'
+import shortid from 'shortid'
 import useNavigableComponent from 'hooks/useNavigableComponent'
 import { selectPropertyTemplate } from 'selectors/templates'
 
 const NestedProperty = (props) => {
   const [navEl, navClickHandler] = useNavigableComponent(props.property.rootSubjectKey, props.property.rootPropertyKey, props.property.key)
+  const propertyLabelId = `labelled-by-${shortid.generate()}`
```
jmartin-sul, line 15: i had initially used `labeled-by-${shortid.generate()}`, and a cursory web search indicates either spelling is acceptable.  but i ended up switching to this spelling, because it's what my browser's accessibility debugger used, and consistency seemed desirable.

src/components/editor/property/NestedPropertyHeader.jsx
```diff
@@ -97,6 +98,7 @@ NestedPropertyHeader.propTypes = {
   hideProperty: PropTypes.func,
   id: PropTypes.string,
   resourceKey: PropTypes.string,
+  propertyLabelId: PropTypes.string,
```
jmartin-sul, line 101: i couldn't tell from `grep`ing the code how this would _not_ be provided, but when i marked it required, i got test failures.  happy to revisit if there was something in the tests i should've fixed.

src/components/editor/property/InputList.jsx
jmartin-sul: cool, that was also my thinking

PR #2660, opened 2020-10-09
Adds result counts for lookup tabs.

### LINKED ISSUE #2653: Result counts in Authorities Tab
A nice feature for the InputLookup component's Authorities tabs would provide a count of the available results from QA.

justinlittman:
closes #2653

## Why was this change made?
So that the user knows which tabs have results.

I will be adding tests for InputLookup in a separate PR.

NA


justinlittman:
image

jermnelson (approved):

PR #2664, opened 2020-10-09
Makes InputLookup search button consistent with main search.

justinlittman:
## Why was this change made?
To make InputLookup search consistent with main search (which recently changed).

Locally.

NA


justinlittman:
image

jermnelson (approved):
Approved pending rebase.

PR #2674, opened 2020-10-09
Makes lookup search input and values look better

### LINKED ISSUE #2640: Lookup values display poorly when > 1 line
image

jgreben:
## Why was this change made?
- Fixes display of lookup values for a non-nested resource lookup, and to be adaptive to browser width.
- Fixes display of resource template search to be adaptive to browser width.

N/A only CSS changes were made

Before:
Screen Shot 2020-10-09 at 2 28 03 PM

After:
Screen Shot 2020-10-09 at 2 28 14 PM

Before:
Screen Shot 2020-10-09 at 1 58 18 PM
After:
Screen Shot 2020-10-09 at 1 58 07 PM

### DIFF
--- a/src/components/editor/property/InputLookup.jsx
+++ b/src/components/editor/property/InputLookup.jsx
@@ -94,7 +94,6 @@ const InputLookup = (props) => {
       <div className="form-inline" style={{ marginBottom: '5px' }}>
         <div className="input-group" onBlur={handleBlurDiacritics} id={id}>
           <input id={inputId} type="text" className={controlClasses.join(' ')}
-                 style={{ width: '750px' }}
                  onChange={handleChangeDiacritics}
                  onKeyDown={handleKeyDown}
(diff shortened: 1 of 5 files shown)


justinlittman (changes requested):
I'm not sure what the desired design was, but:

InputLookup lookup button is now on a separate line (unlike search):
image

The resource templates search is also split onto multiple lines:
image

jgreben:
Oops, the rebase messed it up..I will push another commit.

jgreben:
Looks like another PR fixed the problem reported by @astrid, but this PR fixes an issue with the width of the non-nested lookup input and values not matching the containing Div when the browser size is altered.

justinlittman (changes requested):
Getting closer, but this looks odd:
image

justinlittman (approved):
Thanks!

PR #2718, opened 2020-11-12
OnClick function now takes event

### LINKED ISSUE #2716: Create New button on PCC_BF2_Work_Monograph template does not work
The Create New button which appears with the In series RT on PCC_BF2_Work_Monograph template does not function as intended. When I click on the button, I'm prompted to select a template from a drop-down list. In most cases, a Serial Work template would be the logical choice (either the LD4P Serial Work or PCC Serial Work). No matter what I select, Sinopia instead sends me to a blank Monograph Work template, and the monograph work I was working on is gone. There is an error message that flashes between the time of selecting from the menu and the blank Work template coming up, but it goes by too fast to read.

I am using Sinopia Stage in Firefox 82.0.3.

This is possibly related to issue #2712 .

jermnelson:
## Why was this change made?
Fixes #2716 

Testing suite

n/a

### DIFF
--- a/src/components/editor/property/ResourceList.jsx
+++ b/src/components/editor/property/ResourceList.jsx
@@ -22,7 +22,8 @@ const ResourceList = (props) => {
   const subjectTemplate = useSelector((state) => selectSubjectTemplate(state, subject?.subjectTemplateKey))
 
   useEffect(() => {
-    const handleChange = (resourceTemplateId) => {
+    const handleChange = (resourceTemplateId, event) => {
+      event.preventDefault()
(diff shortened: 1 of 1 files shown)


justinlittman (changes requested):
I'd refactor a bit to get the actual component generation out of the effect. That should allow you to avoid disabling `react-hooks/exhaustive-deps`.

jermnelson:
> I'd refactor a bit to get the actual component generation out of the effect. That should allow you to avoid disabling `react-hooks/exhaustive-deps`.

I was going down another rabbit hole trying to extract the function and was getting different linting errors and warnings. Decided to revert to master and just fix the existing function.

justinlittman (approved):

PR #2760, opened 2021-01-27
Create New button for Sinopia Lookups

### LINKED ISSUE #2716: Create New button on PCC_BF2_Work_Monograph template does not work

jermnelson:
## Why was this change made?
Fixes #2716 . Adjusted fixture handling in sinopiaSearch to better replicate searching with ElasticSearch.

Added new feature test to confirm change

n/a

### DIFF
--- a/__tests__/feature/editing/editingLookup.test.js
+++ b/__tests__/feature/editing/editingLookup.test.js
@@ -399,3 +399,22 @@ describe('selecting a value from lookup', () => {
     await waitFor(() => expect(screen.queryByText('Shania Twain', { selector: 'div.lookup-value span' })).not.toBeInTheDocument())
   })
 })
+
+describe('adding a template from Sinopia lookup', () => {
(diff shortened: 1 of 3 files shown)


jermnelson:
Keeping this PR as a draft because the testing mocks for the **Create New** button includes all of the existing templates instead of filtering based on the authority lookup's class.

mjgiarlo (approved):

PR #2779, opened 2021-03-09
Fix entering value onBlur. 

### LINKED ISSUE #2361: Editor: when you type a literal value and don't hit Enter, Sinopia doesn't save the value when you Save the resource
When editing (on version that's currently in DEV), whether editing a Template or  Resource:
1) type a value in a literal field but **don't** hit enter (example template to try "more:amazing", field "This is field one...")
2) the Save button is blue, click Save, and your resource is saved but the value you entered is not saved

By contrast, on Stage and PRD, this happens:
1) type a value in a literal field but **don't** hit enter (example template "test:template:one_literal", field "this is a literal field"
2) Save button is greyed out
3) click grey Save button, your value is "entered" into the form, and the Save button is now blue
4) save resource and your value is saved

justinlittman:
closes #2361

## Why was this change made?
Fix a regression introduced when moving diacritics into a hook.

Locally, dev environment by PO

NA

### DIFF
--- a/src/components/editor/property/InputLiteral.jsx
+++ b/src/components/editor/property/InputLiteral.jsx
@@ -18,14 +18,14 @@ import useDiacritics from 'hooks/useDiacritics'
 const InputLiteral = (props) => {
   const inputLiteralRef = useRef(null)
   const [lang, setLang] = useState(defaultLanguageId)
+  const id = `inputliteral-${props.property.key}`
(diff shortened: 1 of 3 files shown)


justinlittman:
Definitely need better tests. Not too sure how to write them for this.

jermnelson (approved):

PR #2780, opened 2021-03-09
Fixes tab closing.

### LINKED ISSUE #2183: wrong tabs closing
There was a bug reported a  couple months ago and then corrected, where if there were multiple tabs, you'd close a record and assume that the record you were looking at (the highlighted tab) would close, but one of the other tabs would close instead.  So if you aren't careful you can close an unsaved record (since they are back to saying changes are unsaved again). 
I'm in production, using Firefox on a Windows machine.  I tried to close aw ork record but one of the administrative  metadata records closed instead.

justinlittman:
closes #2183

## Why was this change made?
The close modal was getting confused about which resource to close.

Local, PO on dev.

NA

### DIFF
--- a/src/components/editor/actions/CloseButton.jsx
+++ b/src/components/editor/actions/CloseButton.jsx
@@ -19,7 +19,7 @@ const CloseButton = (props) => {
 
   const handleClick = (event) => {
     if (resourceHasChanged) {
-      dispatch(showModal('CloseResourceModal'))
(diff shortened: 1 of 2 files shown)


justinlittman:
Your're right @mjgiarlo. Fixing.

mjgiarlo (approved):
This looks better. :+1:

mjgiarlo (approved):

Issue #2787, opened 2021-03-16
Nested resources & blank nodes & lookups (PCC wish list no. 2)


NancyL:
Not sure whether to call this a bug or an enhancement request.

When we create a nested resource in a template, we create a blank node. This makes sense for many of these resources such as bf:Contribution or bf:Title, where we would not put a URI in the position of the blank node.

However, in some cases, we create a nested resource only in order to identify the class of the object property. This happens most often combined with a lookup, which, by definition, has an existing URI. In order to have a place for that URI, we have to put in an "rdf:value" as below with bf:agent:

          bf:contribution [a bf:Contribution ;
                bf:agent [a bf:Person ;
                **rdf:value URI ;** 
                rdfs:label "label" 
                ].
          ] .

While this is valid rdf, it is not good rdf, in that it adds in an extra blank node where one is not necessary. The rdf should look as follows:

           bf:contribution [a bf:Contribution ;
                    bf:agent <URI for Agent>  .
                    <URI for Agent> [a bf:Person ;
                          rdfs:label "label" 
                    ].
            ] .

LC fixes this with some programming behind the scenes. They used to use bflc:target to indicate where to do it. They don't any more and I don't understand how they do it now.

This extra blank node is and will remain a major issue in interoperability given that other systems will not be looking for a blank node in that position, but instead a direct URI. I have no idea what the solution is. It is no. 2 on PCC's wish list; no. 1 on mine!

michelleif:
can you say more about why the class needs to be specified for the object?

NancyL:
Many of these have multiple object types to choose from and need to be specified. Not all vocabularies specify they are instances of a particular type (bf:Person is specified in NAF, but not VIAF, ISNI, Wikidata) and in sharing our data the expectation is to see this associations, at least in SVDE & with LC.

justinlittman:
The request makes sense to me; I have no idea what it will take to implement.

michelleif:
a couple follow-up questions:
1) are there other examples besides Agent Class in existing templates where you need to create additional statements about the object? 
2) in the case of needing to say what class an Agent is...are there other ways to express that an Agent is an instance of a particular class? if we are using agent URIs from Wikidata, for example, could you add a statement in Wikidata: agent123 hasType bf:person?

michelleif:
let's ask LC to show how they do this.

justinlittman:
Do what?

michelleif:
[...]

PR #2815, opened 2021-05-06
Adds support for suppressing blank node when a nested resource.

justinlittman:
refs #2787

## Why was this change made?
To produce/load RDF that avoids blank nodes.

Unit, dev

README

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,6 +222,7 @@ The following are only in the resource subject (that is, the base subject).
   author: <author>,
   remark: <remark>,
   date: <date>,
+  suppressible: <true | false>,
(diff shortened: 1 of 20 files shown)


jermnelson (approved):
Thanks @justinlittman, excited that this PR addresses one of the main issues regarding the BF RDF produced in Sinopia verses LOC and ShareVDE. Just one question, should the draft status on the Sinopia API PR [112](https://github.com/LD4P/sinopia_api/pull/112) be changed? I think the API PR needs to be merged at the same time as this PR before Sinopia can use the suppressible resource attribute when editing/creating new resource template.

PR #2855, opened 2021-07-12
Dependency update version bump

aaron-collier:
## Why was this change made?

Monday dependency updates

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,23747 @@
 {
   "name": "sinopia_editor",
-  "version": "3.5.33",
-  "lockfileVersion": 1,
+  "version": "3.5.34",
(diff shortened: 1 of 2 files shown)


jmartin-sul (commented):
any idea why the `package-lock.json` diff is so large for a version bump right after weekly dep updates?  i've definitely seen this happen on the sinopia_editor dep updates PR before, so i don't think it's unique to this one... just hesitant to merge, because it doesn't make much sense to me.

aaron-collier:
No idea, just following the instructions and running the command found in access-update-scripts. I'm seeing the same size diff in another other sinopia repos as well.

jmartin-sul (approved):

PR #2913, opened 2021-08-13
Change to environment specific image building and deploy.

justinlittman:
## Why was this change made?
Deploy goodness


.circleci/config.yml
```diff
@@ -170,7 +237,7 @@ workflows:
               only:
                 - /.*/
             tags:
-              only:
+              ignore:
                 - /v.*/
```
jmartin-sul, line 241: should this be `/.*/`?  i assume we don't want to run `lint` for any tagging?  or do we want to run it for the non `/v.*/` tags?

.circleci/config.yml
```diff
@@ -180,15 +247,15 @@ workflows:
...
```
jmartin-sul, line 251: same question about ignoring all here

.circleci/config.yml
```diff
@@ -180,15 +247,15 @@ workflows:
...
       - integration:
           filters:
             branches:
               only:
                 - /.*/
             tags:
-              only:
+              ignore:
                 - /v.*/
```
jmartin-sul, line 259: same question about ignoring all here

jmartin-sul (approved):
merging despite the nitpicks, since they're pretty minor even if technically correct.

.circleci/config.yml
justinlittman: Yeah, this change was intentional. No need to run linting for version tags.

PR #2968, opened 2021-09-03
Remove x2js

justinlittman:
## Why was this change made?
Cruft


ndushay (approved):
versoSpoof is definitely cruft - I wrote that very early as a way to get profiles in the editor.

Issue #2965, opened 2021-09-03
Upgrade react-bootstrap-typeahead


justinlittman:
See https://www.npmjs.com/package/react-bootstrap-typeahead

PR #2977, opened 2021-09-05
Remove popper.

### LINKED ISSUE #2962: Upgrade popper
See https://www.npmjs.com/package/popper.js

justinlittman:
closes #2962

## Why was this change made?
Popper is a peer dependency of Bootstrap 4. With npm 7, peer dependencies are automatically installed, meaning that we no longer need to declare in `package.json`.


ndushay:
oh no - failing tests

justinlittman:
I think this is causing a problem with the type-ahead component. Try again after #2965

jcoyne:
Isn't popperjs a dependency of Bootstrap?

justinlittman:
@jcoyne See updated `Why was this change made?`

jcoyne:
@justinlittman that reality doesn't seem to be reflected in the package-lock.json.

PR #2978, opened 2021-09-05
switch to CodeClimate for test coverage

### LINKED ISSUE #2970: switch to CodeClimate for test coverage for consistency
- [x] get coverage working in CodeClimate
- [x] remove coveralls

ndushay:
BEWARE:  PR #2976 (use nanoid) seems to conflict with this branch.  I created PR #2979 to revert #2976 so this can be merged before the nanoid changes.

## Why was this change made?

Fixes #2970

Let's be consistent, and let's use fewer services - now we know how to get coverage stats in coveralls.

Note that code coverage won't be shown in badge until this PR is merged.

Added docker latest semver badge, too.

circleci builds.


package.json
```diff
@@ -43,10 +43,8 @@
     "eslint": "eslint --color -c .eslintrc.js --ext js,jsx --max-warnings 0 .",
     "eslint-fix": "eslint --color -c .eslintrc.js --ext js,jsx --fix .",
     "eslint-fix-single": "eslint --color -c .eslintrc.js --fix",
-    "jest-cov": "jest --coverage --colors --silent --runInBand",
-    "jest-ci": "jest  --colors --silent --ci --runInBand --coverage --reporters=default --reporters=jest-junit && cat ./coverage/lcov.info | coveralls",
     "start": "npx babel-node server.js",
-    "test": "jest --colors --silent --detectOpenHandles",
+    "test": "jest --colors --silent --runInBand",
```
justinlittman, line 47: Why this switch?

ndushay: detectOpenHandles is series, runInBand is parallel, as I recall.   runInBand seems to come along for the ride with npm test in other projects, I think???

justinlittman: See https://jestjs.io/docs/cli#--runinband

justinlittman (changes requested):
I don't think that changing `--detectOpenHandles` to `--runInBand` is correct.

package.json
ndushay: reverted this change.

ndushay: hah:  https://jestjs.io/docs/cli#--detectopenhandles   implies runInBand.  snort.

justinlittman (approved):

package.json
ndushay: so we need runInBand to keep memory needs down.  :-P

PR #2979, opened 2021-09-05
Revert "Replace shortid with nanoid."

ndushay:
Reverts LD4P/sinopia_editor#2976 - it causes tests to fail  for the code-climate-coverage branch, when it is rebased on this code.  See PR #2978


justinlittman:
Turns out this isn't necessary.

PR #3004, opened 2021-09-16
Allow selection of owning group and editing groups when saving. Upgra…

### LINKED ISSUE #2930: Update group choice modal
* Per new design
* User can select a group to be owner.
  * If user is a member of a single group, the group will be selected by default.
  * If user is a member of two or more groups, no groups will be selected by default.
* User can select zero more more editing groups:
  * This is from the list of all groups.

### LINKED ISSUE #2953: Upgrade @aws-amplify/auth
See https://www.npmjs.com/package/@aws-amplify/auth

justinlittman:
…des Ampify.

closes #2930

closes #2953

## Why was this change made?
So that the user can select permissions.

Unit

NA

### DIFF
--- a/__tests__/actionCreators/authenticate.test.js
+++ b/__tests__/actionCreators/authenticate.test.js
@@ -24,10 +24,10 @@ describe('authenticate', () => {
   describe('successful', () => {
     sinopiaApi.fetchUser = jest.fn().mockResolvedValue(userData)
     it('dispatches actions to add user', async () => {
-      Auth.currentAuthenticatedUser.mockResolvedValue({ username: 'havram', something: 'else' })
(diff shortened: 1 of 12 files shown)


justinlittman:
image

NancyL:
'Own" is a tricky word in this sphere--it suggests proprietary data with strings attached--though it's hard to find another term. Maybe "Who produced this?" "Who maintains this?" "What group is this in?" Not sure what should be emphasized. @michelleif any ideas?

ndushay (approved):
LGTM, modulo Nancy's concern about wording.

justinlittman:
Wording can be updated later if necessary.

PR #3008, opened 2021-09-16
Get sinopia groups from sinopia_api instead of config

### LINKED ISSUE #2929: Get groups from Sinopia API
Instead of configuration.

peetucket:
## Why was this change made?

Fixes #2929 - remove hardcoded groups from config and get from sinopia_api

- [x] update functions to handle new groups data structure coming from sinopia_api
- [x] Retrieving groups from Sinopia API should be added as a method to sinopiaApi.js.
- [x] The retrieval should be performed when the app is loaded, similar to how languages and exports are retrieved: https://github.com/LD4P/sinopia_editor/blob/main/src/components/App.jsx#L36-L37
- [x] The groups should be stored in state.

- Existing tests
- New test for fetching groups from API

### DIFF
--- a/__tests__/sinopiaApi.test.js
+++ b/__tests__/sinopiaApi.test.js
@@ -58,13 +58,21 @@ afterEach(() => {
 })
 
 describe('getGroups', () => {
+  const groups = [
+    { id: 'stanford', label: 'Stanford University' },
+    { id: 'cornell', label: 'Cornell University' },
+  ]
+
   it('retrieves list of groups', async () => {
+    // mocks call to Sinopia API for a resource
+    global.fetch = jest.fn().mockResolvedValue({
+      json: jest.fn().mockResolvedValue({ data: groups }),
+      ok: true,
+    })
+
     const result = await getGroups()
-    expect(result.data).toEqual(expect.arrayContaining([
-      { id: 'stanford', label: 'Stanford University' },
-      { id: 'cornell', label: 'Cornell University' },
-    ]))
-    expect(result.data.length).toBeGreaterThanOrEqual(25)
+    expect(result).toEqual(groups)
+    expect(global.fetch).toHaveBeenCalledWith('http://localhost:3000/groups', { method: 'GET', headers: { 'Content-Type': 'application/json' } })
   })
 })
 
--- a/__tests__/sinopiaApi.test.js
+++ b/__tests__/sinopiaApi.test.js
@@ -1,6 +1,6 @@
 // Copyright 2020 Stanford University see LICENSE for license
 import {
-  fetchResource, postResource, putResource,
+  fetchResource, postResource, putResource, getGroups,
   postMarc, getMarcJob, getMarc, fetchUser, putUserHistory,
 } from 'sinopiaApi'
 import { selectFullSubject } from 'selectors/resources'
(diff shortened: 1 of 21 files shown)


justinlittman:
A few suggestions:
* Retrieving groups from Sinopia API should be added as a method to `sinopiaApi.js`.
* The retrieval should be performed when the app is loaded, similar to how languages and exports are retrieved: https://github.com/LD4P/sinopia_editor/blob/main/src/components/App.jsx#L36-L37
* The groups should be stored in state.

__tests__/sinopiaApi.test.js
```diff
@@ -58,6 +58,17 @@ afterEach(() => {
...
+  it('retrieves list of groups', async () => {
+    const result = await getGroups()
+    expect(result.data).toEqual(expect.arrayContaining([
+      { id: 'stanford', label: 'Stanford University' },
+      { id: 'cornell', label: 'Cornell University' },
+    ]))
+    expect(result.data.length).toBeGreaterThanOrEqual(25)
+  })
+})
+
```
peetucket, line 71: this test fails, showing `result` as undefined, indicating that the `getGroups` function in `sinopiaApi` isn't working, and I don't know why ....

this then causes test failures throughout the stack

src/sinopiaApi.js
```diff
@@ -40,6 +40,18 @@ export const fetchResource = (uri, isTemplate) => {
...
+  fetch(`${Config.sinopiaApiBase}/groups`, {
+    method: 'GET',
+    headers: {
+      'Content-Type': 'application/json',
+    },
+  })
+    .then((resp) => checkResp(resp)
+      .then(() => resp.json()))
+}
+
```
peetucket, line 54: Not sure why this isn't working.  It is very similar to this function: https://github.com/LD4P/sinopia_editor/blob/main/src/sinopiaApi.js#L113-L121

peetucket:
Looks like `src/components/editor/GroupChoiceModal.jsx` changed significantly from when I started editing it...

I may need to confer on how to deal with this conflict now

src/components/editor/GroupChoiceModal.jsx
```diff
@@ -4,31 +4,32 @@ import React, { useState } from 'react'
...
   const [ownerGroupId, setOwnerGroupId] = useState(userGroupIds[0])
   const [editGroupValues, setEditGroupValues] = useState([])
-  const groupMap = Config.groupsInSinopia
   const show = modalType === 'GroupChoiceModal'
   const dispatch = useDispatch()
 
   // TODO: Handle owner group in edit groups. Should be selected and disabled.
 
-  const ownerGroupOptions = userGroupIds.map((groupId) => <option key={groupId} value={ groupId }>{ groupMap[groupId] || groupId }</option>)
+  const ownerGroupOptions = userGroupIds.map((groupId) => <option key={groupId} value={ groupId }>{ groupNameFromGroup(groupId) }</option>)
```
peetucket, line 29: There is a function that looks up the group name from the group (though it defaults to returning "Unknown" if not found instead of just the group id itself)

src/reducers/groups.js
```diff
@@ -0,0 +1,9 @@
+export const groupsReceived = (state, action) => ({ ...state, groups: action.payload, groupMap: createGroupMap(action.payload) })
+
+export const createGroupMap = (groupList) => {
+  const groupMap = {}
+  groupList.forEach((group) => groupMap[group.id] = group.label)
+  return groupMap
+}
+
+export const noop = () => {}
```
ndushay, line 9: so this exact same code is in `src/reducers` AND in `src/actions` ?

ndushay: do all files need that license comment header?

justinlittman (approved):
GOAL!

src/reducers/groups.js
peetucket: all of the plumbing is needed between the reducers and the actions (still getting it straight myself) ...

and indeed the license is not consistent applied across the project, i'm sure I missed some in my new files

PR #3010, opened 2021-09-17
Update groups

justinlittman:
## Why was this change made?
User can manage permissions for a resource.

Unit, local


justinlittman:
When user can't change permissions:
image

jcoyne (approved):

PR #3028, opened 2021-09-21
Prettier

justinlittman:
## Why was this change made?
More consistent code.

Unit


ndushay (approved):
I see no value in reviewing this in any sort of thorough way.  Merging.

PR #3034, opened 2021-09-22
show marc button on preview modal

### LINKED ISSUE #3015: Add MARC button to Preview
Add:
image

to:
image

peetucket:
## Why was this change made?

Fixes #3015 - show MARC button on the preview modal

Note: this uses the `<MarcButton>` component, which means the MARC button will only be shown for resources with templates of type `http://id.loc.gov/ontologies/bibframe/Instance`.  

Localhost browser (at least verifying the button is shown).  Unclear to verify button operation works as expected.  Should we add another test?

### DIFF
--- a/src/components/editor/actions/MarcButton.jsx
+++ b/src/components/editor/actions/MarcButton.jsx
@@ -85,7 +85,7 @@ const MarcButton = () => {
   const dropDownItemBtnClasses = ["btn", "btn-secondary", "dropdown-item"]
 
   return (
-    <div className="btn-group dropstart">
+    <div className="view-resource-buttons btn-group dropstart">
       <button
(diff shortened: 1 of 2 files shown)


src/components/editor/actions/MarcButton.jsx
```diff
@@ -84,7 +84,7 @@ const MarcButton = () => {
   }
 
   return (
-    <div className="btn-group dropstart">
+    <div className="view-resource-buttons btn-group dropstart">
```
peetucket, line 87: this extra class helps with the spacing of the button in both cases

ndushay (approved):
this LGTM;  I don't know if adding a test is warranted or not.

PR #3041, opened 2021-09-23
show warning to a logged in user with no groups

### LINKED ISSUE #3003: Notify user when member of no groups
* To indicate that Editor is read-only.
* To provide directions on how to get added to a group.

peetucket:
## Why was this change made?

Fixes #3003 - show a warning to a user on the home page when they are not in any groups

- Existing unit tests, and added a test for the new component
- Localhost browser (though had to temporarily fake it in the code to see the warning since I'm not sure how to remove myself from all groups or log in with a user that has no groups)

**When a logged in user is in at least one group:** (behavior unchanged)

Screen Shot 2021-09-23 at 2 59 14 PM

**When a logged in user is NOT in at least one group:** (new alert)

Screen Shot 2021-09-23 at 2 58 57 PM

### DIFF
--- a/src/components/home/UserNotifications.jsx
+++ b/src/components/home/UserNotifications.jsx
@@ -1,0 +1,32 @@
+// Copyright 2021 Stanford University see LICENSE for license
+
+import React from "react"
+import { useSelector } from "react-redux"
+import {
+  hasUser as hasUserSelector,
+  selectGroups,
+} from "selectors/authenticate"
+
+const UserNotifications = () => {
+  const hasUser = useSelector((state) => hasUserSelector(state))
+  const userGroups = useSelector((state) => selectGroups(state))
+
+  if (!hasUser) return null // nothing to show if not logged in
+  if (userGroups.length) return null // nothing to show if the user is logged in but is in at least one group
+
+  if (!userGroups.length) {
+    // show a message if the user is not in any groups
+    return (
+      <div className="alert alert-warning">
+        <strong>Note:</strong> Before you can create new resources or edit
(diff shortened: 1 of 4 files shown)


src/components/home/UserNotifications.jsx
```diff
@@ -0,0 +1,32 @@
...
+  hasUser as hasUserSelector,
+  selectGroups,
+} from "selectors/authenticate"
+
+const UserNotifications = () => {
+  const hasUser = useSelector((state) => hasUserSelector(state))
+  const userGroups = useSelector((state) => selectGroups(state))
+
+  if (!hasUser) return null // nothing to show if not logged in
+  if (userGroups.length) return null // nothing to show if the user is logged in but is in at least one group
```
peetucket, line 15: this gives us the easy ability to extend this component later if we wanted to show something here instead, like which groups the user is in, etc.

src/selectors/authenticate.js
```diff
@@ -2,4 +2,4 @@ export const hasUser = (state) => !!state.authenticate.user
 
 export const selectUser = (state) => state.authenticate.user
 
-export const selectGroups = (state) => state.authenticate.user.groups
+export const selectGroups = (state) => state.authenticate.user?.groups
```
peetucket, line 5: this allows our new component to work since it both uses the group selector and it runs on the home page, which can be rendered if the user is not logged in; this just adds a safe navigation operator, so the `selectGroups` will just get returned as `undefined` instead of throwing a "user is nil" type exception

justinlittman (approved):

PR #3042, opened 2021-09-24
set line length to 120 because it isn't 1982

ndushay:
## Why was this change made?

So prettier doesn't keep reformating lines that are of perfectly reasonable length.


ndushay:
@peetucket, @justinlittman - making you aware of this;  we may want to apply this change to the other repos using prettier ...

ndushay:
hrmm - nm - it doesn't seem to work.  grrr.

justinlittman:
Determined that longer line lengths reduce readability.

PR #3043, opened 2021-09-24
set property as changed when updating language to enable save button

### LINKED ISSUE #3037: unable to save changes to template when adding default language specification
on DEV, using template completely:new

field Example 1 has a default value, but no language specified

I edit the template to specify the language of the default value, and successfully choose a language:

but the Save button is still grayed out so I can't save my change

peetucket:
## Why was this change made?

Fixes #3037 - enable the save button when a language is changed.  Similar fix to what happened in #2991

Localhost browser

### DIFF
--- a/src/reducers/languages.js
+++ b/src/reducers/languages.js
@@ -1,15 +1,22 @@
 // Copyright 2020 Stanford University see LICENSE for license
+import { setSubjectChanged } from "reducers/resources"
 
-export const setLanguage = (state, action) => ({
-  ...state,
-  values: {
-    ...state.values,
-    [action.payload.valueKey]: {
-      ...state.values[action.payload.valueKey],
-      lang: action.payload.lang,
+export const setLanguage = (state, action) => {
+  const valueKey = action.payload.valueKey
+  const value = state.values[valueKey]
+  const property = state.properties[value.propertyKey]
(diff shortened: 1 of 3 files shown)


src/reducers/languages.js
```diff
@@ -1,15 +1,22 @@
...
-  values: {
-    ...state.values,
-    [action.payload.valueKey]: {
-      ...state.values[action.payload.valueKey],
-      lang: action.payload.lang,
+export const setLanguage = (state, action) => {
+  const valueKey = action.payload.valueKey
+  const value = state.values[valueKey]
+  const property = state.properties[value.propertyKey]
+  const newState = setSubjectChanged(state, property.subjectKey, true)
```
peetucket, line 8: this does the magic of causing the save button to enabled

src/reducers/resources.js
```diff
@@ -612,7 +612,7 @@ const stateWithNewValue = (state, valueKey) => ({
   },
 })
 
-const setSubjectChanged = (state, subjectKey, changed) => {
+export const setSubjectChanged = (state, subjectKey, changed) => {
```
peetucket, line 615: we are now going to import and call this function from `languages.js`

justinlittman (approved):

PR #3061, opened 2021-09-28
Refactor Input Lookup component

justinlittman:
## Why was this change made?
Usability

Unit


justinlittman:
image

.circleci/config.yml
```diff
@@ -23,7 +23,7 @@ references:
     main_only: &main_only_filter
       branches:
         only:
-          - main
+          - t3012-lookup
```
ndushay, line 26: oops.

src/components/editor/inputs/InputLookupValue.jsx
```diff
@@ -0,0 +1,220 @@
...
+          property={value.property}
+          propertyTemplate={propertyTemplate}
+          show={showLookup}
+          hideLookup={hideLookup}
+          handleUpdateLiteral={handleUpdateLiteral}
+          handleUpdateURI={handleUpdateURI}
+          query={query}
+        />
+      </div>
+    </React.Fragment>
```
ndushay, line 209: this is long.  I assume you chose to leave it this way rather than breaking it up into pieces?  Not sure what's better but maybe well named pieces would be helpful?

src/components/editor/inputs/LookupTabs.jsx
```diff
@@ -48,44 +48,55 @@ const LookupTabs = (props) => {
     )
   }
 
+  const handleTabClick = (event, authorityUri) => {
+    setCurrentAuthorityUri(authorityUri)
+    event.preventDefault()
+  }
+
+  /* eslint-disable jsx-a11y/anchor-is-valid */
```
ndushay, line 56: does this eslint rule need to be re-enabled?

src/utilities/valueFactory.js
```diff
@@ -16,17 +17,31 @@ const newValue = (
...
+
+export const newBlankLiteralValue = (property) =>
+  newValue(
+    property,
+    null,
+    defaultLanguageId,
+    null,
+    null,
+    null,
+    "InputLiteralValue"
```
ndushay, line 34: that's a lot of args to remember in an ordered way.  are keyword args possible in javascript?

ndushay (changes requested):
two possible whoops:  one in circleci config.yml, one with an eslint rule that may need to be turned back on.

And few thoughts, but not require changes.

.circleci/config.yml
justinlittman: Thanks for catching!

src/utilities/valueFactory.js
justinlittman: I agree this isn't awesome. However, since newValue is only used within `valueFactory` and the methods that are exported have fewer args, this seems less horrible.

src/components/editor/inputs/LookupTabs.jsx
justinlittman: Fixing.

src/components/editor/inputs/InputLookupValue.jsx
justinlittman: I agree. Let me componentize some of this.

ndushay (approved):
I like the refactoring!

PR #3063, opened 2021-09-29
Refactor input lists

justinlittman:
## Why was this change made?
Usability

Unit


src/hooks/useDiacriticsLegacy.js
```diff
@@ -1,88 +0,0 @@
...
-    handleClickDiacritics,
-    currentContent,
-    setCurrentContent,
-    currentPosition,
-    setCurrentPosition,
-    showDiacritics,
-  }
-}
-
-export default useDiacritics
```
ndushay, line 88: how confident are you that all of the uses of this have been removed?

ndushay (approved):

src/hooks/useDiacriticsLegacy.js
justinlittman: Very. (Until I'm wrong.)

PR #3065, opened 2021-09-29
ensure language radio buttons are selected correctly

### LINKED ISSUE #3036: when setting language for default value in template creation, radio button is not getting highlighted
on DEV, using template completely:new

field Example 1 has a default value, but no language specified

I edit the template to specify the language of the default value, and in the language selection box, the top radio button for "Select language" never gets highlighted when I search for and then select a language.

peetucket:
## Why was this change made?

Fixes #3036 -- set the radio buttons correctly based on the language selection

Localhost

### DIFF
--- a/__tests__/feature/editing/editingLiteral.test.js
+++ b/__tests__/feature/editing/editingLiteral.test.js
@@ -143,6 +143,15 @@ describe("editing a literal property", () => {
     // Using getByRole here and below because it limits to the visible modal.
     screen.getByRole("heading", { name: "Languages" })
 
+    const radioButtons = await screen.findAllByRole("radio")
+
+    let checkedRadioButtons = radioButtons.filter((el) => el.checked)
+    let unCheckedRadioButtons = radioButtons.filter((el) => !el.checked)
+
+    // the correct "language selected" radio button is pre-checked
+    expect(checkedRadioButtons[0].value).toEqual("present")
+    expect(unCheckedRadioButtons[0].value).toEqual("absent")
+
     const langInput = screen.getByTestId("langComponent-foo")
 
     fireEvent.click(langInput)
@@ -150,6 +159,14 @@ describe("editing a literal property", () => {
     fireEvent.click(
       screen.getByText("Tai languages", { selector: ".rbt-highlight-text" })
     )
+
+    checkedRadioButtons = radioButtons.filter((el) => el.checked)
+    unCheckedRadioButtons = radioButtons.filter((el) => !el.checked)
+
+    // the correct "language selected" radio button is still checked
+    expect(checkedRadioButtons[0].value).toEqual("present")
+    expect(unCheckedRadioButtons[0].value).toEqual("absent")
+
     fireEvent.click(screen.getByRole("button", { name: "Submit" }))
 
     await waitFor(() =>
@@ -181,9 +198,25 @@ describe("editing a literal property", () => {
     fireEvent.click(langBtn)
     screen.getByRole("heading", { name: "Languages" })
 
+    const radioButtons = await screen.findAllByRole("radio")
+
+    let checkedRadioButtons = radioButtons.filter((el) => el.checked)
+    let unCheckedRadioButtons = radioButtons.filter((el) => !el.checked)
+
+    // the correct "language selected" radio button is pre-checked
+    expect(checkedRadioButtons[0].value).toEqual("present")
+    expect(unCheckedRadioButtons[0].value).toEqual("absent")
+
     // Using testid here because there are multiple modals.
     fireEvent.click(screen.getByTestId("noLangRadio-foo"))
     fireEvent.click(screen.getByTestId("submit-foo"))
+
+    checkedRadioButtons = radioButtons.filter((el) => el.checked)
+    unCheckedRadioButtons = radioButtons.filter((el) => !el.checked)
+
+    // the correct "no language selected" radio button is now checked
+    expect(checkedRadioButtons[0].value).toEqual("absent")
+    expect(unCheckedRadioButtons[0].value).toEqual("present")
 
     await waitFor(() =>
       expect(
--- a/src/components/editor/inputs/InputLang.jsx
+++ b/src/components/editor/inputs/InputLang.jsx
@@ -17,35 +17,45 @@ import { selectLanguages, hasLanguages } from "selectors/languages"
  * See ISO 639 for the list of registered language codes
  */
 const InputLang = (props) => {
+  const noLangSelected = "absent" // the values of the radio buttons for when a language is selected or not
+  const langSelected = "present"
+
   const [lang, setLang] = useState(props.lang)
   const [submitEnabled, setSubmitEnabled] = useState(true)
   const [radioButtonValue, setRadioButtonValue] = useState(
-    props.lang === null ? "absent" : "present"
+    props.lang === null ? noLangSelected : langSelected
   )
 
   const classes = ["modal", "fade"]
   let display = "none"
   const alertClasses = ["alert", "alert-danger"]
-  if (submitEnabled) alertClasses.push("hidden")
+  if (submitEnabled) alertClasses.push("hidden") // hide the alert if the submit button is enabled
 
   if (props.show) {
     classes.push("show")
     display = "block"
   }
 
+  // This function is called when a user picks a language in the type ahead component:
+  //  (1) record the radio button checked in state
+  //  (2) if a language was selected, set it in state to this value and then select the correct radio button
+  //  (3) if no language was selected, set the language in state to null and then select the correct radio button
   const selectLanguage = (selected) => {
     setSubmitEnabled(true)
     if (selected.length === 1) {
       setLang(selected[0].id)
-      setRadioButtonValue("present")
+      setRadioButtonValue(langSelected)
     } else {
       setLang(null)
-      setRadioButtonValue("absent")
+      setRadioButtonValue(noLangSelected)
     }
   }
 
+  // This function is called when a user clicks one of the radio buttons:
+  //  (1) select the correct radio button
+  //  (2) if they selected the "no language" radio button, set language in state to null
   const handleLanguageRadio = (event) => {
-    if (event.target.value === "absent") setLang(null)
+    if (event.target.value === noLangSelected) setLang(null)
     setRadioButtonValue(event.target.value)
   }
 
@@ -55,7 +65,7 @@ const InputLang = (props) => {
   }
 
   const handleLangSubmit = (event) => {
-    if (radioButtonValue === "present" && lang === null) {
+    if (radioButtonValue === langSelected && lang === null) {
       setSubmitEnabled(false)
       return false
     }
@@ -75,15 +85,15 @@ const InputLang = (props) => {
               <input
                 type="radio"
                 className="form-check-input"
-                id="present"
-                value="present"
-                checked={radioButtonValue === "present"}
+                id={langSelected}
+                value={langSelected}
+                checked={radioButtonValue === langSelected}
                 onChange={handleLanguageRadio}
               />
-              <label className="form-check-label" htmlFor="present">
+              <label className="form-check-label" htmlFor={langSelected}>
                 Select language for {props.textValue}
                 <Typeahead
-                  disabled={radioButtonValue === "absent"}
+                  disabled={radioButtonValue === noLangSelected}
                   onChange={selectLanguage}
                   isLoading={props.loading}
                   options={props.options}
@@ -105,8 +115,8 @@ const InputLang = (props) => {
                 className="form-check-input"
                 id={`noLangRadio-${props.textValue}`}
                 data-testid={`noLangRadio-${props.textValue}`}
-                value="absent"
-                checked={radioButtonValue === "absent"}
+                value={noLangSelected}
+                checked={radioButtonValue === noLangSelected}
                 onChange={handleLanguageRadio}
               />
               <label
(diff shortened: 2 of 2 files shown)


src/components/editor/inputs/InputLang.jsx
```diff
@@ -17,10 +17,10 @@ import { selectLanguages, hasLanguages } from "selectors/languages"
...
  */
 const InputLang = (props) => {
-  const [lang, setLang] = useState(props.lang || "")
-  const langPresent = typeof lang !== "undefined"
-  const [languageSelectorEnabled, setLanguageSelectorEnabled] =
-    useState(langPresent)
+  const [lang, setLang] = useState(props.lang)
+  const [radioButtonValue, setRadioButtonValue] = useState(
+    props.lang === null ? "absent" : "present"
+  )
```
peetucket, line 23: this lets us keep track of the selected radio button separately from the existence of the `lang` value and makes the component code below easier to understand ... the default value is based on the existence of a language selection coming in though

src/components/editor/inputs/InputLang.jsx
```diff
@@ -33,16 +33,16 @@ const InputLang = (props) => {
   const selectLanguage = (selected) => {
     if (selected.length === 1) {
       setLang(selected[0].id)
+      setRadioButtonValue("present")
     } else {
-      setLang(undefined)
+      setLang(null)
+      setRadioButtonValue("absent")
     }
```
peetucket, line 40: this method is called when a user picks a language in the type ahead compontent: 

1. we use `null` to indicate no language is selected, `lang` can never be undefined (see line 20 above)
2. set the currently selected radio button appropriately

src/components/editor/inputs/InputLang.jsx
```diff
@@ -33,16 +33,16 @@ const InputLang = (props) => {
...
 
-  const enableLanguageSelector = () => setLanguageSelectorEnabled(true)
-
-  const disableLanguageSelector = () => {
-    setLanguageSelectorEnabled(false)
-    setLang(undefined)
+  const handleLanguageRadio = (event) => {
+    if (event.target.value === "absent") setLang(null)
+    setRadioButtonValue(event.target.value)
   }
```
peetucket, line 46: this function is called when a user clicks one of the radio buttons:

1. record the button checked in state
2. if they selected "no language", we will clear the language in state

src/components/editor/inputs/InputLang.jsx
```diff
@@ -67,16 +71,15 @@ const InputLang = (props) => {
...
                 type="radio"
                 className="form-check-input"
-                name="lang"
                 id="present"
                 value="present"
-                defaultChecked={langPresent}
-                onChange={enableLanguageSelector}
+                checked={radioButtonValue === "present"}
+                onChange={handleLanguageRadio}
               />
```
peetucket, line 78: 1. we now control if the radio button is checked or not based on the current value in state
2. our new function to handle when the user clicks the radio button

src/components/editor/inputs/InputLang.jsx
```diff
@@ -96,12 +99,10 @@ const InputLang = (props) => {
...
-                name="lang"
-                id={`noLangRadio-${props.textValue}`}
-                data-testid={`noLangRadio-${props.textValue}`}
+                id="absent"
                 value="absent"
-                defaultChecked={!langPresent}
-                onChange={disableLanguageSelector}
+                checked={radioButtonValue === "absent"}
+                onChange={handleLanguageRadio}
               />
```
peetucket, line 106: 1. we now control if the radio button is checked or not based on the current value in state
2. our new function to handle when the user clicks the radio button

src/components/editor/inputs/InputLang.jsx
```diff
@@ -96,12 +99,10 @@ const InputLang = (props) => {
               <input
                 type="radio"
                 className="form-check-input"
-                name="lang"
-                id={`noLangRadio-${props.textValue}`}
-                data-testid={`noLangRadio-${props.textValue}`}
+                id="absent"
```
peetucket, line 102: 1. get rid of the `name="lang"` to prevent the browser from trying to set the value of the radio button based on the grouping (we set based on our value in state)
2. i don't think we need an `id` or `data-testid` here, just get rid of for simplicity

src/components/editor/inputs/InputLang.jsx
```diff
@@ -67,16 +71,15 @@ const InputLang = (props) => {
               <input
                 type="radio"
                 className="form-check-input"
-                name="lang"
```
peetucket, line 70: get rid of the `name="lang"` to prevent the browser from trying to set the value of the radio button based on the grouping (we set based on our value in state)

src/components/editor/inputs/InputLang.jsx
```diff
@@ -96,12 +99,11 @@ const InputLang = (props) => {
...
```
peetucket, line 99: get rid of the name="lang" to prevent the browser from trying to set the value of the radio button based on the grouping (we set based on our value in state)

src/components/editor/inputs/InputLang.jsx
```diff
@@ -51,6 +51,10 @@ const InputLang = (props) => {
   }
 
   const handleLangSubmit = (event) => {
+    if (radioButtonValue === "present" && lang === null) {
+      alert("Please select a valid language.")
+      return false
+    }
     close(event)
```
peetucket, line 58: Previously the user could click the "Select language" radio button but then not choose any language before clicking `Submit`.  This gives them an error if they try to do that.

src/components/editor/inputs/InputLang.jsx
```diff
@@ -51,6 +51,10 @@ const InputLang = (props) => {
   }
 
   const handleLangSubmit = (event) => {
+    if (radioButtonValue === "present" && lang === null) {
+      alert("Please select a valid language.")
```
justinlittman, line 55: No alerts. We should disable the Submit button instead when in bad state.

justinlittman (changes requested):

src/components/editor/inputs/InputLang.jsx
peetucket: OK, though this may get pretty tricky to detect the bad state (I think you'd need to be constantly checking as the user types in the text box).  May not be worth the trouble if it hasn't come up before.

peetucket: actually came up with an alternate way of doing this (disable submit button and show an inline alert), required some extra state though

src/components/editor/inputs/InputLang.jsx
```diff
@@ -112,13 +118,17 @@ const InputLang = (props) => {
             </div>
           </div>
           <div className="modal-footer">
+            <span className={alertClasses.join(" ")}>
+              Please select a valid language.
+            </span>
             <button className="btn btn-link" onClick={close}>
```
peetucket, line 124: this alert will be conditionally shown if the user tries to submit the form in a bad state (we also disable the submit button until they correct the issue)

peetucket:
Alert triggered if you select the "language" radio button but don't put anything in there

Screen Shot 2021-09-29 at 3 46 55 PM

src/components/editor/inputs/InputLang.jsx
ndushay: Maybe this could be a comment for the function??  it's pretty intricate.

src/components/editor/inputs/InputLang.jsx
ndushay: ditto on making it a comment for the function

src/components/editor/inputs/InputLang.jsx
ndushay: excellent.

src/components/editor/inputs/InputLang.jsx
ndushay: also excellent.

src/components/editor/inputs/InputLang.jsx
ndushay: MUCH clearer

src/components/editor/inputs/InputLang.jsx
ndushay: ah!  also nicer.

src/components/editor/inputs/InputLang.jsx
ndushay: much much clearer

src/components/editor/inputs/InputLang.jsx
ndushay: this makes sense.  I'm wondering if there is a better name for radioButtonValue so it can be clear as a boolean.  `languageIsSelected` ?  `langIsSelectedRadioButton` ?  `langRadioSelector` ?   dunno if it's worth it, but "present" and "absent" are unusual values for a boolean switch.

ndushay (commented):
Given how hard it was to figure this out ... it makes me want to have tests.  I'll leave it to @justinlittman to decide if tests are worth it.

One suggestion you can take or leave.

src/components/editor/inputs/InputLang.jsx
peetucket: yeah, present and absent are unusual choices, maybe renaming those values to something like "langSelected" (instead of "present") and "noLangSelected" (instead of "absent") will make it clear what is happening -- note this state value is not a boolean, it's the string value of the radio button we will use below to decide which one to check, which is why I called it "radioButtonValue"

peetucket: created constants for those radio button values and then used the better named constants through the code, also makes it more DRY and avoids a case where a typo in a string radio button value that is sprinkled throughout causes things to break

src/components/editor/inputs/InputLang.jsx
peetucket: added inline function comments

src/components/editor/inputs/InputLang.jsx
```diff
@@ -112,13 +128,17 @@ const InputLang = (props) => {
...
               Cancel
             </button>
             <button
               className="btn btn-primary"
               onClick={handleLangSubmit}
               data-testid={`submit-${props.textValue}`}
+              disabled={!submitEnabled}
```
peetucket, line 141: allow us to disable the submit button if the user is in a bad state, this will also trigger the alert above to show

peetucket:
There is one edge case not dealt with elegantly ... since the modal persists even after you close it, you can do this:

1. Pick a field with "no language specified" and click to get the modal.
2. Click the "Select a language" radio button.
2. Pick a language, e.g. type "En" and then select "English".
3. Click Submit.
4. Open the same modal again by selecting "English" next to the same field.
5. Click the "No language selected" radio button.  Note that the type ahead language field is disabled but the value still shows "English" (confusing, but not fatal).
6. Click Submit.
7. The language is now back to unspecified .. this is good and everything worked.
8. But...now open the same language modal again and note that even though the "No language selected" radio button is correctly pre-selected, the "English" language is still shown in the disabled type ahead field.  
9. Now click the "Select language" radio button and then hit "Submit" without doing anything else.  You'll get the alert claiming you need to select a language.  A bit confusing.

I think this stems from the fact that the state correctly has the `lang` value set in all cases, but the type ahead component value is there from the previous entry.  This is a pretty convoluted path and doesn't cause data errors, just a bit confusing for the user, so not sure if we want to spend any time on it or not.

src/components/editor/inputs/InputLang.jsx
```diff
@@ -17,32 +17,46 @@ import { selectLanguages, hasLanguages } from "selectors/languages"
...
+
+  const [lang, setLang] = useState(props.lang)
+  const [submitEnabled, setSubmitEnabled] = useState(true)
+  const [radioButtonValue, setRadioButtonValue] = useState(
+    props.lang === null ? noLangSelected : langSelected
+  )
 
   const classes = ["modal", "fade"]
   let display = "none"
+  const alertClasses = ["alert", "alert-danger"]
```
justinlittman, line 31: I'd prefer to not use an alert here. It is not consistent with the rest of the app. The Submit button should be disabled when the form is in an invalid state (i.e., when the user has selected to provide a language, but has not done so.).

justinlittman (changes requested):
Yeah, so this does need a feature test.

Here is a relevant existing test: https://github.com/LD4P/sinopia_editor/blob/main/__tests__/feature/editing/editingLiteral.test.js#L124-L161

For testing more variations, I would create a separate feature test based on ^^.

src/components/editor/inputs/InputLang.jsx
peetucket: Got it, I'll ditch the alerts but leave the submit button enabling/disabling.

peetucket:
There were two existing tests that I just added some expectations to (looking at the radio buttons that were selected or not to verify).  Do you think this is sufficient?

__tests__/feature/editing/editingLiteral.test.js
```diff
@@ -143,13 +143,33 @@ describe("editing a literal property", () => {
     // Using getByRole here and below because it limits to the visible modal.
     screen.getByRole("heading", { name: "Languages" })
 
+    let radioButtons
+    await waitFor(() => {
```
justinlittman, line 147: You shouldn't need `waitFor` here. Try `const radioButtons = await screen.findAllByRole("radio")`

justinlittman (approved):
Approve pending consideration of comment.

__tests__/feature/editing/editingLiteral.test.js
peetucket: you are correct, i switched to just `await` and it worked

PR #3067, opened 2021-09-30
[HOLD] Adds ability to request transfers.

### LINKED ISSUE #2936: Allow user to request record transfer
When a resource if a BF:instance, for each group that the user is a member of that allows ILS transfer, render a "Transfer to ILS" button. When the user clicks the button, a transfer request is sent to the Sinopia API.

* Similarly, MARC Button only renders when BF:instance. See https://github.com/LD4P/sinopia_editor/blob/main/src/components/editor/actions/MarcButton.jsx#L18
* Groups that support ILS transfer will be provided by configuration.
* Per design for "Transfer to ILS" button (or whatever the design specifies).
* See https://github.com/LD4P/sinopia_api/issues/145 for transfer endpoint.
* "Transfer to ILS" button should be placed on both the Editor page and the Preview page.

justinlittman:
closes #2936

## Why was this change made?
Closing the loop?

Unit


justinlittman:
image

src/components/editor/Editor.jsx
```diff
@@ -23,6 +23,10 @@ import { newResourceErrorKey } from "../templates/SinopiaResourceTemplates"
 export const resourceEditErrorKey = (resourceKey) =>
   `resourceedit-${resourceKey}`
 
+// Error key for warning that occur while editing a resource.
```
ndushay, line 26: warningSSSS ?

__tests__/sinopiaApi.test.js
```diff
@@ -401,3 +402,25 @@ describe("putUserHistory", () => {
     )
   })
 })
+
+describe("postTransfer", () => {
+  describe("success", () => {
```
ndushay, line 407: I'm only looking at the github diffs of this file -- would we want a test for failure?

src/components/editor/Editor.jsx
justinlittman: Fixed.

__tests__/sinopiaApi.test.js
justinlittman: No, it intentionally doesn't handle a failure. (The failure is passed on to the calling code.)

PR #3090, opened 2021-10-01
use icon for creating resource instead of linking resource template name

### LINKED ISSUE #2853: Links to enter data vs editing the template are not clear
**Describe the bug**
Today I had to "repair" a pcc template because someone was trying to enter data using the template-creation mode rather than the data entry mode. This is not so strange, in that the template-creation button just says "edit", which to a newcomer could easily mean "enter your data" (many cataloging templates have this, including our own ILS Symphony--cataloging is done in the "modify" template). And to enter data, you click on the name of template, which is not necessarily obvious to a newcomer. It just is really tempting to click on that "edit" button.

We need to make it clear what to click to enter metadata. Either add another button on the right or put in some note on the screen such as "Click on template name to enter metadata". And the edit button should probably be more explicit, as in "edit template" to specify it is not for entering metadata.

peetucket:
## Why was this change made?

Fixes #2853:
- unlink resource template name as a way of creating a resource
- add the react-bootstrap icon module since we are now using it
- label column on dashboard with "Actions"
- add a new icon for create a resource to the Actions column
- change edit template icon to better match style shown in mockup
- add/adjust hover text for all icons
- fix the tests so they can find the new linking style of creating resources

Example for how it looks now:
Screen Shot 2021-10-01 at 3 45 51 PM

Existing tests

### DIFF
--- a/__tests__/feature/dashboard.test.js
+++ b/__tests__/feature/dashboard.test.js
@@ -28,7 +28,7 @@ describe("viewing the dashboard", () => {
       await screen.findByText(/Uber template1/)
 
       // Click the resource template
-      fireEvent.click(screen.getByText("Uber template1", { selector: "a" }))
+      fireEvent.click(screen.getByTestId("createResource-Uber template1"))
       await waitFor(() =>
         expect(
           screen.getAllByText("Uber template1", { selector: "h3" })
--- a/__tests__/feature/dashboard.test.js
+++ b/__tests__/feature/dashboard.test.js
@@ -28,7 +28,7 @@ describe("viewing the dashboard", () => {
       await screen.findByText(/Uber template1/)
 
       // Click the resource template
-      fireEvent.click(screen.getByTestId("createResource-Uber template1"))
+      fireEvent.click(screen.getByTitle("Create resource for Uber template1"))
       await waitFor(() =>
         expect(
           screen.getAllByText("Uber template1", { selector: "h3" })
--- a/package.json
+++ b/package.json
@@ -85,6 +85,7 @@
     "process": "^0.11.10",
     "rdf-ext": "^1.3.5",
     "react": "^16.14.0",
+    "react-bootstrap-icons": "^1.5.0",
     "react-bootstrap-typeahead": "^5.2.0",
     "react-dom": "^16.14.0",
     "react-helmet-async": "^1.1.2",
@@ -192,7 +193,12 @@
       "lcov"
     ],
     "reporters": [
-      ["jest-summarizing-reporter", {"diffs": true}]
+      [
+        "jest-summarizing-reporter",
+        {
+          "diffs": true
+        }
+      ]
     ]
   },
   "optionalDependencies": {
--- a/src/components/templates/ResourceTemplateRow.jsx
+++ b/src/components/templates/ResourceTemplateRow.jsx
@@ -38,10 +38,10 @@ const ResourceTemplateRow = ({ row, handleClick, handleCopy, handleEdit }) => {
         <div className="btn-group" role="group" aria-label="Result Actions">
           {canCreate && (
             <Link
-              data-testid={`createResource-${row.resourceLabel}`}
               to={{ pathname: "/editor", state: {} }}
               onClick={(e) => handleClick(row.id, e)}
-              title="Create resource"
+              title={`Create resource for ${row.resourceLabel}`}
+              aria-label={`Create resource for ${row.resourceLabel}`}
             >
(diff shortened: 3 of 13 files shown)


src/components/templates/ResourceTemplateRow.jsx
```diff
@@ -20,16 +21,7 @@ const ResourceTemplateRow = ({ row, handleClick, handleCopy, handleEdit }) => {
...
-          <Link
-            to={{ pathname: "/editor", state: {} }}
-            onClick={(e) => handleClick(row.id, e)}
-          >
-            {row.resourceLabel}
-          </Link>
-        ) : (
-          <span>{row.resourceLabel}</span>
-        )}
+        <span>{row.resourceLabel}</span>
```
peetucket, line 24: Unlink the label in the first column (this used to be the way to create a resource)

src/components/templates/ResourceTemplateRow.jsx
```diff
@@ -44,24 +36,34 @@ const ResourceTemplateRow = ({ row, handleClick, handleCopy, handleEdit }) => {
...
+          {canCreate && (
+            <Link
+              data-testid={`createResource-${row.resourceLabel}`}
+              to={{ pathname: "/editor", state: {} }}
+              onClick={(e) => handleClick(row.id, e)}
+              title="Create resource"
+            >
+              <FileEarmarkPlusFill style={{ paddingRight: "10px" }} size={32} />
+            </Link>
+          )}
```
peetucket, line 48: our new icon for creating a resource

src/components/templates/ResourceTemplateRow.jsx
```diff
@@ -6,9 +6,10 @@ import PropTypes from "prop-types"
 import { Link } from "react-router-dom"
 import LongDate from "components/LongDate"
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome"
-import { faCopy, faEdit } from "@fortawesome/free-solid-svg-icons"
+import { faCopy, faPencilAlt } from "@fortawesome/free-solid-svg-icons"
```
peetucket, line 9: this fontawesome icon better matches the new design mockup

package.json
```diff
@@ -85,6 +85,7 @@
     "process": "^0.11.10",
     "rdf-ext": "^1.3.5",
     "react": "^16.14.0",
+    "react-bootstrap-icons": "^1.5.0",
```
peetucket, line 88: new bootstrap-icon module since we are using a bootstrap icon for creating resources

jcoyne: Using two different icon libraries seems like unnecessary bloat.  I think we should pick one icon set and use it.  I'd like to hear what @justinlittman's thought on this are.

jcoyne (changes requested):

__tests__/feature/dashboard.test.js
```diff
@@ -28,7 +28,7 @@ describe("viewing the dashboard", () => {
       await screen.findByText(/Uber template1/)
 
       // Click the resource template
-      fireEvent.click(screen.getByText("Uber template1", { selector: "a" }))
+      fireEvent.click(screen.getByTestId("createResource-Uber template1"))
```
jcoyne, line 31: Dobbs' says 
> it is recommended to use this only after the other queries don't work for your use case. Using data-testid attributes do not resemble how your software is used and should be avoided if possible

https://testing-library.com/docs/queries/bytestid

Have you exhausted all other options both here and in all following tests?

jcoyne (changes requested):

__tests__/feature/dashboard.test.js
peetucket: I couldn't find another way of getting to this link, since it is no longer a text link but an icon with a generic `title` attribute (and there could be multiple of these links on the page).  I thought of giving the link a unique `id`, but noticed the `data-testid` pattern used extensively in other Sinopia tests, and also found this article by Kent C Dodds discussing it: https://kentcdodds.com/blog/making-your-ui-tests-resilient-to-change   But perhaps I missed a selector possibility.

jcoyne: It seems to me that the lack of a unique title for this icon is an accessibility problem.  How is a screen reader user to know which type they are creating when they activate the button? Is there a reason we can't give it a unique title?  Something like  `title="Create resource ${row.resourceLabel}"` ?

peetucket: Makes sense -- my implementation was just following the exact specs of the ticket, but this seems totally reasonable.  And also for the Copy and Edit buttons (which will then require more test selector updates)

jcoyne (approved):

PR #3095, opened 2021-10-04
Switch the property heading color to black for better contrast

### LINKED ISSUE #3076: Accessibility: Color contrast in section header
Make the text in these section headers black.

image

jcoyne:
## Why was this change made?

Fixes #3076


ndushay (approved):
Looks fine, but a Before and After picture would be awesome.

PR #3097, opened 2021-10-04
Update the close button for Bootstrap 5

jcoyne:
## Why was this change made?

This is the prescribed way https://getbootstrap.com/docs/5.1/components/close-button/

It used to look like this:

Screen Shot 2021-10-04 at 5 10 22 PM

Now it looks like:
Screen Shot 2021-10-04 at 5 11 52 PM

### DIFF
--- a/__tests__/components/ModalWrapper.test.js
+++ b/__tests__/components/ModalWrapper.test.js
@@ -15,9 +15,11 @@ const testModal = () => {
         <div className="modal-content">
           <div className="modal-header">
             <h5 className="modal-title">Modal title</h5>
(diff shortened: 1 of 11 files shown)


ndushay (changes requested):
Does this change the appearance of the button at all?  Should @michelleif sign off?  Is this part of an issue?

michelleif:
i think just the X changed. looks fine and again, i'll double-check in DEV, thank you!

jcoyne:
@ndushay I attached some screen shots above.

ndushay (approved):
Thanks for screen shots!

PR #3105, opened 2021-10-05
Replace the check on the left nav headings with a circle

### LINKED ISSUE #3100: Dot on the right indicates whether content has been entered for that field
image

jcoyne:
## Why was this change made?

Fixes #3100

Screen Shot 2021-10-05 at 1 09 17 PM

### DIFF
--- a/cypress/integration/leftNav.spec.js
+++ b/cypress/integration/leftNav.spec.js
@@ -128,7 +128,7 @@ describe("Left-nav test", () => {
     cy.get(".left-nav-header").should("contain", "Uber template4")
     cy.get(".left-nav-header").should("contain", "Uber template4, property1")
   })
-  
+
   it("Pops up tooltips for properties with remarks", () => {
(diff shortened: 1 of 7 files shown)


ndushay (changes requested):
don't we still want some kind of check for the new PresenceIndicator to show it's working right?

jcoyne:
@ndushay I opted not to, because it's trivially simple.  If there's something in the array, the dot shows. 🤷

ndushay (changes requested):
I like having the test - showing that the dot appears appropriately, AND that the dot disappears appropriately.  This is an important behavior in the UI.   Your implementation now is much better and simpler ... but keeping the test seems like a good idea -- it's also documenting the expected behavior of the app.

ndushay (approved):
I know how to do that sort of selector in ruby/capybara, but I haven't worked with cypress yet.  I'm gonna let it go.

PR #3108, opened 2021-10-06
Add the group field to the search query

### LINKED ISSUE #3064: on Resource Templates tab, allow find by Group
Resource Templates tab has several ways to find templates, but not by group

add group to the ways you can find

jcoyne:
## Why was this change made?
Fixes #3064

### DIFF
--- a/__tests__/sinopiaSearch.test.js
+++ b/__tests__/sinopiaSearch.test.js
@@ -107,7 +107,7 @@ describe("getSearchResults", () => {
         bool: {
           must: {
             simple_query_string: {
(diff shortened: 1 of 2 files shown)


michelleif:
the original ticket #3064 was for the search (find) on the Resource Templates tab.

I am not seeing that working.

It looks like this PR added group search to the Sinopia Search tab that searches resources. The Sinopia search tab already has Filter by Group, so we weren't including Group in the search.

we could leave this as is and see if there are unexpected results from including group in search, but it doesn't solve #3064

PR #3124, opened 2021-10-07
enable save button when you create a new resource with default values

### LINKED ISSUE #3089: save is not enabled for a template into which I haven't entered any data, but which already has default data
on DEV, with template literal:required

this is a template with one field, a literal, and it has a default

if I open the template and preview, it shows in the RDF preview that the default value is "there"
Screen Shot 2021-10-01 at 2 16 31 PM

but the Save button is disabled, so I can't save, even if i just want to save as is with the default value

peetucket:
## Why was this change made?

Fixes #3089 - after a resource is created and finished loading, check to see if there are any default values, if so, mark as "changed", which will then trigger the save button to be enabled, just like when a user manually changes a value

Added a new test, and localhost browser


__tests__/testUtilities/stateUtils.js
```diff
@@ -229,6 +229,18 @@ const buildResourceWithLiteral = (state, options) => {
...
+    state.entities.propertyTemplates[
+      "ld4p:RT:bf2:Title:AbbrTitle > http://id.loc.gov/ontologies/bibframe/mainTitle"
+    ].defaults = [
+      {
+        literal: "Default literal1",
+        lang: null,
+      },
+    ]
+  }
+
```
peetucket, line 243: update the tests to allow state to be created with a default value, so we can use in our new test above

justinlittman (approved):

PR #3133, opened 2021-10-08
Update NewsItem.jsx

michelleif:
remove "Latest News" line since there isn't any; remove extra space and period in bulleted item

### DIFF
--- a/src/components/home/NewsItem.jsx
+++ b/src/components/home/NewsItem.jsx
@@ -5,7 +5,8 @@ import Package from "../../../package.json"
 
 const NewsItem = () => (
   <div className="news-item" data-testid="news-item">
-    <h3>Sinopia Version {Package.version} highlights</h3>
+  <h3>Latest News</h3>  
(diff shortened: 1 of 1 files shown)


src/components/home/NewsItem.jsx
```diff
@@ -5,8 +5,7 @@ import Package from "../../../package.json"
 
 const NewsItem = () => (
   <div className="news-item" data-testid="news-item">
-    <h3>Latest news</h3>
```
jermnelson, line 8: Should `<h3>Latest news</h3>` be deleted? It is causing a test failure on this PR.

jermnelson (changes requested):

src/components/home/NewsItem.jsx
michelleif: why is that a problem? is the test looking for that line? if so i'm fine to put it back

michelleif:
we decided to do a more comprehensive release next week, which will need more newsitems, plus this keeps failing for me so instead i will make a ticket asking devs to update newsitem

Issue #2176, opened 2020-05-06
White out screen on overlays/modals


astridu:
When a modal comes up, white out the screen behind it so it pops more.

jcoyne:
This is something that Bootstrap does for us by default, but I think we had some problems with that not working due to how we put it in a wrapping component. @jermnelson Can you confirm?

justinlittman:
This was problematic, but can try again with latest Bootstrap.

michelleif:
is this still an issue?

astridu:
Yes.

PR #3136, opened 2021-10-08
prevent body from scrolling when language modal is open

### LINKED ISSUE #3109: scrolling weirdness when trying to change language
when editing a template or a resource, go to existing field, click language (or "no language specified") next to a literal or label field

when the change language modal comes up, sometimes clicking in it makes the form scroll behind the modal, so you end up looking at a different field behind the modal (language setting does get saved correctly)

it happens both when (a) there's already a language and you want to switch to "no language specified" or (b) "no language specified" is set and you want to select a language instead

attaching a couple screen recordings to show:
https://images.zenhubusercontent.com/149367347/7f15ffe3-ee49-4af0-9baf-d2a810ff03aa/screen_recording_2021_10_06_at_10_15_01_am.mov

https://images.zenhubusercontent.com/149367347/0f36d238-c39a-4393-9b98-b36d8845de03/screen_recording_2021_10_06_at_10_23_50_am.mov

peetucket:
## Why was this change made?

Fixes #3109 and #2176

We are doing some special sauce opening/closing of modals instead of using the standard Bootstrap way shown here: https://getbootstrap.com/docs/5.0/components/modal/

This means we need to deal with dimming the background and preventing scrolling ourselves via CSS/useLayoutEffect.  This does that.

This is how it looks with a language modal open (note: the main part of the page cannot be scrolled either):

Screen Shot 2021-10-11 at 1 01 29 PM

Localhost browser

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -321,3 +321,11 @@ div.rbt div {
   border-color: $nobel;
   background-color: $solitude;
 }
+
+.modal-open {
+  overflow: hidden;
+}
+
+.modal {
+  background: rgba(0, 0, 0, 0.5);
+}
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -1,6 +1,6 @@
 // Copyright 2019 Stanford University see LICENSE for license
 
-import React, { useEffect } from "react"
+import React, { useEffect, useLayoutEffect } from "react"
 import { Helmet, HelmetProvider } from "react-helmet-async"
 import Config from "Config"
 import "react-bootstrap-typeahead/css/Typeahead.css"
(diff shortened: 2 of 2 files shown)


src/styles/main.scss
```diff
@@ -321,3 +321,11 @@ div.rbt div {
...
   background-color: $solitude;
 }
+
+.modal-open {
+  overflow: hidden;
+}
+
+.modal {
+  background: rgba(0, 0, 0, .5);
+}
```
peetucket, line 331: this makes the background a dim grey when a model is open

src/styles/main.scss
```diff
@@ -321,3 +321,11 @@ div.rbt div {
   border-color: $nobel;
   background-color: $solitude;
 }
+
+.modal-open {
+  overflow: hidden;
+}
```
peetucket, line 327: this prevents the content in the main tag from scrolling when a modal is opened (and we add this class to the body tag as shown above)

justinlittman (approved):

PR #3150, opened 2021-10-11
templateValidationHelpers: update error message for nested resource p…

### LINKED ISSUE #2860: rewrite old error message when template is missing specification of nested resource template
if a template has a field of type `nested resource` but no nested resource is specified, then when you try to open the template for cataloging, you get this error message:
`Referenced resource templates (valueTemplateRefs) are required for http://www.property.com.`

The error message is confusing because we no longer have valueTemplateRefs

Change message to:
`The field [Label] with property [Property] has type nested resource, but does not specify a nested template in Resource attributes'

ndushay:
…roperty with no specified nested template

## Why was this change made?

Fixes #2860 

### After

image

### Before:

image

locally ^^

### DIFF
--- a/src/actionCreators/templateValidationHelpers.js
+++ b/src/actionCreators/templateValidationHelpers.js
@@ -94,7 +94,7 @@ const validatePropertyTemplate = (template) => {
     _.isEmpty(template.valueSubjectTemplateKeys)
   ) {
     errors.push(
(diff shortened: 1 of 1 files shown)


ndushay:
@michelleif I changed it slightly from your suggestion to this (removed the word nested so it didn't say "nested template in Nested resource attributes"):

image

PR #3151, opened 2021-10-12
Expand property when necessary.

### LINKED ISSUE #3120: un-loaded field becomes uneditable when loading RDF into a template
when I "Load RDF" into a template, but don't load a value for every field in that template, I still expect to see all the fields in the template, with empty fields where I didn't load a value. Those empty fields should be editable so I can add a value. 

on DEV now, post input interaction improvements, if there's a field in the template that doesn't have data loaded into it, upon Load RDF, the field becomes uneditable.

example from Stage of current, expected behavior:
template: test:two_literals
the template has two literal fields

I want to load data for just one field, then fill in the other field myself, so I Load this RDF:
```
<> <http://sinopia.io/vocabulary/hasResourceTemplate> "test:two_literals" .
<> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://id.loc.gov/ontologies/bibframe/Work> .
<> <http://property.com/field1> "xyz"@eng .
```

I get this:

I have to click +Add on `literal 2`, but it does let me add data to `field 2`, as expected

compare on DEV:
here's the same template on DEV, showing the template ready for editing, to show it has 2 literal fields

now try loading the same RDF we did on stage:
```
<> <http://sinopia.io/vocabulary/hasResourceTemplate> "test:two_literals" .
<> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://id.loc.gov/ontologies/bibframe/Work> .
<> <http://property.com/field1> "xyz"@eng .
```

I get this, where `literal 2` is an uneditable field; the fieldname shows, but I can't enter anything in the field because the entry box doesn't show

i tried saving the resource as is, copying the URI, and going to Search to try to open it for editing that way to see if `literal 2` is now editable, but it's not.

to see the thing I saved, and see the uneditable field, search for: https://api.development.sinopia.io/resource/ab851b64-a19b-4be6-9898-6a1f4ab792cf

justinlittman:
closes #3120

## Why was this change made?
When loading resources, some properties were not in editable state.

Unit, local

### DIFF
--- a/src/components/editor/property/PanelProperty.jsx
+++ b/src/components/editor/property/PanelProperty.jsx
@@ -94,10 +94,7 @@ const PanelProperty = (props) => {
                 onClick={() => props.contractProperty(props.property.key)}
                 data-id={props.id}
               >
-                <FontAwesomeIcon
-                  className="fa-inverse trash-icon"
-                  icon={trashIcon}
-                />
+                <FontAwesomeIcon className="trash-icon" icon={trashIcon} />
               </button>
             )}
           </h5>
--- a/__tests__/__action_fixtures__/addSiblingValueSubject-ADD_VALUE.json
+++ b/__tests__/__action_fixtures__/addSiblingValueSubject-ADD_VALUE.json
@@ -78,8 +78,8 @@
             "properties": [
               {
                 "key": "abc1",
-                "values": null,
-                "show": false
+                "values": [],
+                "show": true
               }
             ]
           }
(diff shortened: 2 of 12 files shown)


src/components/editor/property/PanelProperty.jsx
```diff
@@ -94,10 +94,7 @@ const PanelProperty = (props) => {
                 onClick={() => props.contractProperty(props.property.key)}
                 data-id={props.id}
               >
-                <FontAwesomeIcon
-                  className="fa-inverse trash-icon"
-                  icon={trashIcon}
-                />
+                <FontAwesomeIcon className="trash-icon" icon={trashIcon} />
```
justinlittman, line 97: Fixed a white icon on a white background.

PR #3159, opened 2021-10-12
disable scrolling and clicks behind modals

### LINKED ISSUE #3109: scrolling weirdness when trying to change language

peetucket:
## Why was this change made?

More attempts to fix #3109 as a follow on to work done in #3136 

1. Prevent scrolling when a field is clicked behind a modal
2. Prevent fields from being selected behind a modal

~~Still unclear: clicking a header in a resource template behind the modal still changes the left nav (even though that doesn't really impact you or cause problems, it can be confusing).  The template property headers are rendered here:~~

~~https://github.com/LD4P/sinopia_editor/blob/main/src/components/editor/property/PanelProperty.jsx#L63-L68~~
~~https://github.com/LD4P/sinopia_editor/blob/main/src/components/editor/property/PropertyLabel.jsx~~

~~but I  don't see where we are responding to clicks to change the left nav in response (to disable there too).~~

UPDATE: Now unable to reproduce this behavior :/

Localhost browser

### DIFF
--- a/src/components/editor/leftNav/ActivePanelPropertyNav.jsx
+++ b/src/components/editor/leftNav/ActivePanelPropertyNav.jsx
@@ -1,17 +1,15 @@
 import React from "react"
 import PropTypes from "prop-types"
-import { setCurrentComponent } from "actions/index"
-import { useDispatch, useSelector } from "react-redux"
+import { useSelector } from "react-redux"
 import { displayResourceValidations } from "selectors/errors"
 import { selectNormProperty, selectNormValues } from "selectors/resources"
 import { selectPropertyTemplate } from "selectors/templates"
 import SubjectSubNav from "./SubjectSubNav"
 import PresenceIndicator from "./PresenceIndicator"
-import { selectModalType } from "selectors/modals"
+import useLeftNav from "hooks/useLeftNav"
 import _ from "lodash"
 
 const ActivePanelPropertyNav = (props) => {
-  const dispatch = useDispatch()
   const property = useSelector((state) =>
     selectNormProperty(state, props.propertyKey)
   )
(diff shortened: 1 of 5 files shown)


src/components/editor/leftNav/ActivePanelPropertyNav.jsx
```diff
@@ -44,6 +45,20 @@ const ActivePanelPropertyNav = (props) => {
...
+
+    dispatch(
+      setCurrentComponent(
+        property.rootSubjectKey,
+        property.rootPropertyKey,
+        property.key
+      )
+    )
+  }
+
```
peetucket, line 61: This function is duplicated in other spots, would be better to refactor to be DRY

src/components/editor/leftNav/ActivePanelPropertyNav.jsx
```diff
@@ -21,6 +20,8 @@ const ActivePanelPropertyNav = (props) => {
     selectPropertyTemplate(state, property?.propertyTemplateKey)
   )
 
+  const handleClick = useLeftNav(property)
+
```
peetucket, line 24: our new hook defined below that will prevent clicks from being activated when a modal is open

PR #3107, opened 2021-10-05
update top left nav for visual design

ndushay:
## Why was this change made?

Fixes #3078

### After

image

image

### Before

image

image

PR #3163, opened 2021-10-12
Add Sinopia version to footer and remove now unused related code

### LINKED ISSUE #3149: place version number in footer
the version number disappeared from top nav

this makes it impossible to know what version one is on

@astridu can you help by pointing to ticket that shows where version number should be?

ndushay:
## Why was this change made?

Fixes #3149

Also cleans up unused code and code that is now un-necessary  (`setAppVersion` and `selectAppVersion`)

### After

image

### Before

image

unit tests, locally

### DIFF
--- a/src/store.js
+++ b/src/store.js
@@ -8,10 +8,6 @@ export const initialState = {
   authenticate: {
     user: undefined,
   },
-  app: {
-    version: undefined,
-    lastChecked: Date.now(),
-  },
   editor: {
     // The state of the editor
     copyToNewMessage: {


src/actions/index.js
```diff
@@ -1,10 +1,7 @@
...
-  payload: version,
-})
-
-export const setCurrentComponent = (rootSubjectKey, rootPropertyKey, key) => ({
+const setCurrentComponent = (rootSubjectKey, rootPropertyKey, key) => ({
   type: "SET_CURRENT_COMPONENT",
   payload: { rootSubjectKey, rootPropertyKey, key },
 })
+
+export default setCurrentComponent
```
justinlittman, line 7: Following our existing pattern, please add a `export const noop = () => {}` so that `setCurrentComponent` does not need to be default export.

src/components/Footer.jsx
```diff
@@ -35,7 +36,8 @@ const Footer = () => (
           >
             Creative Commons CC0 1.0 Universal Public Domain Dedication
           </a>
-          . Published <a href="/vocabulary">Vocabulary</a>.
+          . Published <a href="/vocabulary">Vocabulary</a>. &nbsp; Sinopia v
+          {Package.version}
```
ndushay, line 40: since the app version code is only used in a single place, I conferred with @justinlittman and we decided to use the same approach as in `NewsItem.jsx` and have the Footer component get the version number directly from package.json.

If there is some performance problem with this approach, then we can keep this code and `NewsItem.jsx` and the footer should retrieve the app version from state.

src/styles/main.scss
```diff
@@ -179,11 +179,6 @@
   margin-bottom: 0px;
 }
 
-a.editor-subtitle {
```
justinlittman, line 182: Curious why this was removed.

__tests__/reducers/index.test.js
```diff
@@ -1,25 +1,6 @@
...
-
-    const oldState = {
-      version: "3.0.0",
-    }
-
-    const newState = reducer(oldState, action)
-
-    expect(newState.version).toStrictEqual("2000.0.1")
-  })
-})
```
ndushay, line 22: no longer in the code, so removing the test.

justinlittman (changes requested):

src/actionCreators/resources.js
```diff
@@ -37,7 +37,7 @@ import {
 } from "actionCreators/user"
 import { addResourceHistory } from "actionCreators/history"
 import _ from "lodash"
-import { setCurrentComponent } from "actions/index"
+import setCurrentComponent from "actions/index"
```
ndushay, line 40: `setCurrentComponent` became a default export so this simple change is in a number of files.

src/components/Header.jsx
```diff
@@ -9,7 +9,6 @@ import { selectUser } from "selectors/authenticate"
 import { signOut } from "actionCreators/authenticate"
 import { bindActionCreators } from "redux"
 import { selectCurrentResourceKey } from "selectors/resources"
-import { selectAppVersion } from "selectors/index"
 import usePermissions from "hooks/usePermissions"
```
ndushay, line 12: these statements in Header should have been removed in PR #3107, as they stopped being used because of that PR.

src/styles/main.scss
```diff
@@ -179,11 +179,6 @@
...
-  font-family: "Copperplate Gothic Light";
-  color: #2f2424 !important;
-}
-
```
ndushay, line 186: this class is no longer used.  It should have been removed with PR #3107

src/store.js
```diff
@@ -8,10 +8,6 @@ export const initialState = {
   authenticate: {
     user: undefined,
   },
-  app: {
-    version: undefined,
-    lastChecked: Date.now(),
-  },
```
ndushay, line 14: we no longer store the app version, so we don't need this.

src/styles/main.scss
ndushay: it's also a commit message.  :-P

justinlittman (approved):

PR #3167, opened 2021-10-13
Preview buttons

### LINKED ISSUE #3155: uneven button spacing on preview
when previewing (View Resource) from Search results, the buttons on the preview modal have uneven spacing

* Adjust so that there is "little space" between all buttons.
* Change all buttons to text only, i.e., remove icons.

ndushay:
## Why was this change made?

Closes #3155
- removes icons from `Save` and `Copy` buttons
- adjusts the button margins, in some cases by changing css for an existing class, in some cases adding a new css class, and in some cases removing a css class from the MarcButton.

If this needs to be tested, @justinlittman I think this needs to be deployed to dev as I can't get the `Marc` or `Export to Symphony` buttons locally for a resource, I don't believe. 

### After

Running locally on my machine:

image

Simulating the changes by directly editing the css in the web inspector for a resource on dev (which is still showing the icons on the `Save` and `Copy` buttons:

image

### Before

image

locally and with manual changes to css in browser inspector

### DIFF
--- a/src/components/editor/actions/MarcButton.jsx
+++ b/src/components/editor/actions/MarcButton.jsx
@@ -85,7 +85,7 @@ const MarcButton = () => {
   const dropDownItemBtnClasses = ["btn", "btn-secondary", "dropdown-item"]
 
   return (
-    <div className="view-resource-buttons btn-group dropstart">
+    <div className="btn-group dropstart">
       <button
         type="button"
         id="marcBtn"


justinlittman:
If running locally, with fixture, you can search on `http://localhost:3000/resource/9c5bd9f5-1804-45bd-99ed-b6e3774c896e`, which is a bf:Instance.

image

Issue #2897, opened 2021-08-05
when Previewing from Search tab, offer same Preview options as in Editor


michelleif:
this includes offering the requested "user-friendly" preview (#2178)

justinlittman:
This is more critical now that only users that can edit a resource can open preview the RDF.

Note that both the resource preview and the RDF preview are modals, so this needs to be taken into consideration in the design.

astridu:
Could we do something like this where we have 2 tabs in the modal. One shows the resource and the other the RDF preview?
![image](https://user-images.githubusercontent.com/42216982/134558411-2cfe360b-9de6-4670-94a9-fb2cb85ca4c9.png)
![image](https://user-images.githubusercontent.com/42216982/134558502-54210997-019d-44a9-b63e-0acf364bdd3a.png)

michelleif:
why not just one preview and then under Format you choose table, n-triples, json, "summary" (or whatever we call the human-readable one (and add #2178)?

astridu:
Like this?
![image](https://user-images.githubusercontent.com/42216982/134563793-cd4f44c4-8138-4556-93b5-dde834d09a24.png)

justinlittman:
So those are tabs on the preview modal?

astridu:
I was trying to follow the bootstrap style. Is that not correct? Any tab style will do.

michelleif:
no, i'm thinking no tabs, it's just a window, "Preview", and once the window is open, you choose which style preview you want. in my mind, previewing the RDF data is a preview just like previewing the data in the form. why do we need to distinguish between the "View Resource" and "Preview"

astridu:
Like this? And then the pull down determines what you see in the modal?
![image](https://user-images.githubusercontent.com/42216982/134582995-773d7069-5b7f-4450-a86b-5dabebb77bca.png)

michelleif:
Yes!!!!!

justinlittman:
So what is the not-RDF view called?

michelleif:
what is that view for? the view in #2178 is requested also which is called Summary. do we need Summary and current-not-RDF? (I know, Summary doesn't exist yet)

justinlittman:
You go with the summary you have, not the summary you wish you had ...

FYI -- As I replace the input components, I'm also changing the read-only display to just be text (not using disabled inputs like it currently is).

michelleif:
The current "View Resource" style preview avail now only from Search would be called "Form view".

[...]

PR #3172, opened 2021-10-13
Make previews consistent.

justinlittman:
## Why was this change made?
So that all previews showed RDF and form views.

Unit


justinlittman:
image

peetucket (approved):
There is a kind of a lot going on here, so it's hard to pick apart the styling changes from the renaming of actions (to make them clearer?) and refactoring of tests (to avoid unnecessary clicking and just rendering directly instead?).  I don't see anything out of place, though it would have been nice to see it attached to a ticket explaining what the goal is.

justinlittman:
Sorry @peetucket. This fixes https://github.com/LD4P/sinopia_editor/issues/2897, but as you note involves a bunch of refactoring of both code and styling, with a bunch of test changes.

PR #3173, opened 2021-10-13
home page sinopia logo matches editor-logo

### LINKED ISSUE #3156: display environment name in homepage logo
when on SInopia homepage, the logo doesn't tell you what environment you're in (yes, the URL does)

It would be helpful if the homepage made it obvious, like we do once you're in the app:
 

@astridu please advise on logo design, use same as within app?

ndushay:
## Why was this change made?

Closes #3156

### After

The displays below show no environment because it isn't set on my machine, but it will work upon deployment.

Home Page:

image

is now suppose to match (font size, font color) Editor Page (for comparison):

image

### Before

image

locally


src/components/home/Header.jsx
```diff
@@ -3,7 +3,6 @@
 import React from "react"
 import PropTypes from "prop-types"
 import { Link } from "react-router-dom"
-import SinopiaLogo from "styles/sinopia-logo.png"
```
ndushay, line 6: should this file be removed from the app?

ndushay:
@justinlittman code coverage is being silly.

src/components/home/Header.jsx
justinlittman: Please.

PR #3191, opened 2021-10-15
Add preview header with links/permissions info

### LINKED ISSUE #3177: view-only Permissions in Preview
View-only permissions in preview. If the 'editable by' list is very long, add an ellipsis which user can click to see full list.
image

peetucket:
## Why was this change made?

Fixes #3177 - add a header to the preview that uses the URL, label, and permissions info, which is collapsible if there are many editable groups.

Click the three dots to expand, which then shows the entire list and hides the three dots.  See example below:

Screen Shot 2021-10-15 at 9 57 05 AM

Screen Shot 2021-10-15 at 9 57 12 AM


src/components/editor/ResourceComponent.jsx
```diff
@@ -51,7 +51,7 @@ const ResourceComponent = () => {
           <CopyToNewMessage />
           <div className="row">
             <div className="col-md-11">
-              <ResourceURIMessage />
+              <ResourceURIMessage resourceKey={resourceKey} />
             </div>
```
peetucket, line 55: this component was edited to take this parameter so it can be re-used by the new preview header component I created

src/components/editor/preview/ResourcePreviewHeader.jsx
```diff
@@ -0,0 +1,52 @@
...
+      </div>
+    </React.Fragment>
+  )
+}
+
+ResourcePreviewHeader.propTypes = {
+  resource: PropTypes.object.isRequired,
+}
+
+export default ResourcePreviewHeader
```
peetucket, line 52: the new header component with collapsible edit groups

PR #3192, opened 2021-10-15
Refactors search filter components.

### LINKED ISSUE #2611: Make the filter box a standard UX pattern
Here's my proposal. 
- Dropdown uses secondary button style.
- Go uses primary button style
- No boxes selected, user activates Select/Deselect all: All boxes selected
- User then deselects some boxes: Select/Deselect all box becomes unchecked
- User then checks 'Select/Deselect All' again: All boxes checked
- User will see what happens as soon as they activate the checkbox:
   - If no boxes selected, then all boxes are selected. 
   - If some boxes selected, then all boxes are selected. 
   - If all boxes are selected, then all boxes become deselected. 
- For the case where user manually checks/unchecks all the boxes, 'Select/Deselct all' control does not have to respond to the manual selection.
image

_This description consolidates discussion from below._

justinlittman:
closes #2611

## Why was this change made?
Better UX.

Unit


src/components/search/SearchFilter.jsx
```diff
@@ -0,0 +1,189 @@
...
+      setSelectedFilters(facetResults.map((result) => result.key))
+    }
+  }
+
+  // const handleKeyDown = (event) => {
+  //   if (event.key === "Enter") {
+  //     dispatch(updateLiteralValue(value.key, currentContent, value.lang))
+  //     event.preventDefault()
+  //   }
+  // }
```
ndushay, line 86: keeping this comment because?

__tests__/components/search/GroupFilter.test.js
```diff
@@ -97,7 +97,7 @@ describe("<GroupFilter />", () => {
     })
   })
 
-  it("allows selecting only", async () => {
+  it("allows selecting / deselecting all", async () => {
```
ndushay, line 100: wondering if you want to test this functionality one for the SearchFilter component, rather than twice in these tests.  It would remove some duplication, no?

src/components/search/SearchFilter.jsx
justinlittman: Removed.

Issue #3045, opened 2021-09-25
Avoid useSelector unnecessary rerenders


justinlittman:
We currently trigger a huge number of unnecessary rerenders due to our use mis-use of useSelector.

See https://react-redux.js.org/api/hooks#useselector.
Also, https://blog.jjohns.com/react/redux/performance/2020/09/14/avoiding-useselector-unnecessary-rerenders.html
This will definitely require the use of re-select.

justinlittman:
For example, in experimentation, removing the `|| []` in https://github.com/LD4P/sinopia_editor/blob/main/src/selectors/errors.js#L24 reduced rerenders because it wasn't returning a new array on every call.

I would start with ResourceComponent and determine what is triggering rerender.

justinlittman:
I'm primarily concerned with rerenders that occur when the resource is initially loading in the editor.

PR #3206, opened 2021-10-18
Refactors resource reducers to minimize state churn.

justinlittman:
refs #3045

## Why was this change made?
Speedier app.

Unit, local


mjgiarlo (approved):
LGTM

PR #3208, opened 2021-10-19
Changes to error selectors to reduce rerenders.

justinlittman:
refs #3045

## Why was this change made?
Speedier app

### DIFF
--- a/src/selectors/errors.js
+++ b/src/selectors/errors.js
@@ -1,4 +1,4 @@
-import { selectProperty, selectSubject } from "./resources"
+import { selectProperty, selectSubject, selectNormSubject } from "./resources"
 import _ from "lodash"
 
 /**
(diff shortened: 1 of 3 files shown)


src/selectors/errors.js
```diff
@@ -8,20 +8,17 @@ import _ from "lodash"
...
-  state.editor.resourceValidation[resourceKey] || false
+  !!state.editor.resourceValidation[resourceKey]
 
 export const hasValidationErrors = (state, resourceKey) => {
-  if (!resourceKey) return false
-
-  const subject = selectSubject(state, resourceKey)
-  return !_.isEmpty(subject.descWithErrorPropertyKeys)
+  const subject = selectNormSubject(state, resourceKey)
+  return !_.isEmpty(subject?.descWithErrorPropertyKeys)
```
justinlittman, line 15: `selectNormSubject` is faster because does not enhance the subject.

src/selectors/errors.js
```diff
@@ -1,4 +1,4 @@
-import { selectProperty, selectSubject } from "./resources"
+import { selectProperty, selectSubject, selectNormSubject } from "./resources"
```
ndushay, line 1: I'm only looking at the git diff - is `selectSubject` still used?

PR #3209, opened 2021-10-19
Get uri in PropertyComponent instead of entire property to reduce rerenders

justinlittman:
refs #3045

## Why was this change made?
Speedier app

### DIFF
--- a/src/components/editor/property/PropertyComponent.jsx
+++ b/src/components/editor/property/PropertyComponent.jsx
@@ -8,20 +8,18 @@ import NestedResource from "./NestedResource"
 import ReadOnlyInputLiteralOrURI from "../inputs/ReadOnlyInputLiteralOrURI"
 import Alert from "../../Alert"
 import { displayResourceValidations } from "selectors/errors"
-import { selectNormSubject } from "selectors/resources"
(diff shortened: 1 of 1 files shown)


src/components/editor/property/PropertyComponent.jsx
```diff
@@ -8,20 +8,18 @@ import NestedResource from "./NestedResource"
...
 import { displayResourceValidations } from "selectors/errors"
-import { selectNormSubject } from "selectors/resources"
+import { selectUri } from "selectors/resources"
 
 // Decides how to render this property.
 const PropertyComponent = ({ property, propertyTemplate, readOnly }) => {
-  const resource = useSelector((state) =>
-    selectNormSubject(state, property.rootSubjectKey)
-  )
+  const uri = useSelector((state) => selectUri(state, property.rootSubjectKey))
```
justinlittman, line 15: Only need the URI, so only get the URI.

ndushay: only looking at the diff here - is there any other use of uri in the file (i.e. should it be named propertyUri) ?

ndushay (approved):
approving with a question for your consideration.

PR #3283, opened 2021-10-30
Change buttons to spinner while loading. Componentize buttons. Refact…

justinlittman:
…or useResource.

## Why was this change made?
So that users know when Sinopia is working.

### DIFF
--- a/__tests__/feature/dashboard.test.js
+++ b/__tests__/feature/dashboard.test.js
@@ -28,7 +28,7 @@ describe("viewing the dashboard", () => {
       await screen.findByText(/Uber template1/)
 
       // Click the resource template
-      fireEvent.click(screen.getByTitle("Create resource for Uber template1"))
+      fireEvent.click(screen.getByTestId("Create resource for Uber template1"))
       await waitFor(() =>
         expect(
           screen.getAllByText("Uber template1", { selector: "h3" })
--- a/src/components/dashboard/Dashboard.jsx
+++ b/src/components/dashboard/Dashboard.jsx
@@ -14,7 +14,6 @@ import ResourceTemplateSearchResult from "../templates/ResourceTemplateSearchRes
 import SearchList from "./SearchList"
 import Alerts from "../Alerts"
 import _ from "lodash"
-import useResource from "hooks/useResource"
 import PreviewModal from "../editor/preview/PreviewModal"
 
 export const dashboardErrorKey = "dashboard"
@@ -29,7 +28,6 @@ const Dashboard = (props) => {
   const historicalResources = useSelector((state) =>
     selectHistoricalResources(state)
   )
-  const { handleNew, handleCopy, handleEdit } = useResource(dashboardErrorKey)
 
   const showWelcome =
     _.isEmpty(historicalTemplates) &&
@@ -40,7 +38,7 @@ const Dashboard = (props) => {
     <section id="dashboard">
       <Header triggerEditorMenu={props.triggerHandleOffsetMenu} />
       <Alerts errorKey={dashboardErrorKey} />
-      <PreviewModal handleEdit={handleEdit} handleCopy={handleCopy} />
+      <PreviewModal errorKey={dashboardErrorKey} />
       {showWelcome && (
         <div>
           <h2>Welcome to Sinopia.</h2>
@@ -55,9 +53,7 @@ const Dashboard = (props) => {
           <h2>Recent templates</h2>
           <ResourceTemplateSearchResult
             results={historicalTemplates}
-            handleClick={handleNew}
-            handleEdit={handleEdit}
-            handleCopy={handleCopy}
+            errorKey={dashboardErrorKey}
           />
         </div>
       )}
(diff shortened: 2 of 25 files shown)


src/components/dashboard/Dashboard.jsx
```diff
@@ -55,9 +53,7 @@ const Dashboard = (props) => {
           <h2>Recent templates</h2>
           <ResourceTemplateSearchResult
             results={historicalTemplates}
-            handleClick={handleNew}
-            handleEdit={handleEdit}
-            handleCopy={handleCopy}
+            errorKey={dashboardErrorKey}
```
justinlittman, line 56: Notice that we're passing around `errorKey` rather than the various useResource functions.

src/components/editor/leftNav/Versions.jsx
```diff
@@ -27,18 +26,21 @@ const Versions = ({ resource }) => {
   const [compareFrom, setCompareFrom] = useState()
   const [compareTo, setCompareTo] = useState("current")
   const [canDiff, setCanDiff] = useState(true)
+  const [isLoadingView, setLoadingView] = useState(false)
```
justinlittman, line 29: Versions is a little different in that it isn't using `useResource`.

src/components/search/SearchResultRows.jsx
```diff
@@ -3,91 +3,32 @@
...
-              data-testid={`Copy ${row.label}`}
-              aria-label={`Copy ${row.label}`}
-            >
-              <FontAwesomeIcon icon={faCopy} className="icon-lg" />
-            </button>
-          )}
-        </div>
-      </td>
-    </tr>
+    <SearchResultRow
```
justinlittman, line 18: Refactoring into own component.

src/hooks/useResource.js
```diff
@@ -18,7 +18,10 @@ import {
 } from "actions/resources"
 import { useHistory } from "react-router-dom"
 
-const useResource = (errorKey, errorRef) => {
+const useResource = (
+  errorKey,
+  { resourceTemplateId = null, resourceURI = null }
```
justinlittman, line 23: useResource is bound to a particular resourceTemplateId / resourceURI now, rather than providing handle functions in which these are passed in as arguments.

src/hooks/useResource.js
```diff
@@ -27,52 +30,59 @@ const useResource = (errorKey, errorRef) => {
...
 
   const [navigateEditor, setNavigateEditor] = useState(false)
+  const [status, setStatus] = useState("ready")
 
-  useLayoutEffect(() => {
+  useEffect(() => {
     // Forces a wait until the root resource has been set in state
     if (navigateEditor && resourceKey && _.isEmpty(errors)) {
       history.push("/editor")
-    } else if (!_.isEmpty(errors) && errorRef) {
```
justinlittman, line 35: I got rid of the scrolling to errors. If proves to be a problem, I'd probably add the scrolling directly to Alert, rather than here.

__tests__/feature/dashboard.test.js
```diff
@@ -28,7 +28,7 @@ describe("viewing the dashboard", () => {
       await screen.findByText(/Uber template1/)
 
       // Click the resource template
-      fireEvent.click(screen.getByTitle("Create resource for Uber template1"))
+      fireEvent.click(screen.getByTestId("Create resource for Uber template1"))
```
jcoyne, line 31: Why are you switching to `*ByTestId`.  I believe the guide says that we should avoid using this when possible.

jcoyne: > In the spirit of the guiding principles, it is recommended to use this only after the other queries don't work for your use case. Using data-testid attributes do not resemble how your software is used and should be avoided if possible. That said, they are way better than querying based on DOM structure or styling css class names. Learn more about data-testids from the blog post "Making your UI tests resilient to change"

https://testing-library.com/docs/queries/bytestid/

justinlittman: It does. And even better, we should using `*ByRole`. However, these other queries are _very slow_, leading to _very slow_ and _very flaky_ tests. As a matter of convention, we have added `test-id` attributes to mirror other attributes, usually `aria-label`. Due to this convention, querying by test id is roughly equivalent to these better tests but much faster.

In this particular case, the title was changed to be consistent with other actions which had generic titles like "Create new" and specific arial-labels and test-ids like "Create resource for Uber template1".

ndushay (approved):
I'm not sure I fully grok this -- we seem to be passing errorKey around rather than event handlers -- but I'm not seeing any red flags.

I like the components being refactored out to their own files.

src/hooks/useResource.js
ndushay: wondering if this comment should be captured somewhere more useful, like in an issue "if xxx is a problem, then"

justinlittman:
Yes, error key is being passed around. I think there is an opportunity to refactor in the future, perhaps using a React boundary. (Oh, I like that idea ...)

PR #3296, opened 2021-11-03
[HOLD for PO vetting] Literal validation (integer, dateTime and dateTimeStamp and regex)

### LINKED ISSUE #2890: validate data in literal fields against datatype or regex specified in template
if a literal field has a datatype (see #2889) or a regex (see #2907) specified, validate that entered data conforms to the datatype or regex

if it doesn't, show error message "Not a valid [datatype]" or "Entered data must match [regex]" (or whatever wording makes most sense)

ndushay:
## Why was this change made?

closes #2890 

unit tests, running locally, deployed to dev.

README with new properties.


README.md
```diff
@@ -291,10 +291,11 @@ The following are only in the resource subject (that is, the base subject).
   remarkUrl: <remark url, e.g., "http://access.rdatoolkit.org/2.13.html">
   remarkUrlLabel: <a descriptive label about the remark url, e.g., "Title Proper">
   type: <resource | uri | literal>,
+  validationRegex: <regex as a string>,
+  validationDataType: <uri for integer, dateTime, etc.>,
   component: <InputLookup | InputLookupQA | InputList | InputLiteral | InputURI>,
-  valueSubjectTemplateKeys: [<subject template keys>]
+  valueSubjectTemplateKeys: [<subject template keys>],
```
ndushay, line 297: I admit that part of me was itching to alphabetize this list;  if order does matter, I hope I put the new ones in the right place.

__tests__/reducers/resourceValidationHelpers.test.js
```diff
@@ -0,0 +1,480 @@
...
+  // loadResourceFinished,
+  // setResourceGroup,
+  // setValueOrder,
+  // clearResourceFromEditor,
+  // saveResourceFinishedEditor,
+  // updateValue,
+  // setCurrentDiffResources,
+  // setVersions,
+  // clearVersions,
+} from "reducers/resources"
```
ndushay, line 34: I wasn't sure what I needed to write tests;   at this point, I wonder if I need to have tests for updateValue or any other reducers.

__tests__/testUtilities/stateUtils.js
```diff
@@ -247,6 +247,24 @@ const buildResourceWithLiteral = (state, options) => {
     ]
   }
 
+  if (options.hasIntegerValidation) {
+    state.entities.propertyTemplates[
+      "ld4p:RT:bf2:Title:AbbrTitle > http://id.loc.gov/ontologies/bibframe/mainTitle"
```
ndushay, line 252: inappropriate propertyTemplate for these sorts of literals, but gud enuf for testing ?

src/TemplatesBuilder.js
```diff
@@ -284,6 +288,30 @@ export default class TemplatesBuilder {
...
+      propertyTerm,
+      "http://sinopia.io/vocabulary/hasLiteralAttributes"
+    )
+    if (!attributeTerm) return null
+    return this.valueFor(
+      attributeTerm,
+      "http://sinopia.io/vocabulary/hasValidationDataType"
+    )
+  }
+
```
ndushay, line 314: I'm not sure if this is needed??

src/reducers/resourceHelpers.js
```diff
@@ -177,6 +177,9 @@ const removeFromKeyArray = (obj, key, value) => {
   obj[key] = [...obj[key]].filter((checkValue) => checkValue !== value)
 }
 
+// refactor this into more than one function;  move into separate file
+//   look at how we validate templates ... maybe use that as a model??
+//   (see actionCreators/templateValidationHelpers)
```
ndushay, line 182: I wanted to get tests passing and code written before refactoring.  I'm at that point now.

__tests__/reducers/resourceValidationHelpers.test.js
justinlittman: I would add a single test that shows that validation is being performed for updateValue (if there isn't one already). But no need to repeat all of the test.

src/TemplatesBuilder.js
justinlittman: Yup.

src/reducers/resourceHelpers.js
```diff
@@ -202,6 +205,53 @@ export const updateValueErrors = (state, valueKey) => {
...
   }
 
+  if (
+    propertyTemplate.type === "literal" &&
+    propertyTemplate.validationDataType ===
+      "http://www.w3.org/2001/XMLSchema/integer" &&
+    Number.isNaN(parseInt(value.literal, 10))
+  )
+    errors.push(
+      `Literal validationDataType is 'http://www.w3.org/2001/XMLSchema/integer' but '${value.literal}' is not an integer`
```
justinlittman, line 215: Would prefer `Expected datatype is ...` rather than `Literal validationDataType ...` here and below.

src/reducers/resourceHelpers.js
```diff
@@ -202,6 +205,53 @@ export const updateValueErrors = (state, valueKey) => {
     if (value.uri && !isValidURI(value.uri)) errors.push("Invalid URI")
   }
 
+  if (
+    propertyTemplate.type === "literal" &&
+    propertyTemplate.validationDataType ===
+      "http://www.w3.org/2001/XMLSchema/integer" &&
```
justinlittman, line 211: Is this going to give a validation error for null/no value? (We don't want it to.)

src/TemplatesBuilder.js
```diff
@@ -127,6 +127,10 @@ export default class TemplatesBuilder {
     const propertyTemplate = this.newBasePropertyTemplate(propertyTerm)
     propertyTemplate.type = "literal"
     propertyTemplate.defaults = this.defaultsForLiteral(propertyTerm)
+    propertyTemplate.validationRegex =
```
justinlittman, line 130: This needs a test. (I know you're not done, but thought I would point out just in case.)

ndushay: will do.

ndushay (commented):
I'm not sure my tests are checking for the correct state conditions when there is a validation error.

src/reducers/resourceHelpers.js
ndushay: now done.

src/reducers/resourceHelpers.js
ndushay: done.

ndushay:
build is passing now, regardless of what it says here ...

ndushay (commented):
@justinlittman I think I've addressed everything except for adding the regex stuff (which is next);   Please review my test conditions as I'm not 100% sure I'm looking for the right results in state.

This should be ready to go out for vetting by Michelle.

Note:  any backslashes in regex entered by user need to be (at least) double escaped.  In the javascript CODE, they must be `"\\d*"` to get `/\d*/`.   There may be more escaping necessary due to browser stuff.   I also highly recommend using beginning and end of string indicators `"^yer-stuff-here$"`.    

Where do I document so users will see this stuff at appropriate time?

README.md
```diff
@@ -318,6 +319,8 @@ The following are only in the resource subject (that is, the base subject).
   literal: <literal>,
   lang: <language for literal or URI label>,
   -> langLabel: <label for language>,
+  validationRegex: <regex for literal validation, as a string with backslashes double-escaped (e.g. "^\\d+$" becomes /^\d+$/)>,
+  validationDataType: <uri for integer, dateTime, etc. for literal validation>,
```
ndushay, line 323: I trust this and the new lines 294-297 are sufficient documentation of the new properties.

__tests__/__action_fixtures__/expandProperty-ADD_VALUE.json
```diff
@@ -110,6 +110,8 @@
               "remarkUrlLabel": null,
               "remarkUrl": null,
               "defaults": [],
+              "validationDataType": null,
+              "validationRegex": null,
```
ndushay, line 114: adding these was annoying and felt brittle, but it's over now.

static/literalDataType.json
```diff
@@ -10,9 +10,5 @@
   {
     "uri": "http://www.w3.org/2001/XMLSchema/dateTimeStamp",
     "label": "Date and time with required timezone (xsd:dateTimeStamp)"
-  },
-  {
-    "uri": "http://id.loc.gov/datatypes/edtf",
-    "label": "Extended Date/Time Format (EDTF)"
```
ndushay, line 16: postponing this for issue #3301

ndushay (commented):
I think #2890 is done with this PR except:
- PO vetting of these changes (@michelleif) - especially, e.g.  how users must enter strings like `\d` in the UI to get the right thing in the regex validation for a digit ...
- end user documentation (@michelleif )
- vetting that the tests I added in `resourceValidationHelper.tests` have the correct conditions (that non-error cases do the right thing, and that error cases do the right thing, for both addValue and updateValue)  (@justinlittman)

src/reducers/resourceValidationHelpers.js
```diff
@@ -0,0 +1,103 @@
...
+) => {
+  const errors = []
+  if (value.key === property.valueKeys[0] && propertyTemplate.required) {
+    if (!value.uri) errors.push("URI required")
+    if (!value.label) errors.push("Label required")
+  } else {
+    if (value.uri && !value.label) errors.push("Label required")
+    if (!value.uri && value.label) errors.push("URI required")
+  }
+  if (value.uri && !isValidURI(value.uri)) errors.push("Invalid URI")
```
ndushay, line 104: this method can generate multiple errors, even if the other ones can't.

src/reducers/resourceHelpers.js
```diff
@@ -181,27 +189,26 @@ export const updateValueErrors = (state, valueKey) => {
...
-      if (value.uri && !value.label) errors.push("Label required")
-      if (!value.uri && value.label) errors.push("URI required")
-    }
-    if (value.uri && !isValidURI(value.uri)) errors.push("Invalid URI")
+    errors.push(
+      validateUriPropertiesAndValue(value, property, propertyTemplate)
+    )
   }
 
+  errors = errors.flat(1) // all that pushing makes for nested arrays
```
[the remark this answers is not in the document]
mjgiarlo: If the error functions above all return (one-dimensional) arrays, you could use array destructuring for this, I think? Especially since `errors` is already a mutable value. So, something like:

```javascript
let errors = []
// ...
errors = [
  ...errors, 
  ...literalRegexValidationError(value, propertyTemplate),
  ...literalIntegerValidationError(value, propertyTemplate),
  ...literalDateTimeValidationError(value, propertyTemplate),
  ...literalDateTimeStampValidationError(value, propertyTemplate),
]
// ... 
errors = [...errors, ...uriPropertiesAndValueErrors(value, property, propertyTemplate)]
```

You can see how it works in this dumbed-down example:

```javascript
> let errors = []
> const error1 = ['one']
> const error2 = ['two', 'three']
> const error3 = []
> const error4 = ['four']
> errors = [...errors, ...error1, ...error2, ...error3, ...error4]
[ 'one', 'two', 'three', 'four' ]
```

src/reducers/resourceValidationHelpers.js
mjgiarlo: Is it worth putting the invalid URI value in the error string? Or would that be obvious to whoever uses the error text or otherwise useless?

src/reducers/resourceValidationHelpers.js
```diff
@@ -0,0 +1,106 @@
...
+  return errors
+}
+
+export const literalRegexValidationError = (value, propertyTemplate) => {
+  const errors = []
+  if (
+    propertyTemplate.validationRegex !== null &&
+    propertyTemplate.validationRegex !== ""
+  ) {
+    // NOTE: RegExp constructor sanitizes the string argument, so a string of "^\d+$" becomes /^d+$/
```
mjgiarlo, line 21: That sounds like a problem for supporting this feature.

mjgiarlo (commented):
A few comments.

PR #3297, opened 2021-11-03
Supports multiple URIs per property template.

### LINKED ISSUE #3220: add option for cataloger to choose a property
see #2763 for use cases and prior discussion

Root template change:
1) make the Property URI (<http://sinopia.io/vocabulary/hasPropertyUri>) field repeatable. This will enable the template creator add property choices for the cataloger for a particular property template, if desired.

Editor changes (applies to all kinds of property templates: literal, lookup, etc.)

1) for each field in Editor, display the property label
2) if there's more than one Property URI (in other words, template creator added repetitions of Property URI), then display all the Property URI **labels** in a dropdown for user to select.

@astridu need your design here for Editor's display of property label and display of dropdown if there's a choice of properties; needs to work for all types of inputs; this is adding an additional display/choice area--user still has to enter the "object" in the field like they do now

--for the dropdown, display the property labels in alphabetical order (order of repetitions in template is not significant)
--user cannot enter their own value, they can only choose from one of the options

justinlittman:
closes #3220 

## Why was this change made?
Introduce complexity.


src/TemplatesBuilder.js
```diff
@@ -164,10 +169,13 @@ export default class TemplatesBuilder {
...
     return {
       // This key will be unique for resource templates, property templates.
-      key: `${this.subjectTemplate.key} > ${propertyUri}`,
+      key: `${this.subjectTemplate.key} > ${Object.keys(propertyUrisMap)
+        .sort()
+        .join(", ")}`,
       subjectTemplateKey: this.subjectTemplate.key,
       label: this.valueFor(propertyTerm, rdfsLabel),
-      uri: propertyUri,
+      uris: propertyUrisMap,
```
justinlittman, line 177: This is the change in model for property templates -- a property template can have multiple property URIs.

defaultUri below is just a shortcut to getting the first. In the future, may allow configuring this in template.

justinlittman:
When multiple property URIs:
image

When single property URI:
image

justinlittman:
Also, this has been tested/approved by @michelleif

src/actionCreators/resourceHelpers.js
```diff
@@ -87,15 +87,20 @@ export const addResourceFromDataset =
   (dispatch) => {
     const subjectTerm = rdf.namedNode(chooseURI(dataset, uri))
     const newUri = asNewResource ? null : uri
-    const usedDataset = rdf.dataset()
-    usedDataset.addAll(
-      dataset.match(
+    const context = {
```
justinlittman, line 90: I declared bankruptcy on passing around to many params and moved some to context.

jcoyne (approved):

src/actionCreators/templateValidationHelpers.js
```diff
@@ -71,48 +71,55 @@ const validatePropertyTemplates = (propertyTemplates) => {
 
 const validatePropertyTemplate = (template) => {
   const errors = []
-  if (!template.uri) {
+  const firstUri = _.first(Object.keys(template.uris || {}))
```
justinlittman, line 74: Need to account for multiple property URIs.

PR #3304, opened 2021-11-04
[HOLD] allow language suppression to be set as a property attribute

### LINKED ISSUE #3274: Allow suppressing language selection for a property
Often there are literals where a language is not appropriate or relevant, like an ISBN (see original request #2837)

The goal of this ticket is to allow a template creator to specify that language selection should be suppressed for a property. When suppressed for a property, the user language should default to none (i.e., language should not be added to the RDF; this is the equivalent of the user selecting "none specified" for the language) and the user should not be able to select a language.

To accomplish this:
- [x] Add a "suppress language" value to `propertyAttributes.json`. This will allow a cataloger to add "suppress language" when specifying a property in the Editor.
- [x] Add a suppress language attribute to the property template model. The value should true when the "suppress language" attribute is present.
- [ ] When a property is set in state to have the language suppressed:
- [ ] The language should default to none, rather than "eng".
- [x] The Language Button should not be displayed.
- [x] Note that this applies to both literals and URI labels (both which can have languages).

(note that we have a new feature to support a datatype attribute for literals, and if a property template is a literal and has a datatype set, we won't set a language, since a literal cannot have both. So this new "no language" attribute would be irrelevant if property template has a datatype.)

peetucket:
## Why was this change made?

Fixes #3274 - allow users to indicate a given property template should be able to suppress language selection; if selected, don't show language selection box for that template and then set the default value to null instead of 'eng'

Note: a lot of the diff is updating the fixture data to add the new attribute.

- Updated existing tests verifying new property being set in state
- Added a new test verifying language dialog is suppressed when attribute is set
- Localhost browser

Readme


src/utilities/Utilities.js
```diff
@@ -11,8 +11,6 @@ const concatStream = require("concat-stream")
 const Readable = require("stream").Readable
 const SerializerJsonld = require("@rdfjs/serializer-jsonld-ext")
 
-export const defaultLanguageId = "eng"
-
 export const isResourceWithValueTemplateRef = (property) =>
```
peetucket, line 14: as requested by Justin L, moved this value from `Utilities` to `Config` -- and then adjustments to it's use

src/Config.js
```diff
@@ -143,6 +143,10 @@ class Config {
       // Can add additional transfer targets, e.g., discovery
     }
   }
+
+  static get defaultLanguageId() {
+    return "eng"
+  }
 }
```
peetucket, line 150: as requested by Justin L, moved this value from `Utilities` to `Config` -- and then adjustments to it's use

src/utilities/valueFactory.js
```diff
@@ -20,25 +20,33 @@ const newValue = (
...
-    defaultLanguageId,
+    propertyTemplate.languageSuppressed ? null : Config.defaultLanguageId,
     null,
     null,
     null,
     "InputLiteralValue"
   )
 
-export const newUriValue = (property, uri, label, lang = defaultLanguageId) =>
+export const newUriValue = (property, uri, label, lang) =>
```
peetucket, line 37: Note: this method is always passed a language, so we do not need defaultLanguageId here

src/utilities/valueFactory.js
```diff
@@ -20,25 +20,33 @@ const newValue = (
   component,
 })
 
-export const newLiteralValue = (property, literal, lang = defaultLanguageId) =>
+export const newLiteralValue = (property, literal, lang) =>
```
peetucket, line 23: Note: this method is always passed a language, so we do not need defaultLanguageId here

src/reducers/resources.js
```diff
@@ -512,14 +512,14 @@ const addFirstValue = (state, propertyKey) => {
...
+        newBlankLiteralValue(property, propertyTemplate.languageSuppressed),
         null,
         propertyTemplate.required
       )
     case "InputURI":
       return addValueToNewState(
         state,
-        newBlankUriValue(property),
+        newBlankUriValue(property, propertyTemplate.languageSuppressed),
         null,
```
peetucket, line 523: we need to pass in this property of the template so we know how to set the default language correctly

src/reducers/resources.js
```diff
@@ -512,14 +512,14 @@ const addFirstValue = (state, propertyKey) => {
     case "InputLiteral":
       return addValueToNewState(
         state,
-        newBlankLiteralValue(property),
+        newBlankLiteralValue(property, propertyTemplate.languageSuppressed),
         null,
```
peetucket, line 516: we need to pass in this property of the template so we know how to set the default language correctly

__tests__/feature/editing/editingLiteral.test.js
```diff
@@ -177,6 +177,25 @@ describe("editing a literal property", () => {
...
+    renderApp(null, history)
+
+    await screen.findByText("Uber template1", { selector: "h3" })
+
+    // Add a value
+    const input = screen.getByPlaceholderText("Uber template1, property9")
+    fireEvent.change(input, { target: { value: "foo" } })
+    fireEvent.keyDown(input, { key: "Enter", code: 13, charCode: 13 })
+
+    // There is foo text.
```
justinlittman, line 190: This test isn't necessary.

__tests__/feature/editing/editingLiteral.test.js
```diff
@@ -177,6 +177,25 @@ describe("editing a literal property", () => {
...
+    await waitFor(() =>
+      expect(screen.getByText("foo")).toHaveClass("form-control")
+    )
+    // There is not a language button.
+    const langBtn = screen.queryByTestId("Change language for foo")
+    expect(langBtn).toBeNull()
```
justinlittman, line 196: Could just be `expect(screen.queryByTestId("Change language for foo")).not.toBeInTheDocument()`

src/components/vocabulary/Vocab.js
```diff
@@ -111,6 +111,10 @@ const vocabulary = {
     description: "Values are ordered",
     url: "http://sinopia.io/vocabulary/propertyAttribute/ordered",
   },
+  "propertyAttribute/supressLanguage": {
```
justinlittman, line 114: Typo.

justinlittman (changes requested):
Minor tweaks, please.

__tests__/feature/editing/editingLiteral.test.js
peetucket: are you suggesting the entire test is not necessary, or the setup of filling in the field?  My interpretation of the reason we are doing the first part of the test of filling text into the field is to provide a unique testId for the language button to ensure the expectation (for it to be there or not) is for the field we are testing.  In this case, it is `Uber template1, property9`, which I adjusted in the fixtures to have language suppressed in order to test here.  If you change the field that is filled in, this test will fail because the language button will be shown for that field (as it should).  I adjusted the text being filled in to perhaps make this more clear.

justinlittman: Sorry, L190-193 aren't necessary.

peetucket: Got it, removed.

justinlittman (approved):

justinlittman:
I'm doing a major refactoring of some tests and this will cause a massive merge conflict. Let's hold this and I may create a new PR for merging this.

PR #3311, opened 2021-11-08
set z-index for dropdown so it stays on top

### LINKED ISSUE #3310: Change or remove sticky header
The current sticky header in the Editor is causing problems in which things are stuck below it. For example:

image

Please try to change the sticky header so that it works like a Github ticket (only becomes sticky when scrolled to top) or remove the sticky feature.

peetucket:
## Why was this change made?

Fixes #3310:

- adjust zindex for drop-downs to ensure they stay on top
- adjust scroll position for left nav links to account for header

**Drop down menu no longer obscured:**

Screen Shot 2021-11-08 at 4 36 56 PM

**After clicking "Template Label" in left nav, scroll position set correctly:**

Screen Shot 2021-11-08 at 4 37 03 PM

Localhost

### DIFF
--- a/src/hooks/useNavLink.js
+++ b/src/hooks/useNavLink.js
@@ -37,7 +37,10 @@ const useNavLink = (navObj) => {
     event.preventDefault()
 
     const elem = document.querySelector(`#${navTargetId}`)
-    if (!isInViewport(elem)) elem.scrollIntoView({ behavior: "smooth" })
+    if (!isInViewport(elem)) {
+      const y = elem.getBoundingClientRect().top + window.pageYOffset - 90 // do not scroll to top, leave space for header row
+      window.scrollTo({ top: y, behavior: "smooth" })
+    }
     dispatch(
       setCurrentComponent(
         navObj.rootSubjectKey,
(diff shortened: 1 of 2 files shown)


src/hooks/useNavLink.js
```diff
@@ -37,7 +37,10 @@ const useNavLink = (navObj) => {
     event.preventDefault()
 
     const elem = document.querySelector(`#${navTargetId}`)
-    if (!isInViewport(elem)) elem.scrollIntoView({ behavior: "smooth" })
+    if (!isInViewport(elem)) {
+      const y = elem.getBoundingClientRect().top + window.pageYOffset - 90 // do not scroll to top, leave space for header row
```
jcoyne, line 41: Is this appropriate if the title spills onto multiple lines?

src/styles/bootstrap-override.scss
```diff
@@ -25,6 +25,10 @@ $primary-disabled: lighten(map-get($theme-colors, "secondary"), 20%);
   text-decoration: underline;
 }
 
+.dropdown-menu {
+  z-index: 3000;
```
jcoyne, line 29: Can this be `$zindex-sticky + 1` so we can see this is relative to that value.

jcoyne (changes requested):

src/hooks/useNavLink.js
peetucket: hmm, excellent question - I'm trying to figure out how to account for that (and to find an example of a label that spans lines)  ... shrinking window size doesn't wrap any headers I can find

peetucket:
Closed in favor of #3317

PR #3312, opened 2021-11-08
Replace swagger-based QA client with fetch calls

### LINKED ISSUE #3262: Rewrite QA client
Currently our QA client is implemented using `swagger-client`, which dynamically builds a client based on a local openapi spec. This is problematic because:
* The client code is complex and hard to make changes.
* The local openapi spec is not maintained.

For this ticket, rewrite the QA client to be more consistent with our other API clients. It should retain the existing functionality, as well as adding the functionality described in #3189.

mjgiarlo:
## Why was this change made?

Fixes #3262

This commit replaces the swagger-based QA client. Why?

1. the API specification is incomplete;
2. it's not clear there is a plan in place to keep the specification up to date; and
3. we only use a fraction of the QA API in the Sinopia editor

In short, it's to pay down technical debt. This commit otherwise retains the same interface for interacting with the QA API, returning a promise that resolves to the HTTP response.

CI

None


jcoyne:
I'm quite against the lack of a shared and validatable contract about this API that is maintained by two entities.  I've recently worked with Lynette to get the API matching the swagger spec they are publishing. I think we should continue to use it.

mjgiarlo:
@jcoyne 💬 
> I'm quite against the lack of a shared and validatable contract about this API that is maintained by two entities. I've recently worked with Lynette to get the API matching the swagger spec they are publishing. I think we should continue to use it.

After chatting with @justinlittman this morning, it seems the fetch-based approach is preferred at this time. Maybe a stand-up discussion is warranted.

justinlittman (approved):

PR #3317, opened 2021-11-10
Rejiggers sticky header.

### LINKED ISSUE #3310: Change or remove sticky header

justinlittman:
closes #3310

## Why was this change made?
So the sticky header functions as hoped / intended.

Local

### DIFF
--- a/src/utilities/Utilities.js
+++ b/src/utilities/Utilities.js
@@ -168,10 +168,13 @@ export const isTopInViewport = (elem) => {
   )
 }
 
-export const stickyScrollIntoView = (selector) => {
-  const elem = document.querySelector(selector)
+export const stickyScrollIntoView = (
+  targetSelector,
+  headerSelector = "#sticky-resource-header"
+) => {
+  const elem = document.querySelector(targetSelector)
   if (isTopInViewport(elem)) return
-  const stickyHeader = document.querySelector("#sticky-resource-header")
+  const stickyHeader = document.querySelector(headerSelector)
   // do not scroll to top, leave space for sticky header
   const y =
     elem.getBoundingClientRect().top +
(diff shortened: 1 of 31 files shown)


src/utilities/Utilities.js
```diff
@@ -156,3 +156,26 @@ export const isInViewport = (elem) => {
...
+  if (!elem) return false
+  const bounding = elem.getBoundingClientRect()
+  return (
+    bounding.top >= 0 &&
+    bounding.top <=
+      (window.innerHeight || document.documentElement.clientHeight)
+  )
+}
+
+export const stickyScrollIntoView = (selector) => {
```
jcoyne, line 171: Can you parameterize the selector for the stickyHeader too?  maybe `(targetSelector, headerSelector = "#sticky-resource-header")`

peetucket:
Seems more flexible than my approach in allowing for variable width headers in the smooth scroll.  Do we still want to ensure the z-index of the `.dropdown-menu` class still wins? (https://github.com/LD4P/sinopia_editor/pull/3311/files#diff-d6c5465dec7690d11d69e5841d4517dcea9eddae3932c3be344267fad22a017dR28-R31)

Also, looks like a number of test failures.

jcoyne:
FWIW, I think that setting the zindex for the dropdown menus is way easier to comprehend than the code that determines if the main header is visible.

src/utilities/Utilities.js
peetucket: Done.

peetucket:
Fixed the tests by ensuring the selectors that look for the headers are unique (now that the sticky header adds a second copy).

peetucket (approved):

PR #3320, opened 2021-11-11
remove unnecessary import

peetucket:
## Why was this change made?

Import may not needed and causes problems

Localhost


mjgiarlo (approved):
Not sure if unnecessary or not, but happy to approve.

peetucket:
Tests pass and works on localhost for me, but not sure if I am hitting that part of the code or not in my localhost browser.

PR #3321, opened 2021-11-12
allow language suppression to be set as a property attribute

### LINKED ISSUE #3274: Allow suppressing language selection for a property

justinlittman:
closes #3274

## Why was this change made?
Same code as #3304 (with a few tweaks); refactored tests.


justinlittman:
@peetucket Can you take a look at this and make sure I got it right?

src/components/editor/inputs/InputLiteralValue.jsx
```diff
@@ -69,6 +69,8 @@ const InputLiteralValue = ({
     handleKeyDownDiacritics(event)
   }
 
+  const showLang = !propertyTemplate.languageSuppressed || value.lang
+
```
peetucket, line 73: +1

peetucket (approved):
LGTM

PR #3322, opened 2021-11-12
Literal validation (integer, dateTime and dateTimeStamp and regex)

justinlittman:
## Why was this change made?
This is based on #3296 but has refactored tests.

The only other substantive changes are:
* Removed having validation keys on values; they are only on property templates.
* Added a test template (`resourceTemplate:testing:literalValidation`)


justinlittman:
@ndushay Can you take a look at this and make sure I got it right?

README.md
```diff
@@ -313,9 +313,10 @@ The following are only in the resource subject (that is, the base subject).
   remarkUrlLabel: <a descriptive label about the remark url, e.g., "Title Proper">
   type: <resource | uri | literal>,
   component: <InputLookup | InputLookupQA | InputList | InputLiteral | InputURI>,
-  valueSubjectTemplateKeys: [<subject template keys>]
-  authorities: [{authority}, ...]
-
+  valueSubjectTemplateKeys: [<subject template keys>],
+  authorities: [{authority}, ...],
+  validationRegex: <regex for literal validation, as a string with backslashes double-escaped (e.g. "^\\d+$" becomes /^\d+$/)>,
```
ndushay, line 318: I'm assuming this was confirmed by Michele as correct?  We didn't get that far before my vacation.

ndushay (approved):
Looks great to me, but maybe I'm biased.

README.md
justinlittman: I tested this out and I think it handles the escaping just fine. I don't believe that the user has to escape when creating template.

PR #3326, opened 2021-11-12
Change the input alignment for inputURI

### LINKED ISSUE #3319: Change Input alignment for inputURI
image

jcoyne:
## Why was this change made?

Fixes #3319

Screen Shot 2021-11-12 at 2 37 34 PM

### DIFF
--- a/src/components/editor/inputs/InputURIValue.jsx
+++ b/src/components/editor/inputs/InputURIValue.jsx
@@ -123,10 +123,10 @@ const InputURIValue = ({
     <React.Fragment>
       <ValuePropertyURI propertyTemplate={propertyTemplate} value={value} />
       <div className="row my-2">
-        <div className="col">
-          <label htmlFor={uriId} className="form-label">
-            URI
-          </label>
+        <label htmlFor={uriId} className="col-sm-1 col-form-label">
+          URI
+        </label>
+        <div className="col-sm-11">
           <TextareaAutosize
             required={propertyTemplate.required}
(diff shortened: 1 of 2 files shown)


justinlittman:
Can you shrink the property select (so that it stays consistent with the property select for other inputs)?
image

jcoyne:
@justinlittman I didn't change the property select value. So I believe it's still consistent with how the other properties display.  If you shrink it, you wouldn't be able to read the property that is currently set.

justinlittman:
It was changed from a `col-auto` to `col-sm-11`: https://github.com/LD4P/sinopia_editor/pull/3326/files#diff-b74052ea82c9b90e614e48590026aa71aba4cb3ad89618592b3379e38ee08d30R45

jcoyne:
Yes, but col-auto allowed it to show the enirity of what was selected like this
Screen Shot 2021-11-13 at 8 34 01 AM

Cropping it down would make it more difficult to see what they had selected as the URL would get cropped.

justinlittman:
I'm confused. It was previously `col-auto`, which allowed the select to resize as necessary. However, in this PR the select was changed to `col-sm-11`, which made it cover the width. The preference is for the select to be sized to the options, not cover the width.

jcoyne:
Yes, the out of date mockups were causing confusion. I've made that auto now so it looks like this:
Screen Shot 2021-11-15 at 10 17 10 AM

justinlittman (approved):

PR #3330, opened 2021-11-16
Allow duplicate property templates with different classes.

### LINKED ISSUE #3221: allow multiple fields of type nested resource to have the same property (provided the nested resources are of different classes)
_this was the original discussion; we will implement the solution described in title_

use case: 

the property bf:contribution has two possible objects, each represented by its own template: 
bflc:PrimaryContribution
and
bf:Contribution

right now Sinopia doesn't allow two fields with the same property, so we can't make separate fields for
<> <bf:contribution> <bflc:PrimaryContribution>
and
<> <bf:contribution> <bf:Contribution>

instead they're both nested in one field (see template pcc:bf2:Monograph:Work on stage):

problem is, cataloging practice specifies that there can be only one Primary Contribution. but there can be multiple Contributions, and there's no way to make the repeatability different for each nested resource

possible solutions:
1) allow there to be more than one field with same property (then there could be a separate field Primary Contribution vs Contribution, one repeatable, one not repeatable)

2) allow specification of whether each nested template is repeatable, at the nested template level, not at the property level, so you can have different repeatability of mulitple nested templates for a field

other ideas?

justinlittman:
closes #3221

## Why was this change made?
Better modeling.

Unit

### DIFF
--- a/__tests__/testUtilities/fixtureLoaderHelper.js
+++ b/__tests__/testUtilities/fixtureLoaderHelper.js
@@ -1,5 +1,8 @@
 import TemplatesBuilder from "../../src/TemplatesBuilder"
-import { datasetFromJsonld } from "../../src/utilities/Utilities"
+import {
+  datasetFromJsonld,
+  resourceToName,
+} from "../../src/utilities/Utilities"
 import _ from "lodash"
 
 const resourceFilenames = {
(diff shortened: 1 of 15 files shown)


__tests__/testUtilities/fixtureLoaderHelper.js
```diff
@@ -46,21 +49,25 @@ const templateFilenames = {
...
   )
+}
 
 export const getFixtureResource = (uri) => {
   // A special URI for testing.
   if (uri === "http://error") throw new Error("Ooops")
   if (uri === "http://localhost:3000/resource/ld4p:RT:bf2:xxx")
     throw new Error("Error retrieving resource: Not Found")
-  const id = normUri(uri)
+  const id = resourceToName(uri)
```
justinlittman, line 70: Switched to using this utility function in a few places.

src/TemplatesBuilder.js
```diff
@@ -224,7 +225,10 @@ export default class TemplatesBuilder {
       propertyTemplate.valueSubjectTemplateKeys = this.valuesFor(
         attributeTerm,
         "http://sinopia.io/vocabulary/hasResourceTemplateId"
-      )
+      ).map((subjectTemplateKey) => resourceToName(subjectTemplateKey))
```
justinlittman, line 228: Keys get more complicated for nested resource properties in order to make them unique.

src/actionCreators/templateValidationHelpers.js
```diff
@@ -101,28 +104,87 @@ const validatePropertyTemplate = (template) => {
...
-  const dupes = []
-  const propertyTemplateUris = []
-  propertyTemplates.forEach((propertyTemplate) => {
-    if (!_.isEmpty(propertyTemplate.uris)) {
-      Object.keys(propertyTemplate.uris).forEach((uri) => {
-        if (propertyTemplateUris.includes(uri)) {
-          dupes.push(uri)
-        } else {
-          propertyTemplateUris.push(uri)
+const validateRepeatedPropertyTemplates =
```
justinlittman, line 107: Yes, the validation of repeated property templates gets really complicated.

jcoyne (approved):

PR #3331, opened 2021-11-16
literal with validationDataType does not display language button

### LINKED ISSUE #3287: when literal has validataionDataType, do not display language
Per https://www.w3.org/TR/rdf11-concepts/#section-Graph-Literal, a literal can have a datatype or a language.

Thus, when a literal has a datatype (i.e., the property template for the value's property specifies a datatype), do not display the LanguageButton.

ndushay:
## Why was this change made?

Fixes #3287

When a property template has a literal which has a validationDataType specified, then a resource using that property template should not display the language button for the literal with the validationDataType.

### After

Look!  no language button!

image

from this property template:

image

with a new unit test, and locally (see pix above)

### DIFF
--- a/src/TemplatesBuilder.js
+++ b/src/TemplatesBuilder.js
@@ -129,8 +129,11 @@ export default class TemplatesBuilder {
     propertyTemplate.defaults = this.defaultsForLiteral(propertyTerm)
     propertyTemplate.validationRegex =
       this.validationRegexForLiteral(propertyTerm)
-    propertyTemplate.validationDataType =
-      this.validationDataTypeForLiteral(propertyTerm)
+    const validationDataType = this.validationDataTypeForLiteral(propertyTerm)
+    if (validationDataType) {
+      propertyTemplate.languageSuppressed = true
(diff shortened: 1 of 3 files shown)


src/TemplatesBuilder.js
```diff
@@ -129,8 +129,11 @@ export default class TemplatesBuilder {
     propertyTemplate.defaults = this.defaultsForLiteral(propertyTerm)
     propertyTemplate.validationRegex =
       this.validationRegexForLiteral(propertyTerm)
-    propertyTemplate.validationDataType =
-      this.validationDataTypeForLiteral(propertyTerm)
+    if (this.validationDataTypeForLiteral(propertyTerm)) {
```
justinlittman, line 132: Can you assign `this.validationDataTypeForLiteral(propertyTerm)` to a variable instead of invoking twice?

justinlittman (changes requested):
Minor tweak

ndushay:
tweaked per @justinlittman's suggestion.

justinlittman (approved):

PR #3333, opened 2021-11-16
Multiple classes per resource

### LINKED ISSUE #2785: Pick lists or drop downs for subclasses (PCC priority no. 5)
Many of the bf classes we create templates for have multiple subclasses. Currently, to make use of those subclasses, we need to create individual templates for each subclass. This is particularly an issue with bf:Instance, which has multiple possible subclasses for a book template alone, but where we have good reason to only want to create one template representing the instance (other classes include bf:Work, bf:SoundCharacteristic, bf:DigitialCharacteristic, bf:Identifier). 

It would be easier for the template creator and the cataloger and result in fewer templates if there could be a pick list or dropdown for subclasses of bf classes. Then one could create a generic bf:Instance template (as an example), but still allow the choice of subclass--bf:Electronic, bf:Print, bf:Manuscript, bf:Tactile.

justinlittman:
closes #2785

## Why was this change made?
Better resource modeling, PCC request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -237,6 +237,7 @@ The sinopia_editor code displays the properties and classes prefixed "http://sin
   -> properties: [{property}, ...]
   labels = [labels of property/resource templates of self and ancestors, ...]
   showNav: <true | false>
+  classes: [resource type URI, e.g., http://id.loc.gov/ontologies/bibframe/Instance, ...]
 }
 ```
 -> Added by selector, not stored in state.
@@ -259,7 +260,8 @@ The following are only in the resource subject (that is, the base subject).
   key: <resource template id, e.g., resourceTemplate:bf2:Monograph:Instance>,
   id: <resource template id, e.g., resourceTemplate:bf2:Monograph:Instance>,
   uri: <resource template uri, e.g. http://datastore/resource/resourceTemplate:bf2:Monograph:Instance>,
-  class: <resource URI, e.g., http://id.loc.gov/ontologies/bibframe/Instance>,
+  class: <required resource type URI, e.g., http://id.loc.gov/ontologies/bibframe/Instance>,
+  classes: {resource type URI: resource URI label, ... This includes required resource and optional URIs}
   label: <resource label, e.g., "BIBFRAME Instance">,
   author: <author>,
   remark: <remark>,
--- a/__tests__/testUtilities/resourceBuilderUtils.js
+++ b/__tests__/testUtilities/resourceBuilderUtils.js
@@ -8,9 +8,11 @@ export default class ResourceBuilder {
   constructor({
     injectPropertyKeyIntoValue = false,
     injectPropertyIntoValue = false,
+    injectClassesIntoSubject = false,
   } = {}) {
     this.injectPropertyKeyIntoValue = injectPropertyKeyIntoValue
     this.injectPropertyIntoValue = injectPropertyIntoValue
+    this.injectClassesIntoSubject = injectClassesIntoSubject
   }
 
   key() {
@@ -20,6 +22,7 @@ export default class ResourceBuilder {
   subjectTemplate({
     id,
     clazz,
+    classes = null,
     uri = null,
     label = null,
     author = null,
@@ -36,6 +39,7 @@ export default class ResourceBuilder {
       key: id,
       id,
       class: clazz,
+      classes: classes || { [clazz]: clazz },
       uri,
       label,
       author,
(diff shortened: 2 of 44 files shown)


README.md
```diff
@@ -259,7 +259,8 @@ The following are only in the resource subject (that is, the base subject).
   key: <resource template id, e.g., resourceTemplate:bf2:Monograph:Instance>,
   id: <resource template id, e.g., resourceTemplate:bf2:Monograph:Instance>,
   uri: <resource template uri, e.g. http://datastore/resource/resourceTemplate:bf2:Monograph:Instance>,
-  class: <resource URI, e.g., http://id.loc.gov/ontologies/bibframe/Instance>,
+  class: <required resource URI, e.g., http://id.loc.gov/ontologies/bibframe/Instance>,
+  classes: {resource URI: resource URI label, ... This includes required resource and optional URIs}
```
justinlittman, line 263: Notice the addition of classes to the subject template state model.

src/GraphBuilder.js
```diff
@@ -28,13 +28,15 @@ export default class GraphBuilder {
   }
 
   buildSubject(subject, subjectTerm) {
-    this.dataset.add(
-      rdf.quad(
-        subjectTerm,
-        rdf.namedNode("http://www.w3.org/1999/02/22-rdf-syntax-ns#type"),
-        rdf.namedNode(subject.subjectTemplate.class)
+    subject.classes.forEach((clazz) => {
```
justinlittman, line 31: Add the classes from the subject (i.e., the classes selected by the user), not the subject template.

src/TemplatesBuilder.js
```diff
@@ -67,6 +68,25 @@ export default class TemplatesBuilder {
     }
   }
 
+  buildClasses() {
```
justinlittman, line 71: Need the optional classes and the default class.

src/actionCreators/resourceHelpers.js
```diff
@@ -214,6 +209,28 @@ export const recursiveResourceFromDataset =
       )
     )
 
+const newSubjectFromDataset =
```
justinlittman, line 212: Need to add the classes from the dataset to the subject model in state. Previously, the class from the subject template was used.

src/actionCreators/templateValidationHelpers.js
```diff
@@ -260,27 +260,35 @@ const validateUniqueResourceURIs =
...
+            Object.keys(subjectTemplate.classes),
             subjectTemplate.id,
           ])
           .catch(() => {
             /* nothing */
           })
       )
     ).then((results) => {
-      const classResourceTemplateIds = {}
+      // No other nested template can have (required) class or optional class that is the same as this (required) class.
```
justinlittman, line 271: Once again, template validation got more complex.

jcoyne: 🤕

src/components/vocabulary/Vocab.js
```diff
@@ -17,9 +17,13 @@ const vocabulary = {
     url: "http://sinopia.io/vocabulary/hasAuthority",
   },
   hasClass: {
-    description: "The RDF class for a property",
+    description: "The RDF class for a resource",
```
ndushay, line 20: thanks for this correction!

src/reducers/resources.js
```diff
@@ -163,6 +163,10 @@ const addSubjectToNewState = (state, subject, valueSubjectOfKey) => {
     newSubject.labels = [subjectTemplate.label]
   }
 
+  // Add classes if it doesn't already have.
```
ndushay, line 166: "if you read the code out loud and it is the same as what you put in the comment ... then you don't need the comment" -- from Sandi Metz when she did OOP class at Stanford.    But we have plenty of these comments, so no biggie.

static/templates/rt_resource_template_doc.json
```diff
@@ -154,6 +154,35 @@
...
+      }
+    ],
+    "http://sinopia.io/vocabulary/hasRemark": [
+      {
+        "@value": "Optional classes for the resource."
+      }
+    ],
+    "http://sinopia.io/vocabulary/hasPropertyUri": [
+      {
+        "@id": "http://sinopia.io/vocabulary/hasOptionalClass"
```
ndushay, line 172: I suspect I missed it ... but is there a test that exercises this?  Do we update a fixture with this or otherwise exercise this property?   I saw plenty with multiple classes, so yes, implied ... but ??

ndushay (approved):
Will this require a migration of existing data?

Is this PR related to a ticket?

I am marking approved, because I don't understand this PR fully, but I have a couple questions.

ndushay:
merge conflicts - maybe due to the most recent "finish implementing literal dataValidationType" ?

__tests__/testUtilities/resourceBuilderUtils.js
```diff
@@ -115,24 +121,36 @@ export default class ResourceBuilder {
...
+  orderedSubject({ properties, uri = null, classes = null }) {
     assertProps({ properties })
-    return {
+    const subject = {
       key: this.key(),
       properties,
       uri,
     }
+
+    if (classes) subject.classes = classes
```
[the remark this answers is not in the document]
justinlittman: Fixed.

src/reducers/resources.js
justinlittman: That's assuming that your future self remembers how read JS code ...

static/templates/rt_resource_template_doc.json
justinlittman: Yes, the optional classes are what allows multiple classes.

justinlittman:
Added related ticket in description.

src/GraphBuilder.js
ndushay: If this needs a comment - maybe put it in the code?

src/TemplatesBuilder.js
ndushay: should this be a code comment?

src/actionCreators/resourceHelpers.js
ndushay: should this be a comment?  "Need to add the classes from the dataset to the subject model in state"

PR #3342, opened 2021-11-17
Replace swagger-based QA client with simple fetch calls

### LINKED ISSUE #3262: Rewrite QA client

justinlittman:
Fixes #3262

This commit replaces the swagger-based QA client. Why?

1. the API specification is incomplete;
2. it's not clear there is a plan in place to keep the specification up to date; and
3. we only use a fraction of the QA API in the Sinopia editor

In short, it's to pay down technical debt. This commit otherwise retains the same interface for interacting with the QA API, returning a promise that resolves to the HTTP response.

## Why was this change made?
This is an update on @mjgiarlo's earlier PR.


__tests__/utilities/QuestioningAuthority.test.js
```diff
@@ -61,22 +58,78 @@ describe("createLookupPromise()", () => {
...
+    it("returns a promise from a search", async () => {
+      const authorityConfig = findAuthorityConfig("urn:ld4p:qa:geonames:water")
+      await createLookupPromise("Artic Sea", authorityConfig)
+      expect(global.fetch).toHaveBeenCalledWith(
+        "https://lookup.ld4l.org/authorities/search/linked_data/geonames_ld4l_cache/water?q=Artic+Sea&maxRecords=8&lang=en&context=true&response_header=true&startRecord=1"
+      )
+    })
+  })
+
+  describe("when non-LD authority", () => {
```
ndushay, line 89: should there be a test with subauthority for this one?   You mention 4 types in the class comments, so ... four cases in the test??

ndushay (approved):
LGTM; one quibble and a question.  You prolly want Mike to review, yes?

src/utilities/QuestioningAuthority.js
```diff
@@ -1,62 +1,65 @@
...
-  const subauthority = lookupConfig.subauthority
-  const language = lookupConfig.language
-
+const baseUrlFromConfig = (nonldLookup, authority, subauthority) => {
   /*
-   *  There are two types of lookup: linked data and non-linked data. The API calls
-   *  for each type are different, so check the nonldLookup field in the lookup config.
-   *  If the field is not set, assume false.
+   * There are four types of lookup: linked data and non-linked data, authority
+   * andsubauthority. The API calls for each type are different, so construct
```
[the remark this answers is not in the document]
justinlittman: Fixed.

__tests__/utilities/QuestioningAuthority.test.js
justinlittman: This actually is an authority with subauthority. In addition, all of our non-LD authorities have subauthorities.

Made this clearer in test.

PR #3343, opened 2021-11-17
Upgrade JSON-LD library

justinlittman:
## Why was this change made?
Security.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -90,6 +90,8 @@ jobs:
           path: /home/circleci/sinopia_editor/screenshots
   test:
     executor: default
+    environment:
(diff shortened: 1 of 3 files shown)


src/reducers/resources.js
```diff
@@ -1,5 +1,4 @@
 import _ from "lodash"
-import { resourceEditErrorKey } from "components/editor/Editor"
```
justinlittman, line 2: This was causing the Editor to be loaded in running tests, which was leading to an open handle. Why did this become a problem when JSON-LD library was updated? No idea whatsoever.

mjgiarlo (approved):
You did it! 🏆

PR #3349, opened 2021-11-17
display labels for literals

### LINKED ISSUE #2892: display help text for literals with datatype or regex
consider adding help text for literal fields such as the below. (but see also #2906 about overhauling input components, which might have a better design)

for literal fields without a datatype, display text above the empty field:
"Enter a literal"
for literal fields *with* a datatype (see #2889) display text above the empty field:
"Enter a literal of type [datatype]" (ideally, [datatype] could be a link to that datatype's URI)
for literal fields *with* a regex specification (see #2907), include specific help text?

See https://github.com/LD4P/sinopia_editor/issues/2890#issuecomment-961359007 and other comments nearby

peetucket:
## Why was this change made?

Fixes #2892 - display help text next to literals based on validation type 

Issues:
- this label shows up when editing resource templates as well, perhaps unnecessarily (e.g. when entering a validation regex, it says "Enter a literal")

Localhost and new component test


src/components/editor/inputs/InputLiteralValue.jsx
```diff
@@ -69,7 +70,9 @@ const InputLiteralValue = ({
     handleKeyDownDiacritics(event)
   }
 
-  const showLang = !propertyTemplate.languageSuppressed || value.lang
+  const showLang =
+    (!propertyTemplate.languageSuppressed || value.lang) &&
+    !propertyTemplate.validationDataType
```
peetucket, line 76: don't show the language field if we have validating a type (e.g. integer or datetime) as noted in the ticket

src/components/editor/inputs/InputLiteralValue.jsx
```diff
@@ -78,6 +81,7 @@ const InputLiteralValue = ({
   return (
     <React.Fragment>
       <ValuePropertyURI propertyTemplate={propertyTemplate} value={value} />
+      <LiteralTypeLabel propertyTemplate={propertyTemplate} />
       <div className="row my-2">
```
peetucket, line 85: new component below for showing the label

src/components/editor/property/LiteralTypeLabel.jsx
```diff
@@ -0,0 +1,29 @@
...
+      if (propertyTemplate.validationDataType) {
+        return `Enter a "${propertyTemplate.validationDataType}" in the form "${propertyTemplate.validationRegex}"`
+      }
+      return `Enter a literal in the form "${propertyTemplate.validationRegex}"`
+    }
+    if (propertyTemplate.validationDataType) {
+      return `Enter a ${propertyTemplate.validationDataType}`
+    }
+    return "Enter a literal"
+  }
```
peetucket, line 16: 1. is there a better way of determining the  literal type other than checking the "validationDataType" ?

2. the validation data type is stored as a URI and this is what is displayed to the user ... is there a better label to show?  eg. as in https://github.com/LD4P/sinopia_editor/blob/main/static/literalDataType.json    if so, how to access that label?

src/components/editor/property/LiteralTypeLabel.jsx
```diff
@@ -0,0 +1,34 @@
+import React from "react"
+import PropTypes from "prop-types"
+
+const LiteralTypeLabel = ({ propertyTemplate }) => {
+  const labelText = (propertyTemplate) => {
+    const typeLabelLookup = {
+      "http://www.w3.org/2001/XMLSchema/integer": "integer",
+      "http://www.w3.org/2001/XMLSchema/dateTime": "date time",
+      "http://www.w3.org/2001/XMLSchema/dateTimeStamp":
```
jcoyne, line 9: This URI doesn't resolve. Is it a typo?

justinlittman: According to https://www.w3.org/TR/rdf11-concepts/#section-Datatypes, it should be `http://www.w3.org/2001/XMLSchema#<DATETYPE>`, e.g., `http://www.w3.org/2001/XMLSchema#integer`.

Can you make this change here and everywhere else in the codebase?

__tests__/components/editor/property/LiterTypeLabel.test.js
```diff
@@ -0,0 +1,84 @@
...
+
+describe("<LiteralTypeLabel />", () => {
+  it("displays the correct text for unvalidated literal", () => {
+    const state = createState({ hasTwoLiteralResources: true })
+    const store = mockStore(state)
+    const propertyTemplate =
+      state.entities.propertyTemplates[
+        "ld4p:RT:bf2:Title:AbbrTitle > http://id.loc.gov/ontologies/bibframe/mainTitle"
+      ]
+    const label = render(
```
justinlittman, line 20: Can you use `renderComponent()` from `testUtils`? It will take care of the provider for you.

__tests__/components/editor/property/LiterTypeLabel.test.js
```diff
@@ -0,0 +1,84 @@
...
+      <Provider store={store}>
+        <LiteralTypeLabel propertyTemplate={propertyTemplate} />
+      </Provider>
+    )
+    expect(label.getByText("Enter a literal")).toBeInTheDocument
```
justinlittman, line 25: Note that `toBeInTheDocument` is a function, so must be called as `toBeInTheDocument()`.

However, in this case `label.getByText("Enter a literal")` is adequate, because it will raise if no match. (`expect` not needed.)

__tests__/components/editor/property/LiterTypeLabel.test.js
```diff
@@ -0,0 +1,84 @@
...
+    const propertyTemplate =
+      state.entities.propertyTemplates[
+        "ld4p:RT:bf2:Title:AbbrTitle > http://id.loc.gov/ontologies/bibframe/mainTitle"
+      ]
+    const label = render(
+      <Provider store={store}>
+        <LiteralTypeLabel propertyTemplate={propertyTemplate} />
+      </Provider>
+    )
+    expect(label.getByText("Enter a integer")).toBeInTheDocument
```
justinlittman, line 43: "an integer"?

__tests__/components/editor/property/LiterTypeLabel.test.js
justinlittman: Same below as well.

justinlittman (changes requested):

src/components/editor/property/LiteralTypeLabel.jsx
peetucket: According to https://www.w3.org/TR/rdf11-concepts/#section-Datatypes, should it be `http://www.w3.org/2001/XMLSchema#xsd:integer` instead of `http://www.w3.org/2001/XMLSchema#integer` ?

justinlittman: I've poked around the internets. It is `http://www.w3.org/2001/XMLSchema#integer`.

peetucket: ok, changed

__tests__/components/editor/property/LiterTypeLabel.test.js
peetucket: ok, found this pattern in a few other tests, will do a separate PR for those

__tests__/components/editor/property/LiterTypeLabel.test.js
```diff
@@ -0,0 +1,75 @@
+// Copyright 2021 Stanford University see LICENSE for license
+
+import React from "react"
+import { createStore, renderComponent } from "testUtils"
+import { createState } from "stateUtils"
+import LiteralTypeLabel from "components/editor/property/LiteralTypeLabel"
+
+describe("<LiteralTypeLabel />", () => {
```
justinlittman, line 8: Can you fix the name of this file? (LiterTypeLabel.test.js)

justinlittman (changes requested):

__tests__/components/editor/property/LiterTypeLabel.test.js
peetucket: fixed

justinlittman (approved):

PR #3350, opened 2021-11-17
GraphBuilder: add datatype declaration to literal value when present

### LINKED ISSUE #2891: add datatype declaration to literal statements, if datatype specified by template
when user enters data in literal field that has a datatype (see #2889), add datatype IRI to resulting RDF statement, preceded by '^^' (U+005E U+005E)

example for datatype EDTF, property URI: http://id.loc.gov/ontologies/bibframe/originDate, value entered: 1964/1966:
`
<>  <http://id.loc.gov/ontologies/bibframe/originDate> "1964/1966"^^<http://id.loc.gov/datatypes/edtf> .
`

ndushay:
## Why was this change made?

Closes #2891

When there is a validationDataType, the URI is now included in the RDF as the @type of the literal property's value.

### After

we indicate the datatype in the graph:

image

unit test and locally on my laptop

### DIFF
--- a/__tests__/GraphBuilder.test.js
+++ b/__tests__/GraphBuilder.test.js
@@ -2,7 +2,10 @@ import GraphBuilder from "GraphBuilder"
 import ResourceBuilder from "resourceBuilderUtils"
 
 describe("GraphBuilder", () => {
-  const build = new ResourceBuilder({ injectPropertyIntoValue: true })
+  const build = new ResourceBuilder({
+    injectPropertyTemplateIntoValue: true,
+    injectPropertyIntoValue: true,
+  })
   describe("graph()", () => {
     it("builds a graph for literals", () => {
       const resource = build.subject({
--- a/__tests__/GraphBuilder.test.js
+++ b/__tests__/GraphBuilder.test.js
@@ -38,6 +38,44 @@ describe("GraphBuilder", () => {
       expect(new GraphBuilder(resource).graph.toCanonical()).toMatch(rdf)
     })
 
+    it("builds a graph for literal with validationDataType", () => {
+      const resource = build.subject({
+        subjectTemplate: build.subjectTemplate({
+          id: "resourceTemplate:testing:literalValidation",
+          clazz: "http://sinopia.io/testing/LiteralValidation",
+        }),
+        properties: [
+          build.literalProperty({
+            propertyTemplate: build.propertyTemplate({
+              subjectTemplateKey: "resourceTemplate:testing:literalValidation",
+              label: "literalValidation, integer validationDataType",
+              uris: {
+                "http://sinopia.io/testing/LiteralValidation/property3":
+                  "http://sinopia.io/testing/LiteralValidation/property3",
+              },
+              type: "literal",
+              validationDataType: "http://www.w3.org/2001/XMLSchema#integer",
+              languageSuppressed: true,
+              component: "InputLiteral",
+            }),
+            values: [
+              build.literalValue({
+                literal: "literal with dataType",
+                lang: null,
+                propertyUri:
+                  "http://sinopia.io/testing/LiteralValidation/property3",
+              }),
+            ],
+          }),
+        ],
+      })
+
+      const rdf = `<> <http://sinopia.io/testing/LiteralValidation/property3> "literal with dataType"^^<http://www.w3.org/2001/XMLSchema#integer> .
+<> <http://sinopia.io/vocabulary/hasResourceTemplate> "resourceTemplate:testing:literalValidation" .
+<> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://sinopia.io/testing/LiteralValidation> .`
+      expect(new GraphBuilder(resource).graph.toCanonical()).toMatch(rdf)
+    })
+
     it("builds a graph for uris", () => {
       const resource = build.subject({
         subjectTemplate: build.subjectTemplate({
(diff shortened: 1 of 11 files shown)


__tests__/GraphBuilder.test.js
```diff
@@ -38,6 +38,44 @@ describe("GraphBuilder", () => {
...
+          build.literalProperty({
+            propertyTemplate: build.propertyTemplate({
+              subjectTemplateKey: "resourceTemplate:testing:literalValidation",
+              label: "literalValidation, integer validationDataType",
+              uris: {
+                "http://sinopia.io/testing/LiteralValidation/property3":
+                  "http://sinopia.io/testing/LiteralValidation/property3",
+              },
+              type: "literal",
+              validationDataType: "http://www.w3.org/2001/XMLSchema/integer",
```
justinlittman, line 57: Turns out I had this wrong. It should be `http://www.w3.org/2001/XMLSchema#integer` (and same for other datatypes).

__tests__/testUtilities/resourceBuilderUtils.js
```diff
@@ -154,7 +154,12 @@ export default class ResourceBuilder {
 
     // Inject property into value
     if (!_.isEmpty(values) && this.injectPropertyIntoValue)
-      values.forEach((value) => (value.propertyKey = property.key))
+      values.forEach((value) => {
+        value.propertyKey = property.key
+        if (value.literal && property?.propertyTemplate?.validationDataType) {
```
justinlittman, line 159: Not thrilled with this solution. It is kludging in your particular test, but would be unclear in future why this is done. I would prefer to have this be controlled by separate configuration, like `injectPropertyIntoValue`.

From a naming perspective, probably best to rename `injectPropertyIntoValue` to ``injectPropertyKeyIntoValue`, but I defer to your judgment.

src/GraphBuilder.js
```diff
@@ -106,7 +106,15 @@ export default class GraphBuilder {
   }
 
   buildLiteralValue(value, subjectTerm, propertyTerm) {
-    const valueTerm = rdf.literal(value.literal, value.lang)
+    const validationDataType =
+      value.property?.propertyTemplate?.validationDataType
```
justinlittman, line 110: What happens if we change this to `value.property.propertyTemplate.validationDataType`? My concern is that when the property template is missing, we're defaulting to using language, when it is probably an error.

justinlittman (changes requested):

ndushay (commented):
thanks for the review, @justinlittman;  Some things I can address but others I'm not so sure.

__tests__/GraphBuilder.test.js
ndushay: @peetucket this relates to your PR as well.   I guess your PR now needs to depend on this PR, which will contain the fix.

peetucket: yes, that request was also noted in my PR, and I fixed it and then changed it everywhere else in the code that it occurred: https://github.com/LD4P/sinopia_editor/pull/3349

that PR was just approved, so we can merge it 

i think this is a new spot though, so you should only need to change this and not the other ones

src/GraphBuilder.js
ndushay: I'm not clear on when the propertyTemplate would be missing in real data.   In our test data, removing the `?` gives  things like:

```
Failed test diffs:
 FAIL  /__tests__/actionCreators/resources.newResourceFromDataset.test.js
  ● newResourceFromDataset › loading a resource › dispatches actions

    TypeError: Cannot read properties of undefined (reading 'propertyTemplate')

      108 |   buildLiteralValue(value, subjectTerm, propertyTerm) {
      109 |     const validationDataType =
    > 110 |       value.property.propertyTemplate?.validationDataType
          |                      ^
      111 |     let valueTerm
      112 |     if (validationDataType) {
      113 |       const namedNode = rdf.namedNode(validationDataType)

      at GraphBuilder.buildLiteralValue (src/GraphBuilder.js:110:22)
      at GraphBuilder.buildValue (src/GraphBuilder.js:104:12)
      at forEach (src/GraphBuilder.js:87:14)
          at Array.forEach (<anonymous>)
      at GraphBuilder.buildProperty (src/GraphBuilder.js:86:14)
      at forEach (src/GraphBuilder.js:39:12)
          at Array.forEach (<anonymous>)
      at GraphBuilder.buildSubject (src/GraphBuilder.js:38:24)
      at GraphBuilder.graph (src/GraphBuilder.js:25:12)
      at _callee$ (__tests__/actionCreators/resources.newResourceFromDataset.test.js:82:9)
...
```

And using the new "injectPropertyIntoValue" flag doesn't help:

```
Failed test diffs:
 FAIL  /__tests__/actionCreators/resources.newResourceFromDataset.test.js
  ● Test suite failed to run

    TypeError: Converting circular structure to JSON
        --> starting at object with constructor 'Object'
        |     property 'values' -> object with constructor 'Array'
        |     index 0 -> object with constructor 'Object'
        --- property 'property' closes the circle
        at stringify (<anonymous>)

      at messageParent (node_modules/jest-worker/build/workers/messageParent.js:34:19)
```

And I once agin checked in my browser and am certain we don't have easy access to the `validationDataType` property from any but the `value` argument.   Unless we should be loading up the property from the key ...

ndushay (commented):
Two of your suggestions taken.  The third ... I don't know how to proceed.

justinlittman (approved):

PR #3356, opened 2021-11-18
show network error on lookup instead of no results

### LINKED ISSUE #3341: QA errors should be displayed to user
1. Load Sinopia Editor (you can use http://localhost:8888/editor/resourceTemplate:testing:inputs).
2. Turn off wifi.
3. Perform a lookup.

Expected result:
Error is displayed to user where indicated below using `text-danger`.

Actual result:
User is told no results.

image

peetucket:
## Why was this change made?

Fixes #3341 - show an error on QA lookups if there are network issues instead of showing "No results"

Localhost

### DIFF
--- a/src/components/editor/inputs/LookupTab.jsx
+++ b/src/components/editor/inputs/LookupTab.jsx
@@ -30,12 +30,12 @@ const LookupTab = (props) => {
     )
   }
 
-  if (!props.result.totalHits) {
-    return <strong>No results</strong>
+  if (props.result.error) {
+    return <span className="text-danger">{props.result.error}</span>
   }
 
-  if (props.result.error) {
-    return <span className="dropdown-error">{props.result.error}</span>
+  if (!props.result.totalHits) {
+    return <strong>No results</strong>
   }
 
   const tabResults = props.result.results.map((hit) => (


src/components/editor/inputs/LookupTab.jsx
```diff
@@ -30,12 +30,21 @@ const LookupTab = (props) => {
     )
   }
 
-  if (!props.result.totalHits) {
+  if (props.result.totalHits === 0) {
```
justinlittman, line 33: Actually, looking at this now, I think all that needs to happen is `if (!props.result.totalHits) { ...}` needs to be placed AFTER `if (props.result.error) { ... }`. Can you try that?

justinlittman (changes requested):

src/components/editor/inputs/LookupTab.jsx
peetucket: Are you sure?  The `totalHits` node missing completely is from a network error, while being  `=== 0` means we got a valid response but there were no results.  Aren't these two separate cases?

justinlittman: Doesn't the network error produce an error which is caught and returned in `result.error`? (If not ping me and we can pair on this.)

peetucket: Oh, you are absolutely correct.  Just fixed this up as you suggested.

If there is a network error, you get this message: `NetworkError when attempting to fetch resource.`

peetucket: Note: the diff looks weird, but it's really swapping the order and changing the class for the error message.

justinlittman (approved):

PR #3361, opened 2021-11-19
implement EDTF validationDataType for Literals

### LINKED ISSUE #3301: Implement EDTF validation option for literals
See PR #3296 (for issue #2890) to see how this was done for integers and dateTime validation.

About EDTF:
- https://www.loc.gov/standards/datetime/

Suggest thinking about/discussing with JLitt and/or Michelle:  
- use cases for EDTF.   
    - given that we have the xsd types, are we focusing on EDTF level 0 time intervals?
    - on EDTF level 1 date qualifications?  unspecified digits?  seasons?  negative years (i.e. B.C.)
- what can be implemented easily (we may be able to punt on hard stuff).  Will the same approach as used for dateTime validations work?  (i.e. a regex combined with some javascript Date object parsing)

Some possible javascript packages of use (check for age and support as well as desired features)
- EDTF.js
    - https://www.npmjs.com/package/edtf
    - https://github.com/inukshuk/edtf.js
- edtf-parser
    - https://npm.io/package/edtf-parser
    - https://github.com/nicompte/edtfy
- https://www.loc.gov/standards/datetime/implementations.html

Begin by adding this to `static/literalDataType.json`:
```
  {
    "uri": "http://id.loc.gov/datatypes/edtf",
    "label": "Extended Date/Time Format (EDTF)"
  }
```

This will add EDTF to the validationDataType pulldown menu for editing literal attributes in a propertyTemplate.

image

ndushay:
## Why was this change made?

Fixes #3301 - implements EDTF validation for Literals.

### After:

image

unit tests added akin to those for dateTimeStamp, dateTime and integer literal validationDataType


ndushay:
BUILD:  integration test appears to be choking on `edtf` package.  Does something need to change in our docker container?

src/reducers/resourceValidationHelpers.js
```diff
@@ -88,6 +88,32 @@ export const literalDateTimeStampValidationError = (
   return errors
 }
 
+export const literalEdtfValidationError = (value, propertyTemplate) => {
+  const errors = []
+  const edtf = require("edtf")
```
mjgiarlo, line 93: If this is causing a test issue, you could try importing the library like we do the other ones, by putting this line near the top of the file:

```javascript
import edtf from "edtf"
```

🤷🏻‍♂️

ndushay: import statement doesn't work;  there's something about how edtf module is written.  https://www.npmjs.com/package/edtf suggests

```
import edtf, { Date, Interval } from 'edtf'
```

but that won't work.   The require did work to get the test passing.

```
import { edtf } from 'edtf`
```

fails the test.

mjgiarlo (approved):

ndushay:
@justinlittman   It appears that the `esm` module needed for `serializer-jsonld-ext` upgrade does NOT play nicely with the `edtf` module.

justinlittman:
From the failed integration test, it looks like a polyfill is needed for `assert`. I added it.

justinlittman:
What makes you say
|  It appears that the esm module needed for serializer-jsonld-ext upgrade does NOT play nicely with the edtf module.
?

justinlittman:
See https://github.com/inukshuk/edtf.js/issues/35

PR #3368, opened 2021-11-23
Adds display for initial metrics.

### LINKED ISSUE #3365: Initial metrics display
Provide a display for the metrics implemented by https://github.com/LD4P/sinopia_api/pull/227

justinlittman:
closes #3365

## Why was this change made?
Metrics!

### DIFF
--- a/src/components/metrics/ResourceMetrics.jsx
+++ b/src/components/metrics/ResourceMetrics.jsx
@@ -1,0 +1,35 @@
+// Copyright 2019 Stanford University see LICENSE for license
+
+import React from "react"
+import PropTypes from "prop-types"
+import MetricsWrapper from "./MetricsWrapper"
+import CountCard from "./CountCard"
+import useMetric from "hooks/useMetric"
+
+const ResourceMetrics = ({ triggerHandleOffsetMenu }) => {
+  const resourceCountMetric = useMetric("getResourceCount")
+
+  return (
+    <MetricsWrapper
+      triggerHandleOffsetMenu={triggerHandleOffsetMenu}
+      title="Resource metrics"
+    >
+      <div className="row">
+        <div className="col-md-3">
+          {resourceCountMetric?.count && (
+            <CountCard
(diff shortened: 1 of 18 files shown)


justinlittman:
image

justinlittman:
image

image

src/components/metrics/ResourceMetrics.jsx
```diff
@@ -0,0 +1,35 @@
...
+      </div>
+    </MetricsWrapper>
+  )
+}
+
+ResourceMetrics.propTypes = {
+  triggerHandleOffsetMenu: PropTypes.func,
+}
+
+export default ResourceMetrics
```
ndushay, line 35: I wonder if it makes sense for the component to take an argument as to which kind of metric is desired, since these are all so similar.   I also am fine with it as is;  I think the 4th one is the tipping point, not the 3rd.

src/sinopiaApi.js
```diff
@@ -11,14 +11,21 @@ import {
 } from "../__tests__/testUtilities/fixtureLoaderHelper"
 import GraphBuilder from "GraphBuilder"
 import { v4 as uuidv4 } from "uuid"
-import Auth from "@aws-amplify/auth"
```
ndushay, line 14: this removal surprises me, but I trust you.

ndushay (approved):
couple of stray thoughts;  approving.

src/components/metrics/ResourceMetrics.jsx
justinlittman: These will diverge as more metrics are added. It just so happens that the first 3 easiest metrics were all counts.

justinlittman: And, this is going to be an evolving interface, so I suspect that there will be more opportunities for refactoring emerging.

src/sinopiaApi.js
justinlittman: It was moved to a helper module.,

PR #3370, opened 2021-11-23
Move to a singleton InputLang modal.

justinlittman:
## Why was this change made?
One modal is more efficient than one modal per input.

Unit

### DIFF
--- a/src/store.js
+++ b/src/store.js
@@ -18,6 +18,7 @@ export const initialState = {
     currentPreviewResource: undefined,
     currentComponent: {},
     currentModal: undefined,
+    currentLangModalValue: undefined, // the value key of the value to be displayed in the InputLang modal.
     currentDiff: {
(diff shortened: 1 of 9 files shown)


src/store.js
```diff
@@ -18,6 +18,7 @@ export const initialState = {
     currentPreviewResource: undefined,
     currentComponent: {},
     currentModal: undefined,
+    currentLangModalValue: undefined, // the value key of the value to be displayed in the InputLang modal.
```
ndushay, line 21: "value key"?   vs "key"?

ndushay (approved):
I wouldn't have come up with this in a million years - great job!

src/store.js
justinlittman: Yeah, I thought about this, but dropping the "key" is consistent with the naming conventions in the editor section of the store.

PR #3373, opened 2021-11-24
Display relationships in search results.

### LINKED ISSUE #3269: give user easy way to copy Shareable URL for resources
#2990 is great in that we now have a shareable URL for any resource or template, but we don't show it to the user so user can easily copy it and share it

can we expose this to the user so that for example, when I see a resource in search results, or I'm previewing or editing that resource, I can "copy shareable URL" (or whatever makes sense, or an icon to represent this) to send it to someone else? right now the user would have to construct the URL themselves

@astridu we need a design :)

we do already have "Copy URI" which is useful if you want to use a Sinopia URI to fill in a field in a template. so we wouldn't want to encourage users to confuse "Copy URI" and "Copy shareable URL"

justinlittman:
closes #3269

## Why was this change made?
Easier discovery of BF resources.

### DIFF
--- a/__tests__/actionCreators/relationships.test.js
+++ b/__tests__/actionCreators/relationships.test.js
@@ -1,4 +1,7 @@
-import { loadRelationships } from "actionCreators/relationships"
+import {
+  loadRelationships,
+  loadSearchRelationships,
+} from "actionCreators/relationships"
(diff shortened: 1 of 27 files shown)


justinlittman:
image

mjgiarlo (approved):

PR #3375, opened 2021-11-29
Allow selecting a default language for a resource.

### LINKED ISSUE #3372: Default language for a resource
Based on https://github.com/LD4P/sinopia_editor/issues/2106

image

When the language is submitted from the Lang modal and if this checkbox is selected, this language should be used as the default for this resource. That means that all blank literals and labels should have their language changed to this language and all new blank literals and labels should be set to this language. Existing non-blank literals and labels should not be changed.

When the Lang modal is re-opened, if the default language is the language displayed by the modal, then the checkbox should be checked. Otherwise, it should not be checked.

The default language should be specified to a resource and should only last the duration of the session. (Thus, it can be stored in redux state and does not need to be otherwise persisted.)

justinlittman:
closes #3372

## Why was this change made?
Less boilerplate for catalogers.


justinlittman:
image

mjgiarlo (approved):

Issue #3379, opened 2021-11-29
Template and resource creation metrics


justinlittman:
https://github.com/LD4P/sinopia_api/pull/233 adds an API endpoint for creation count metrics for templates and resources. For this ticket, these metrics should be displayed in the Editor.

![image](https://user-images.githubusercontent.com/588335/143943533-812a4556-f6df-426a-bbe8-dbf56fb96645.png)

Start date should default to today - 30 days.
End date should default to today.

* Add support for new endpoints to `sinopiaMetrics.js`.
* Generalize `useMetrics` hook to accept parameters (in this case, start date, end date, and groups).
* Create a new card (like CountCard) that allows the user to specify start date, end date, and groups (alternatively, enhance the existing CountCard).
* Add the new card to the templates metrics and resource metrics pages.

justinlittman:
To assist with this ticket and since we don't have UI designs, I've stubbed out the card: https://github.com/LD4P/sinopia_editor/pull/3380

PR #3380, opened 2021-11-29
Template and resource creation metrics

justinlittman:
refs #3379

## Why was this change made?
Provide additional metrics.

### DIFF
--- a/src/components/metrics/CountCard.jsx
+++ b/src/components/metrics/CountCard.jsx
@@ -1,20 +1,34 @@
 import React from "react"
 import PropTypes from "prop-types"
+import _ from "lodash"
 
-const CountCard = ({ title, count }) => (
+const CountCard = ({ title, count = null, footer = null }) => (
   <div className="card">
     <div className="card-header">
       <h5>{title}</h5>
     </div>
     <div className="card-body">
-      <p className="card-text display-1 text-center">{count}</p>
+      <div className="card-text display-1 text-center">
+        {_.isNull(count) ? (
+          <div className="spinner-border" role="status">
+            <span className="visually-hidden">Loading...</span>
+          </div>
+        ) : (
+          <React.Fragment>{count}</React.Fragment>
+        )}
+      </div>
     </div>
+    {footer && <div className="card-footer">{footer}</div>}
   </div>
 )
 
 CountCard.propTypes = {
   title: PropTypes.string.isRequired,
-  count: PropTypes.number.isRequired,
+  count: PropTypes.number,
+  footer: PropTypes.oneOfType([
+    PropTypes.node,
+    PropTypes.arrayOf(PropTypes.node),
+  ]),
 }
 
 export default CountCard
--- a/src/components/metrics/MetricsWrapper.jsx
+++ b/src/components/metrics/MetricsWrapper.jsx
@@ -1,22 +1,32 @@
 // Copyright 2019 Stanford University see LICENSE for license
 
-import React from "react"
+import React, { useEffect } from "react"
+import { useDispatch } from "react-redux"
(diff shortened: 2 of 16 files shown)


src/components/metrics/CountCard.jsx
```diff
@@ -1,20 +1,34 @@
 import React from "react"
 import PropTypes from "prop-types"
+import _ from "lodash"
 
-const CountCard = ({ title, count }) => (
+const CountCard = ({ title, count = null, footer = null }) => (
```
justinlittman, line 5: Made the count card more flexible by allowing a footer to be supplied. For the new cards, this contains the inputs.

src/components/metrics/CountCard.jsx
```diff
@@ -1,20 +1,34 @@
...
   <div className="card">
     <div className="card-header">
       <h5>{title}</h5>
     </div>
     <div className="card-body">
-      <p className="card-text display-1 text-center">{count}</p>
+      <div className="card-text display-1 text-center">
```
justinlittman, line 11: Added a spinner while waiting.

src/components/metrics/MetricsWrapper.jsx
```diff
@@ -1,22 +1,32 @@
...
-      <ContextAlert />
-      <h3>{title}</h3>
-      {children}
-    </div>
-  </AlertsContextProvider>
-)
+const MetricsWrapper = ({ title, children, triggerHandleOffsetMenu }) => {
+  const dispatch = useDispatch()
+
+  useEffect(() => {
```
justinlittman, line 15: Changed so that the error is cleared in the wrapper.

src/components/metrics/ResourceCountMetric.jsx
```diff
@@ -0,0 +1,13 @@
+// Copyright 2019 Stanford University see LICENSE for license
+
+import React from "react"
+import CountCard from "./CountCard"
+import useMetric from "hooks/useMetric"
+
+const ResourceCountMetric = () => {
```
justinlittman, line 7: Moved the count metrics into their own components.

src/utilities/SinopiaApiHelper.js
```diff
@@ -20,8 +21,8 @@ export const checkResp = (resp) => {
     })
 }
 
-export const getJson = (url) =>
-  fetch(url, {
+export const getJson = (url, queryParams = {}) =>
```
justinlittman, line 24: Added support for optional query params.

src/components/metrics/CountCard.jsx
```diff
@@ -1,20 +1,34 @@
...
 )
 
 CountCard.propTypes = {
   title: PropTypes.string.isRequired,
-  count: PropTypes.number.isRequired,
+  count: PropTypes.number,
+  footer: PropTypes.oneOfType([
+    PropTypes.node,
+    PropTypes.arrayOf(PropTypes.node),
+  ]),
```
mjgiarlo, line 31: Can it be one or the other, currently? Below, I'm only seeing fragment instances being provided as footers but maybe I'm overlooking something. Or maybe this is future-proofing?

mjgiarlo (approved):
Have a minor question about a possibly too permissive prop, but this is good to go.

src/components/metrics/CountCard.jsx
justinlittman: Per Google, this is the recommended approach for this sort of property, so yeah, future-proofing.

justinlittman: Also, this is used elsewhere in our code, so this is consistent.

mjgiarlo: 🍻

Issue #3357, opened 2021-11-18
User id literal default


justinlittman:
The goal of this ticket is to allow a literal value to be defaulted with a user's cognito user id.

- [x] Add a new literal attributes field to the `rt_literal_property_attrs` template. See https://github.com/LD4P/sinopia_editor/blob/main/static/templates/rt_property_template_doc.json#L3-L35 for where similar is done for properties. This will be a lookup to a file, but the lookup will only have a single entry (something like `http://sinopia.io/vocabulary/literalAttribute/userIdDefault`).
- [x] Register new URIs in Vocab component
- [x] Add a `userIdDefault` field (true or false) to the property template state model which is set by the TemplatesBuilder.
- [x] Validate in templateValidators that a literal cannot have defaults and a user id default. (It only makes sense to have one type of default for literal property.)
- [ ] When creating a blank literal value, if that value is for a property that has a user id default set that value to the cognito user id (instead of blank).

mjgiarlo:
@justinlittman see #3382 if helpful

PR #3382, opened 2021-11-29
Add a new literal data type for user ID default

mjgiarlo:
Connects to #3357

## Why was this change made?

...

...

...

### DIFF
--- a/__tests__/TemplatesBuilder.test.js
+++ b/__tests__/TemplatesBuilder.test.js
@@ -21,7 +21,7 @@ describe("TemplatesBuilder", () => {
 <http://id.loc.gov/ontologies/bibframe/Uber3> <http://www.w3.org/2000/01/rdf-schema#label> "Uber3"@en .`
 
     const dataset = await datasetFromN3(rdf)
-    const subjectTemplate = new TemplatesBuilder(dataset, "", "stanford", [
+    const subjectTemplate = new TemplatesBuilder(dataset, "", "", "stanford", [
       "cornell",
     ]).build()
     expect(subjectTemplate).toStrictEqual(
@@ -99,7 +99,7 @@ _:b1_c14n0 <http://www.w3.org/2000/01/rdf-schema#label> "Uber template1, propert
     )
   })
 
-  it("builds literal property template with validationDataType value", async () => {
+  it("builds literal property template", async () => {
     const rdf = `<> <http://sinopia.io/vocabulary/hasClass> <http://id.loc.gov/ontologies/bibframe/Uber1> .
 <> <http://sinopia.io/vocabulary/hasPropertyTemplate> _:b1_c14n1 .
 <> <http://sinopia.io/vocabulary/hasResourceId> <resourceTemplate:testing:uber1> .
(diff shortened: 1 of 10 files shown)


static/templates/rt_literal_property_attrs_doc.json
```diff
@@ -100,6 +100,50 @@
...
+      }
+    ],
+    "http://sinopia.io/vocabulary/hasRemark": [
+      {
+        "@value": "Data type specifying a user ID default literal."
+      }
+    ],
+    "http://sinopia.io/vocabulary/hasPropertyUri": [
+      {
+        "@id": "http://sinopia.io/vocabulary/userIdDefaultDataType"
```
mjgiarlo, line 118: @justinlittman Is this an acceptable value? I made this up. Not sure if this should be the same or different from the URI in `static/userIdDefaultDataType.json`.

static/userIdDefaultDataType.json
```diff
@@ -0,0 +1,6 @@
+[
+  {
+    "uri": "http://sinopia.io/vocabulary/literalAttribute/userIdDefault",
```
mjgiarlo, line 3: This value was suggested in #3357.

static/userIdDefaultDataType.json
```diff
@@ -0,0 +1,6 @@
...
+    "label": "userid default"
```
mjgiarlo, line 4: Will this be exposed to users? In which case perhaps something more like `user ID default` might be better?

static/templates/rt_literal_property_attrs_doc.json
```diff
@@ -100,6 +100,50 @@
...
+    "@id": "_:b6",
+    "@type": ["http://sinopia.io/vocabulary/PropertyTemplate"],
+    "http://www.w3.org/2000/01/rdf-schema#label": [
+      {
+        "@value": "Literal property attributes"
+      }
+    ],
+    "http://sinopia.io/vocabulary/hasRemark": [
+      {
+        "@value": "Whether literal property is repeatable, required, ordered, etc."
```
justinlittman, line 113: Needs to be changed.

src/components/vocabulary/Vocab.js
```diff
@@ -36,6 +36,10 @@ const vocabulary = {
     description: "Attributes for a literal",
     url: "http://sinopia.io/vocabulary/hasLiteralAttributes",
   },
+  hasLiteralPropertyAttribute: {
```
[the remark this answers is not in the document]
mjgiarlo: @justinlittman Hmmm, the example I was basing this off of (from #3357) is `hasPropertyAttribute`: https://github.com/LD4P/sinopia_editor/blob/main/static/templates/rt_property_template_doc.json#L3-L35 But I do see other `has...Attributes` terms below. 

I can go with a plural here, no problem. Does this suggest that `hasPropertyAttribute` should be pluralized at some point too?

static/templates/rt_literal_property_attrs_doc.json
mjgiarlo: Sure, what text would you like?

static/templates/rt_literal_property_attrs_doc.json
```diff
@@ -100,6 +100,50 @@
...
+    "http://sinopia.io/vocabulary/hasPropertyUri": [
+      {
+        "@id": "http://sinopia.io/vocabulary/hasLiteralPropertyAttribute"
+      }
+    ],
+    "http://sinopia.io/vocabulary/hasPropertyAttribute": [
+      {
+        "@id": "http://sinopia.io/vocabulary/propertyAttribute/repeatable"
+      }
+    ],
```
mjgiarlo, line 125: @justinlittman I'm also not sure whether this is the right value.

src/components/vocabulary/Vocab.js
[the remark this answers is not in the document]
justinlittman: Yes, but no easy way to make that change without breaking everything.

static/templates/rt_literal_property_attrs_doc.json
justinlittman: "Whether literal property has user ID default."

src/components/vocabulary/Vocab.js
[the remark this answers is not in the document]
mjgiarlo: @justinlittman Yeah, let's not do that.

__tests__/TemplatesBuilder.test.js
```diff
@@ -142,48 +147,6 @@ _:b2_c14n1 <http://sinopia.io/vocabulary/hasValidationDataType> <http://www.w3.o
     )
   })
 
-  it("builds literal property template without validationDataType value", async () => {
```
justinlittman, line 145: This test was redundant, so removed.

src/TemplatesBuilder.js
```diff
@@ -147,7 +148,24 @@ export default class TemplatesBuilder {
...
+      propertyTerm,
+      "http://sinopia.io/vocabulary/hasLiteralAttributes"
+    )
+    if (attributeTerm) {
+      propertyTemplate.defaults = this.defaultsForLiteral(attributeTerm)
+      const literalAttrValues = this.valuesFor(
+        this.attributeTerm,
+        "http://sinopia.io/vocabulary/hasLiteralPropertyAttributes"
+      )
+      if (
```
justinlittman, line 161: Here's where the user id gets added as a default.

src/components/vocabulary/Vocab.js
```diff
@@ -111,6 +115,11 @@ const vocabulary = {
     description: "Class for a property template",
     url: "http://sinopia.io/vocabulary/PropertyTemplate",
   },
+  "literalPropertyAttribute/userIdDefault": {
```
ndushay, line 118: should this start with "has" ?  "hasLiteralAttributeUserIdDefault" ?

ndushay (changes requested):
Looks fine;  couple of quibbles.

src/components/vocabulary/Vocab.js
justinlittman: This is referencing the `userIdDefault` in `literalPropertyAttribute` so this is consistent naming.

static/templates/rt_literal_property_attrs_doc.json
justinlittman: This is a general attribute field; in the future in might contain attributes that are unrelated to defaults.

justinlittman:
@ndushay Thanks for paying attention to the naming. The naming gets really confusing (would make some different choices if starting over...). However, I think the naming in this PR is consistent with other naming in the base templates.

ndushay (approved):

PR #3387, opened 2021-11-30
Fix paging for lookups. Add test template for lookups.

justinlittman:
## Why was this change made?
So that paging works correctly for lookups and to allow easier template testing.

Local


__tests__/testUtilities/fixtureLoaderHelper.js
```diff
@@ -19,7 +19,6 @@ const resourceFilenames = {
 
 const templateFilenames = {
   "test:resource:DiscogsLookup": "DiscogsLookup.json",
-  "test:bf2:soundCharacteristics": "multiple_loc.json",
```
justinlittman, line 22: Not used.

mjgiarlo (approved):

PR #3394, opened 2021-12-01
Refactor modals.

### LINKED ISSUE #3288: Accessibility, focus is lost when clicking save and the modal for groups appears
For good accessibility, the focus should go to the first element of the modal and focus should not escape the modal until the modal is closed. 

I suspect this problem exists because we are trying to handle the modal logic in Redux rather than using the javascript provided by bootstrap for this purpose.

justinlittman:
closes #3288

## Why was this change made?
Improved accessibility, less boilerplate.


mjgiarlo (approved):
Feel free to merge when green and working!

PR #3401, opened 2021-12-01
Adds help text for metrics.

justinlittman:
## Why was this change made?
Clearer for users.

image

### DIFF
--- a/src/hooks/useMetric.js
+++ b/src/hooks/useMetric.js
@@ -1,4 +1,4 @@
-import { useEffect, useState } from "react"
+import { useEffect, useState, useRef } from "react"
 import { useDispatch } from "react-redux"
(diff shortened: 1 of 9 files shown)


src/hooks/useMetric.js
```diff
@@ -7,22 +7,22 @@ import * as sinopiaMetrics from "../sinopiaMetrics"
 const useMetric = (name, params = null) => {
   const dispatch = useDispatch()
   const [metric, setMetric] = useState(null)
-  const [isMounted, setMounted] = useState(false)
+  const isMountedRef = useRef(false)
```
justinlittman, line 10: This keeps the tests happier because it avoids actions after the test is completed.

mjgiarlo (approved):

PR #3409, opened 2021-12-02
add user engagement metrics to editor ui

### LINKED ISSUE #3398: User engagement metrics
See https://github.com/LD4P/sinopia_api/pull/243

This should be added to the user metrics screen.

peetucket:
## Why was this change made?

Fixes #3398 - add user engagement metrics to user metrics UI page

I also tried putting the cards into sets of either two or three columns (in a second commit), instead of all stacked one on top of the other, since I think it looks better.  See screenshots.

Note: i created a component for the existing user count metric to match the style of the other metrics pages (where each card was its own component)

Screen Shot 2021-12-02 at 12 57 40 PM
Screen Shot 2021-12-02 at 12 57 46 PM
Screen Shot 2021-12-02 at 12 57 51 PM

Localhost

### DIFF
--- a/__tests__/feature/metrics/viewResourceMetrics.test.js
+++ b/__tests__/feature/metrics/viewResourceMetrics.test.js
@@ -47,7 +47,7 @@ describe("viewing resource metrics", () => {
       await screen.findByText("Resource editing")
       screen.getByText("10", { selector: ".card-text" })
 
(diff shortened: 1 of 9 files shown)


ndushay (approved):
LGTM (but what do I know?)

Issue #3154, opened 2021-10-12
confusing proliferation of label fields in template creation


michelleif:
Now that we require a label for every URI, we have several places in template creation where there **appear** to be 2 labels for the same thing. One of those labels seems to be extraneous and never surfaced anywhere. This is confusing for the template creator.

1) A template has a single `Class` (which is a URI) and a single `Template Label`. But now that every URI requires a label, there's an additional label field.

Former design (3.7.1 still on PRD):
![Screen Shot 2021-10-12 at 7.22.39 AM.png](https://images.zenhubusercontent.com/5d825bf74def500001d12e70/f119b20e-52ba-4a05-b028-59cd57b66fad)

Post release 3.9.0:
![Screen Shot 2021-10-12 at 7.23.14 AM.png](https://images.zenhubusercontent.com/5d825bf74def500001d12e70/96b8976d-6991-4417-85ac-f7ea214dbf31)

I know that one is labeled `Template Label` and one is labeled `Label for Class` but this is confusing. And we never expose `Label for Class`.

2) Every property requires a URI and a label, but now that every URI requires a label, there's an additional label field:

Former design (3.7.1 still on PRD):
![Screen Shot 2021-10-12 at 7.48.50 AM.png](https://images.zenhubusercontent.com/5d825bf74def500001d12e70/daf90464-4a5d-4e6d-8671-55489f096391)

Post release 3.9:
![Screen Shot 2021-10-12 at 7.49.36 AM.png](https://images.zenhubusercontent.com/5d825bf74def500001d12e70/3c5c3734-fd61-4c0d-a3c2-6c3325dc17d5)

3) URI and lookup fields allow definition of a default URI and label. But now that every URI requires a label, there's an additional label field:
Former design (3.7.1 still on PRD):
![Screen Shot 2021-10-12 at 7.25.33 AM.png](https://images.zenhubusercontent.com/5d825bf74def500001d12e70/2b80e6d7-e826-49f4-acb0-0e6122ec3c17)

Post release 3.9:
![Screen Shot 2021-10-12 at 7.27.18 AM.png](https://images.zenhubusercontent.com/5d825bf74def500001d12e70/025c599d-080e-4729-8a60-19940b7784a9)

This is even more confusing because to enter your default label into the **correct** label field, you have to expand the field, whereas the newly added label field on same row as URI field **seems** like the place to supply your default label to go with your default URI.

**Proposed solution**
[...]

PR #3412, opened 2021-12-03
Adds suppressing of URI labels and applies to base templates.

justinlittman:
refs #3154

## Why was this change made?
Less confusing templates.

image

image


README.md
```diff
@@ -310,6 +310,7 @@ The following are only in the resource subject (that is, the base subject).
   repeatable: <true | false>
   ordered: <true | false>
   languageSuppressed: <true | false>
+  labelSuppressed: <true | false>
```
justinlittman, line 313: This is the new property template attribute in the state model.

__tests__/__action_fixtures__/subjectTemplate-suppressible.js
```diff
@@ -18,7 +18,7 @@ const subjectTemplate = build.subjectTemplate({
       uris: {
         "http://sinopia.io/testing/Suppressible/property1": "Property1",
       },
-      valueSubjectTemplateKeys: ["resourceTemplate:testing:uri"],
+      valueSubjectTemplateKeys: ["resourceTemplate:testing:suppressedUri"],
```
justinlittman, line 21: I switched around two template fixtures, so a handful of changes like this where they were used.

src/TemplatesBuilder.js
```diff
@@ -169,6 +169,11 @@ export default class TemplatesBuilder {
     )
     if (attributeTerm) {
       propertyTemplate.defaults = this.defaultsForUri(attributeTerm)
+      const uriAttrValues = this.valuesFor(
```
justinlittman, line 172: This is where suppressed label gets loaded in the TemplateBuilder.

src/actionCreators/resourceHelpers.js
```diff
@@ -504,7 +504,7 @@ const newUriFromObject = (obj, property, propertyUri, context) => {
   const labelQuads = context.dataset
     .match(obj, rdf.namedNode("http://www.w3.org/2000/01/rdf-schema#label"))
     .toArray()
-  let label = uri
+  let label = null
```
justinlittman, line 507: No longer defaulting the label for a URI to that URI when a label is not present.

mjgiarlo (approved):

Issue #3416, opened 2021-12-03
Make language suggestion


justinlittman:
Provide a language suggestion using the API endpoint implemented for https://github.com/LD4P/sinopia_api/issues/246

![image](https://user-images.githubusercontent.com/588335/144622978-e9833570-c21a-4803-9361-e98c6ea5b355.png)
_The working should be "Detected Spanish (spa). Click to use." rather than what shown above._

* Only the top suggestion should be displayed.
* The suggestion should only be displayed when certainty is >= .75.
* The suggestion should only be displayed when the language is different than the currently selected language.
* The language tag might contain a script subtag (e.g., `zh-TW`) so parsing the tag is necessary. BONUS: Display suggestions for script as well when script found.)

justinlittman:
See https://github.com/LD4P/sinopia_api/pull/249

justinlittman:
As part of developing the language detection endpoint, I stubbed out the client method in Sinopia Editor. See https://github.com/LD4P/sinopia_editor/pull/3433

justinlittman:
Fixed by #3433

PR #3433, opened 2021-12-07
Provide language suggestions.

justinlittman:
refs #3416

## Why was this change made?
Less work for catalogers.

### DIFF
--- a/src/components/editor/inputs/InputLang.jsx
+++ b/src/components/editor/inputs/InputLang.jsx
@@ -12,9 +12,11 @@ import {
   selectScripts,
   selectTransliterations,
   selectDefaultLang,
+  selectLanguageLabels,
 } from "selectors/languages"
 import { selectNormValue } from "selectors/resources"
 import { parseLangTag, stringifyLangTag } from "utilities/Language"
+import { detectLanguage } from "sinopiaApi"
 import _ from "lodash"
 
 const InputLang = () => {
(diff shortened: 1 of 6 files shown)


justinlittman:
image

src/components/editor/inputs/InputLang.jsx
```diff
@@ -66,6 +74,25 @@ const InputLang = () => {
       setSelectedTransliterationOptions(newTransliterationOptions)
   }, [value, langOptions, scriptOptions, transliterationOptions])
 
+  useEffect(() => {
+    if (_.isEmpty(textValue)) return
+
+    detectLanguage(textValue)
+      .then((resp) => {
+        if (!_.isEmpty(resp) && resp[0].score >= 0.75) {
```
mjgiarlo, line 82: Worth stuffing the `0.75` value in config and referencing it here? Or not really worth making the change at this point?

mjgiarlo (approved):
One minor question. Approving.

src/components/editor/inputs/InputLang.jsx
justinlittman: My general philosophy is that if a value is only used once and it won't be changed at runtime, then put it with the code rather than config.

mjgiarlo: @justinlittman Good philosophy!

PR #3465, opened 2021-12-15
Add subclasses to RDF for nested resources.

### LINKED ISSUE #3463: Ability to choose subclasses not reflected in the rdf for a contribution template
**Describe the bug**
Not sure if this is a bug or intentional. I created a template in dev (stanford:bf2:Contribution:Collapsed) to see if I could simply list the subclasses of agent rather than have an individual nested resource for each. While I can create the list, it is not reflected in the rdf when I select a subclass by clicking on the checkbox. I tried this both within a larger template (stanford:bf2:Monograph:Work) and on its own.

- What is the impact on your work? 
I'm looking for a way to keep the subclasses in a particular order, which I can't do with the nested templates. But mainly, it seems logical if you can do it with a work template & an instance template, it should work with a contribution template.
- What environment are you in? This is in dev
- What OS and browser are you using? Windows Firefox
stanford:bf2:Monograph:Work
stanford:bf2:Contribution:Collapsed
image
image

justinlittman:
closes #3463

## Why was this change made?
Bug

### DIFF
--- a/src/components/search/Search.jsx
+++ b/src/components/search/Search.jsx
@@ -20,6 +20,7 @@ import AlertsContextProvider from "components/alerts/AlertsContextProvider"
 import ContextAlert from "components/alerts/ContextAlert"
 import { searchErrorKey } from "utilities/errorKeyFactory"
 import TemplateGuessSearchResults from "./TemplateGuessSearchResults"
+import PreviewModal from "../editor/preview/PreviewModal"
 
 const Search = (props) => {
   const { fetchSearchResults } = useSearch()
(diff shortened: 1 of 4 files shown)


src/components/search/Search.jsx
```diff
@@ -44,6 +45,7 @@ const Search = (props) => {
       <div id="search">
         <Header triggerEditorMenu={props.triggerHandleOffsetMenu} />
         <ContextAlert />
+        <PreviewModal />
```
justinlittman, line 48: I realized this was in the wrong place -- it needed to be higher in the component hierarchy.

jcoyne (approved):

PR #3566, opened 2022-07-25
No longer suppress labels for optional classes.

### LINKED ISSUE #3551: Can no longer add labels for "optional" classes
**Describe the bug**
When we first added the "optional" classes (i.e., subclasses of the template class) to Sinopia, we were able to provide labels (or perhaps the earlier template just still contained them). However, when we now want to add an optional class, we are no longer able to provide a label and no label is generated. We really need these labels, given that a. we are not expecting catalogers to be reading URIs; and b. We might in some cases actually change the label for cataloging purposes (e.g., give the RDA term vs the BF term. It also simply brings inconsistencies in our templates.

We can currently work around it, but I don't think it a minor issue. Inconsistency creates confusion, as does the resulting inconsistent terminology or lack thereof
- What environment are you in? Stage
- What OS and browser are you using? Firefox/Chrome
- If you're cataloging, what Resource Template were you using pcc:bf:Monograph:Instance

Screenshot is in template creation mode. The first 3 were from before this problem began and show labels; the last is a new addition and shows the lack of place for a label to be added
image

And this is what it looks like when the new subclass is added in data entry mode:
image

justinlittman:
closes #3551

## Why was this change made?
Per @NancyL.

Local

### DIFF
--- a/static/templates/rt_resource_template_doc.json
+++ b/static/templates/rt_resource_template_doc.json
@@ -195,20 +195,6 @@
       {
         "@id": "http://sinopia.io/vocabulary/propertyType/uri"
       }
-    ],
-    "http://sinopia.io/vocabulary/hasUriAttributes": [
-      {
-        "@id": "_:b19"
-      }
-    ]
-  },
-  {
-    "@id": "_:b19",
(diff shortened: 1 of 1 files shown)


justinlittman:
image

jermnelson (approved):

PR #3567, opened 2022-07-25
No longer suppress labels for chooseable properties.

### LINKED ISSUE #3552: Provide customizable label creation for chooseable property lists
We now have the ability for catalogers to choose a property through a dropdown menu (which we love!). However, the template only allows you to provide a URI for the property, not a label. This is a major drawback. Generally, we don't want to force catalogers to read and interpret URIs as they catalog. Also, in many cases the URI is not human-readable (example follows) or we want to provide a label that conforms with the cataloging standard rather than BIBFRAME.
**Give an example**
A screenshot below shows what it looks like now. Having it with the format "Label (URI in parentheses)" would be good.
Examples of what we like it to look like:

- Date cataloged (http://id.loc.gov/ontologies/bibframe/creationDate)
- Abridgement of (http://rdaregistry.info/Elements/u/P60359)

**Describe alternatives you've considered**
image

justinlittman:
closes #3552

## Why was this change made?
Per @NancyL 

Local

### DIFF
--- a/static/templates/rt_property_template_doc.json
+++ b/static/templates/rt_property_template_doc.json
@@ -267,20 +267,6 @@
       {
         "@id": "http://sinopia.io/vocabulary/propertyType/uri"
       }
-    ],
-    "http://sinopia.io/vocabulary/hasUriAttributes": [
-      {
-        "@id": "_:b27"
-      }
-    ]
-  },
-  {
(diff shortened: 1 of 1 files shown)


justinlittman:
image

jermnelson (approved):

PR #3650, opened 2023-02-13
Update specialcharacters.json

NancyL:
## Why was this change made?
Addition of sound recording (phonogram) copyright date ℗to Symbols beside the regular copyright date. This is necessary for cataloging sound recordings
Character take from Unicode character table


justinlittman (approved):
Thanks @NancyL!

PR #3769, opened 2023-12-08
Add scroll to top button to editor.

### LINKED ISSUE #3763: retain position after saving
Users request the ability to retain their position within a template after saving.

**Give an example**
The close/save menu bar is frozen which is very helpful for saving while working no matter where one is located within a template. When the cataloger saves their work, the cursor and page returns to the top of the template. If the cataloger is saving while they work through a template, they need to scroll down to continue working each time they utilize save.

**Describe alternatives you've considered**
_A clear and concise description of any alternative solutions or features you've considered._
At the current moment, catalogers move back to where they want to work after saving returns them to the top of the template. As an alternative, enabling templates to periodically auto-save would eliminate the need for a default position after saving.

justinlittman:
closes #3763

Note that this was added to the sticky header instead of the bottom of the page.

## Why was this change made?
Editor joy.

image

### DIFF
--- a/src/components/editor/ResourceComponent.jsx
+++ b/src/components/editor/ResourceComponent.jsx
@@ -18,6 +18,7 @@ import SaveAndPublishButton from "./actions/SaveAndPublishButton"
 import CopyToNewButton from "./actions/CopyToNewButton"
 import PreviewButton from "./actions/PreviewButton"
 import ResourceTitle from "./ResourceTitle"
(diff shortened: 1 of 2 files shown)


jermnelson (approved):

jermnelson:
Approved, not sure why the codeclimate/diff-coverage declined so much but feel free to merge if ignorable.

Issue #3776, opened 2023-12-19
@type": "http://id.loc.gov/datatypes/edtf/" has an extra "/"


jimfhahn:
**Describe the bug**
What went wrong? What did you expect to happen?

The JSON-LD serialization of the @type": "http://id.loc.gov/datatypes/edtf/" has an extra "/" at the end. Should it look like this: @type": "http://id.loc.gov/datatypes/edtf"

- What is the impact on your work? (can't work / work stoppage; workaround exists; minor issue)

Major issue -- breaks conversion when we serialize RDF Sinopia in RDF/XML and then try to apply an LC converter

- What environment are you in? (stage or production)
DEV
https://api.development.sinopia.io/resource/90fb418e-4ff4-4296-af69-1e30e2ac2f8a

- What OS and browser are you using? (Mac/Windows; Safari, Firefox, etc.)

Mac/Firefox

- If you're cataloging, what Resource Template were you using (name or ID)

pcc:bf2:Monograph:Instance

** a screenshot**

<img width="658" alt="Screenshot 2023-12-19 at 10 17 42 AM" src="https://github.com/LD4P/sinopia_editor/assets/1310509/565870fa-643a-49ce-b3b1-0b4a05cb9636">

PR #3786, opened 2024-01-24
Datatype edtf

jimfhahn:
## Why was this change made?

addresses #3776 

using npm run test-verbose

no major configurations were updated


package.json
```diff
@@ -106,7 +106,8 @@
     "redux-thunk": "^2.3.0",
     "stream-browserify": "^3.0.0",
     "uuid": "^8.3.1",
-    "weak-napi": "^2.0.2"
+    "weak-napi": "^2.0.2",
+    "cypress": "^9.0.0"
```
jermnelson, line 110: Is there a reason why cypress is included here as a production dependency? It is all ready included later as an optional dependency https://github.com/LD4P/sinopia_editor/blob/66371b31d1552ba6a97e1ea53d52951acbe0501a/package.json#L214-L216

jermnelson (changes requested):
Just a question about the inclusion of the cypress as a production dependency.

package.json
jimfhahn: I inadvertenly caused that addition with some npm process-- I will remove it.

jermnelson (approved):

PR #3805, opened 2024-03-20
Removing cached lookups

sfolsom:
Removing most of the cached lookups that either have replacements or no plans to replace. Will remove other cached lookups once they have API-based lookups when they are available.


justinlittman:
I suspect the build is failing because "urn:ld4p:qa:agrovoc" is heavily used in the tests, but was removed in this PR.

jermnelson (approved):

PR #4046, opened 2025-09-25
Update Config.js to add iowa-state group

kallimathios:
## Why was this change made?

Iowa State University is setting up an integration.


jermnelson:
> Thanks, @jermnelson . I replaced the hyphen with an underscore. Do I need to also match this in the AWS Cognito group ID?

Yes, the AWS Cognito group will need to be modified to match the value in the `src/Config.js` file.

jermnelson (approved):

PR #4118, opened 2026-05-04
Version 3.17.186

jermnelson:
## Why was this change made?
Dependency updates

n/a

`package.lock` and `package.json`


jermnelson:
Updated from changes

